PR #45, opened 2019-01-16
Refactor MH operators

alex-lew:
## Summary
This small PR consolidates three of the MH functions in the standard inference library (`default_mh`, `custom_mh`, and `general_mh`) into a single function (`metropolis_hastings`, with `mh` as an alias). There is now no equivalent to what used to be `simple_mh`. The PR also makes minor compatibility changes / fixes to the examples.

## Examples
The examples in `run_examples.jl` all run, and produce the same output as they do on master (with the exception of the MALA example, for which I've changed a couple parameters). (The `examples/coal/mixture.jl` is not run by `run_examples.jl` and has not been updated. It does not run on master, nor in this branch.)

## Documentation
I left the documentation essentially unchanged: even though `metropolis_hastings` is now one function, each of its three methods is documented separately. Do we want to try to consolidate this documentation? e.g., "You can either provide a set of addresses or a custom proposal. If the custom proposal makes random choices at addresses not present in the original model's trace, it is also necessary to provide an involution."


marcoct:
Re the documentation for `metropolis_hastings` -- I ran the documentation generator on this branch and it seems to put the three definitions into one box. It makes sense.

PR #60, opened 2019-02-03
use logpdf_grad instead of logpdf

marcoct:


marcoct:
Fixes https://github.com/probcomp/Gen/issues/56

PR #102, opened 2019-05-23
Vectorized `Gen.normal` distribution

bzinberg:
# Summary

This PR implements `Gen.broadcasted_normal`, which is like `Gen.normal` except that the parameters `mu` and `std` are broadcast-compatible arrays, and the samples from this distribution have shape `Broadcast.broadcast_shapes(size(mu), size(std))`.  The distribution represents a random `Array` of independent normals.  This is equivalent to (a reshape of) a multivariate normal with diagonal covariance matrix, but its implementation is more efficient than that of the more general `Gen.mvnormal`.

One common use case this distribution serves is modelling IID Gaussian noise.

### DIFF
--- a/src/modeling_library/normal.jl
+++ b/src/modeling_library/normal.jl
@@ -75,7 +75,7 @@ function logpdf_grad(::Normal,
     diff = mu - x
     deriv_x = diff .* precision
     deriv_mu = -deriv_x
-    deriv_std = -1.0 ./ std + (diff .* diff) / (std .* std .* std)
+    deriv_std = -1.0 ./ std + (diff .* diff) ./ (std .* std .* std)
     (deriv_x, deriv_mu, deriv_std)
 end
 
--- a/src/modeling_library/normal.jl
+++ b/src/modeling_library/normal.jl
@@ -71,7 +71,7 @@ function logpdf_grad(::Normal,
                      mu::Array{U, N},
                      std::Array{V, N}) where {T<:Real, U<:Real, V<:Real, N}
     broadcast_compatible_or_crash(x, mu, std)
-    precision = 1.0 / (std .* std)
+    precision = 1.0 ./ (std .* std)
     diff = mu - x
     deriv_x = diff .* precision
     deriv_mu = -deriv_x
--- a/src/modeling_library/normal.jl
+++ b/src/modeling_library/normal.jl
@@ -1,9 +1,43 @@
 struct Normal <: Distribution{Float64} end
 
+function broadcast_compatible_or_crash(args...)
+    # This expression produces a DimensionMismatch exception if the shapes are
+    # incompatible.  There seems to currently be no ready-made exception-free
+    # way to check this, and I think it makes more sense to wait until that
+    # capability exists in Julia proper than to re-implement it here.
+    s = Base.Broadcast.broadcast_shape(map(size, args)...)
+    true
+end
+
+
 """
     normal(mu::Real, std::Real)
 
 Samples a `Float64` value from a normal distribution.
+
+    normal(mu::Array{Real, N}, std::Array{Real, N})
+
+Samples an `Array{Float64, N}` where each element is independently normally
+distributed.  This is equivalent to a multivariate normal with diagonal
+covariance matrix, but its implementation is more efficient than that of the
+more general `mvnormal` for this case.
+
+The shapes of `mu` and `std` must be broadcast-compatible.  For methods such as
+`logpdf(x, mu, std)` which involve an element of the support of the
+distribution, the shapes of `x`, `mu` and `std` must be mutually
+broadcast-compatible.
+
+If `N == 0`, then distribution-related methods such as `logpdf`, `logpdf_grad`
+and `random` return `Float64`s rather than properly returning `Array{Float64,
+0}`s.  This is consistent with Julia's own inconsistency on the matter:
+
+```jldoctest
+julia> typeof(ones())
+Array{Float64,0}
+
+julia> typeof(ones() .* ones())
+Float64
+```
 """
 const normal = Normal()
 
@@ -11,6 +45,14 @@ function logpdf(::Normal, x::Real, mu::Real, std::Real)
     var = std * std
     diff = x - mu
     -(diff * diff)/ (2.0 * var) - 0.5 * log(2.0 * pi * var)
+end
+
+function logpdf(::Normal, x::Array{Real, N},
+                mu::Array{Real, N}, std::Array{Real, N}) where N
+    broadcast_compatible_or_crash(x, mu, std)
+    var = std .* std
+    diff = x - y
+    sum(-(diff .* diff) ./ (2.0 * var) - 0.5 * log.(2.0 * pi * var))
 end
 
 function logpdf_grad(::Normal, x::Real, mu::Real, std::Real)
@@ -22,7 +64,22 @@ function logpdf_grad(::Normal, x::Real, mu::Real, std::Real)
     (deriv_x, deriv_mu, deriv_std)
 end
 
+function logpdf_grad(::Normal, x::Array{Real, N},
+                     mu::Array{Real, N}, std::Array{Real, N}) where N
+    broadcast_compatible_or_crash(x, mu, std)
+    precision = 1.0 / (std .* std)
+    diff = mu - x
+    deriv_x = diff .* precision
+    deriv_mu = -deriv_x
+    deriv_std = -1.0 ./ std + (diff .* diff) / (std .* std .* std)
+    (deriv_x, deriv_mu, deriv_std)
+end
+
 random(::Normal, mu::Real, std::Real) = mu + std * randn()
+function random(::Normal, mu::Array{Real, N}, std::Array{Real, N}) where N
+    broadcast_compatible_or_crash(mu, std)
+    mu + std .* randn(size(mu))
+end
 
 (::Normal)(mu, std) = random(Normal(), mu, std)
 
--- a/src/modeling_library/normal.jl
+++ b/src/modeling_library/normal.jl
@@ -6,7 +6,6 @@ function broadcast_compatible_or_crash(args...)
     # way to check this, and I think it makes more sense to wait until that
     # capability exists in Julia proper than to re-implement it here.
     s = Base.Broadcast.broadcast_shape(map(size, args)...)
-    true
 end
 
 
@@ -15,12 +14,12 @@ end
 
 Samples a `Float64` value from a normal distribution.
 
-    normal(mu::Array{Real, N}, std::Array{Real, N})
+    normal(mu::Array{T}, std::Array{U}) where {T<:Real, U<:Real}
 
-Samples an `Array{Float64, N}` where each element is independently normally
-distributed.  This is equivalent to a multivariate normal with diagonal
-covariance matrix, but its implementation is more efficient than that of the
-more general `mvnormal` for this case.
+Samples an `Array{Float64}` of shape `broadcast(size(mu), size(std))` where each
+element is independently normally distributed.  This is equivalent to a
+multivariate normal with diagonal covariance matrix, but its implementation is
+more efficient than that of the more general `mvnormal` for this case.
 
 The shapes of `mu` and `std` must be broadcast-compatible.  For methods such as
 `logpdf(x, mu, std)` which involve an element of the support of the
@@ -41,20 +40,33 @@ Float64
 """
 const normal = Normal()
 
+function logpdf(::Normal,
+                x::Union{Array{T}, T},
+                mu::Union{Array{U}, U},
+                std::Union{Array{V}, V}) where {T<:Real, U<:Real, V<:Real}
+    broadcast_compatible_or_crash(x, mu, std)
+    var = std .* std
+    diff = x .- mu
+    -(diff .* diff) ./ (2.0 * var) .- 0.5 * log.(2.0 * pi * var)
+end
+
 function logpdf(::Normal, x::Real, mu::Real, std::Real)
     var = std * std
     diff = x - mu
     -(diff * diff)/ (2.0 * var) - 0.5 * log(2.0 * pi * var)
 end
 
-function logpdf(::Normal,
-                x::Array{T, N},
-                mu::Array{U, N},
-                std::Array{V, N}) where {T<:Real, U<:Real, V<:Real, N}
+function logpdf_grad(::Normal,
+                     x::Union{Array{T}, T},
+                     mu::Union{Array{U}, U},
+                     std::Union{Array{V}, V}) where {T<:Real, U<:Real, V<:Real}
     broadcast_compatible_or_crash(x, mu, std)
-    var = std .* std
-    diff = x - y
-    sum(-(diff .* diff) ./ (2.0 * var) - 0.5 * log.(2.0 * pi * var))
+    precision = 1.0 ./ (std .* std)
+    diff = mu .- x
+    deriv_x = diff .* precision
+    deriv_mu = -deriv_x
+    deriv_std = -1.0 ./ std .+ (diff .* diff) ./ (std .* std .* std)
+    (deriv_x, deriv_mu, deriv_std)
 end
 
 function logpdf_grad(::Normal, x::Real, mu::Real, std::Real)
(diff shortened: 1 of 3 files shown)


bzinberg:
Comments pasted from Slack:

@alex-lew
> Nice! This makes sense to me as long as you never need to constrain individual elements and leave others unconstrained — for that case, the Map combinator is the solution (it transforms a generative function to take vector versions of its arguments, and maps the original function over the vectors). In general, before merging this, it seems good to think about what Gen’s relationship with tensors / broadcasting will be — seems a little strange to support this on Normal but not other distributions? But maybe it’s fine!
>
> Unless we decide to make all distributions behave well with arrays/matrices of parameters (which could make sense), it seems like this sort of thing could be added easily to the `@dist` DSL — enabled by a map combinator for distributions.

@bzinberg
> the question about broadcasting is definitely a good one.  I think the decision in question is mostly one of "do we auto-broadcast" -- since the user might not expect that, depending on how we prime them with the rest of Gen's design.  Something more explicit, such as the user specifying the desired shape directly, seems like a pretty much risk-free thing since there's no API extension in that case, just an efficient implementation of a distribution.  Wdyt?

@alex-lew
> This would introduce two different ways of making groups of independent choices: one in which they are all considered “one choice” (a single trace address), and one in which they are given separate addresses (the Map combinator). It seems the first is strictly less flexible than the second? It seems that long-term, the right thing might be to add better (more efficient) support for the second one, so that it is more usable for this type of use case, than to add the first one as an option.

@bzinberg
> My approach there is specifically tailored toward reducing the overhead of having a large number of traced quantities.  Seems to be a bottleneck in Cora work.
> 
> And broadcastability is a big part of Julia so seems like we may want to leverage that.

@alex-lew
> Yes, and I think short term, your solution is the easiest to implement. But maybe it belongs in Cora, and the real solution long term is to reduce overhead of having arrays of independently distributed traced quantities.
>
> “Maybe” because I’m not sure
>
> But it seems plausible to me that we could have the same user interface that you’re suggesting (broadcast-compatible distribution arguments), and have the semantics of that be to have many independent trace addresses. The implementation could still just store a matrix of floats in the trace, and we could have helpers for extracting that matrix / constraining the whole thing to a particular value.
>
> I think that could be made to have little or no overhead. The benefit of this would be that if needed, users could still constrain / observe / propose updates to only part of the matrix.
>
> I’m open to the “distribution” way too, just raising this as a possible alternative solution. And for making progress on Cora right now, I think it definitely makes sense to use the code you’ve written here

@bzinberg
> cool
>
> Btw, I would also claim that there's no API extension in allowing `mu` and `sigma` to broadcast, if we require `x` to have the fully broadcasted shape.  We could call the distribution something like `BroadcastedNormal`.
>
> Although requiring `x` to have the fully broadcasted shape loses some of the benefits of broadcasting.
>
> so if I understand your recommendation: you're suggesting I patch this change into my Cora repo and work from there, while we have a longer-term discussion about this PR?

@alex-lew
> Yes, if that sounds all right to you — but also, if Marco is ok merging, we could do that too! I don’t think Gen is committed to backward compatibility at the moment, so we could always revisit the question later.

bzinberg:
Just re-read Alex's above comment

> This would introduce two different ways of making groups of independent choices: one in which they are all considered “one choice” (a single trace address), and one in which they are given separate addresses (the Map combinator). It seems the first is strictly less flexible than the second? It seems that long-term, the right thing might be to add better (more efficient) support for the second one, so that it is more usable for this type of use case, than to add the first one as an option.

I think it may make sense to have both options available even long-term.  My thoughts are:
1. Constraining individual entries of the matrix is (AFAICS) fundamentally incompatible with broadcasting.
2. There is a very common use case for the version implemented here: IID noise (where we don't propose directly to the noise value, but instead propose to a RV that serves as the mean of that noise).
3. It's not clear to me that we should expect the overhead of large numbers of traced quantities to be negligible.  If we do, that seems like a separate project that should not block us at least having a broadcast option available in the medium term.

bzinberg:
> Now, I'm much more sympathetic to cases like `fill(1) .+ fill(2)` or `fill(1) .== 1` because there you are already starting with an `Array{T,0}`, and there is some good sense in preserving that. But in the case of `1 .+ 2` I'm more strongly opposed to wrapping its result in an `Array{T,0}`. You are starting with something that has `axes` and supports `getindex` and you end up with the same sort of thing that has `axes` and supports `getindex`. The result is also far more useful.

I would describe the situation as this: There is a type `U` and a function
```julia
f(w::W, x::U, y::U)::U
```
(in our case `f = broadcast`).  There is an implicit conversion from `Float64` to `U`.  We then write `z = f(w, 1.0, 2.0)`.  I think we don't have any reason to expect `z` to have all the same behaviors as `Float64` unless there is also an implicit conversion from `U` to `Float64`.  (Also, `Array{Float64,0}` does support `axes` and `getindex`?)

alex-lew:
> Constraining individual entries of the matrix is (AFAICS) fundamentally incompatible with broadcasting.

@bzinberg Can you explain your thinking here a bit more clearly? Broadcasting is about the shapes of the parameters being broadcast to determine how many random samples to make. I don't see how that is incompatible with posing an inference query like, "suppose the fifth random sample was 2.3"?

> There is a very common use case for the version implemented here: IID noise (where we don't propose directly to the noise value, but instead propose to a RV that serves as the mean of that noise).

You don't propose to the noisy value, but you constrain it (with observations) -- yes? If you only observed part of the matrix, you'd presumably want to constrain only part of the matrix. Of course, there are cases (including Cora?) where you always observe all of the matrix.

> It's not clear to me that we should expect the overhead of large numbers of traced quantities to be negligible. If we do, that seems like a separate project that should not block us at least having a broadcast option available in the medium term.

Hmm, we can definitely discuss further, but I'm relatively confident that we can achieve little or no overhead. Note that in your version, all the quantities are also "traced" -- just not at separate addresses. I imagine we could keep the exact same memory footprint, but add an addressing scheme for addressing elements of iid random arrays. For example, we could implement a `IIDBroadcastedDistribution{T}` generative function type, whose trace (like the `Map` combinator's trace type) is just an array, but whose trace and choicemap types have special implementations of accessor methods (like `get_choices` and `getindex`) that interpret addresses as indices into that array.

bzinberg:
Looking at this a second time, I think this is simpler than I originally realized.

The only thing that's incompatible with broadcasting is broadcasting `x` together with `mu` and `std` -- if we require `x` to have shape exactly equal to `broadcast_shapes(size(mu), size(std))`, then this just just another distribution following the same API as the existing ones -- it just happens to take its parameters in a compact form and do an efficient computation.  And luckily, that is (AFAICS) not something anyone would ever want to do -- I don't see any usefulness of broadcasting the observed value of a normal.  This may have been obvious to you and that's why you were confused :slightly_smiling_face:

Agreed that this distribution should have a separate name like `BroadcastedNormal` (or maybe `BroadcastedIndependentNormal`), at least until we decide how to handle vectorization in Gen (which is longer term and does not need to be decided now).

Very cool about achieving negligible overhead and a more granular addressing scheme!  It's not strictly needed for this PR since we're requiring `x` to have full shape, but I'll definitely be interested to see this develop.

I'll make the changes above and then I think this is pretty safe to merge, WDYT?

bzinberg:
Done implementing https://github.com/probcomp/Gen/pull/102#issuecomment-495963392, I think this is ready to merge.

alex-lew:
Hi @bzinberg -- hmm, I wonder if we are talking past each other a bit? (My fault for mixing multiple concerns together!)

My main concern isn't about broadcasting, but about vector-valued distributions. 

There is a question of whether to have vector-valued distributions, or generative functions with efficient trace implementations that return vectors and expose a list of random choices at indexed addresses.

I'll leave you and Marco to discuss merging this PR (I don't think there's any harm), but I'm a little hazy myself on when this (or any other custom distribution) goes in "Core Gen" vs. whatever project is using it (it's totally kosher for a Gen project to define a custom distribution), and suspect that long-term, the generative-function approach might make more sense than the vector-valued-distribution approach.

bzinberg:
@alex-lew Ah, I see.  I was completely unaware of that concern -- I thought the existing `Gen.mvnormal` distribution was already vector-valued, hence I thought some version of that API decision was already made.

alex-lew:
Yeah, `mvnormal` is the one current example of a vector-valued distribution. Before moving further in that direction, I wanted to express these concerns, as I think that in many cases, it is more natural to think of multivariate distributions as generative functions (with efficient implementations). I would be in favor of making `mvnormal` a generative function, if we could get the overhead down, so that we could build in the ability to condition on some elements of the vector (especially since, as in the IID case, conditioning a multivariate normal is tractable to do exactly).

bzinberg:
Ah, I didn't realize that `Gen.mvnormal` was the only existing example and the design of vector-valued distributions-or-should-they-be-GFs is still up in the air.  I totally get why you wanted to hold off until further discussion is had - and would be great to have you, me and @marcoct meet and discuss this when we have time.  Thanks for sharing your concern!

marcoct (commented):
Nice job, @bzinberg!

A few comments and one request.

It looks like `broadcasted_normal` works for all scalar arguments, subsuming the functionality of `normal`. To reduce the duplicated logic of `normal` and `broadcasted_normal`, I was going to suggest replacing `normal` with the `broadcasted_normal` implementation, but: (1) the behavior of `broadcasted_normal` on all scalars depends on the behavior of `.*` on two zero-dimensional arrays, which seems a little brittle, and (2) there is a performance cost:
```
@time for i=1:10000 broadcasted_normal(0, 1) end
  0.000376 seconds (10.00 k allocations: 937.500 KiB)

@time for i=1:10000 normal(0, 1) end
  0.000067 seconds
```
Note that the performance difference comes from `randn()` versus `randn(())`.

Regarding putting this in Gen vs out of Gen, this vectorized normal is used often enough within inverse-graphics style projects that it seems useful to have it built-in. Let's keep this as part of Gen for now. However, I think it would be natural to include broadcasted versions of the other primitive distributions, and to do this in a way that avoids duplicated logic and lots of boilerplate code. For example, it seems possible to at least auto-generate the test code for each new vectorized distribution. My one request is to  add a vectorized Bernoulli as part of this pull request, in order to help identify how the implementation pattern generalizes to other distributions, and to identify low-hanging fruit opportunities for automation and reduction of boilerplate.

Regarding longer term plans for multivariate distributions in Gen -- it is true that generative functions for certain common multivariate distributions in which random variables are dependent on other variables would be useful. Examples include multivariate normal, and exchangeable distributions like normal-inverse-gamma multivariate distribution (collapsed mean and variance of the normal). However, these are subtly different from one another in their properties (e.g. exchangeability is a special property that the multivariate normal does not possess), and they are both different from the special case handled here in which all variables are independent. I suspect that achieving the same performance using a more flexible generative function will not be possible or at least would require much more complicated code. I think that simple vectorization of primitive distributions as done here is a sufficiently different use case to warrant a distinct implementation strategy.

bzinberg:
**TLDR:** I propose to merge this PR as-is, and I will follow up with the vectorized Bernoulli distribution you suggested in a separate PR since that's less time sensitive.

> It looks like `broadcasted_normal` works for all scalar arguments, subsuming the functionality of `normal`. To reduce the duplicated logic of `normal` and `broadcasted_normal`, I was going to suggest replacing `normal` with the `broadcasted_normal` implementation, but: (1) the behavior of `broadcasted_normal` on all scalars depends on the behavior of `.*` on two zero-dimensional arrays, which seems a little brittle,

Yeah, I agree.  Alex and I discussed this above in https://github.com/probcomp/Gen/pull/102#issuecomment-495480617.  The zero-dimensional case honestly is a pretty significant Julia issue in my opinion, and I participated in the discussion at some length in https://github.com/JuliaLang/julia/issues/28866.  I slightly prefer keeping the separate name for `broadcasted_normal`, because of (i) Julia's inconsistent conflation of scalars and zero-dimensional arrays, and (ii) makes it less of a breaking change if/when in the future we decide to change the design of multivariate distributions.

> and (2) there is a performance cost:
> 
> ```
> @time for i=1:10000 broadcasted_normal(0, 1) end
>   0.000376 seconds (10.00 k allocations: 937.500 KiB)
> 
> @time for i=1:10000 normal(0, 1) end
>   0.000067 seconds
> ```
>
> Note that the performance difference comes from `randn()` versus `randn(())`.

Nice catch!  Didn't realize.  (I wonder why there's such a difference...)

> Regarding putting this in Gen vs out of Gen, this vectorized normal is used often enough within inverse-graphics style projects that it seems useful to have it built-in. Let's keep this as part of Gen for now. However, I think it would be natural to include broadcasted versions of the other primitive distributions, and to do this in a way that avoids duplicated logic and lots of boilerplate code.

Agreed with all the above.

> For example, it seems possible to at least auto-generate the test code for each new vectorized distribution.

Cool idea!

> My one request is to add a vectorized Bernoulli as part of this pull request, in order to help identify how the implementation pattern generalizes to other distributions, and to identify low-hanging fruit opportunities for automation and reduction of boilerplate.

**[Response requested]** I would prefer to do this in a separate PR, because the current content of this PR blocks my Cora work (unless I patch it in, but seems like that's not necessary at this point).  I will do a vectorized Bernoulli PR as you suggested but Cora is more time sensitive at the moment.  Sound ok?

> Regarding longer term plans for multivariate distributions in Gen -- it is true that generative functions for certain common multivariate distributions in which random variables are dependent on other variables would be useful. Examples include multivariate normal, and exchangeable distributions like normal-inverse-gamma multivariate distribution (collapsed mean and variance of the normal). However, these are subtly different from one another in their properties (e.g. exchangeability is a special property that the multivariate normal does not possess), and they are both different from the special case handled here in which all variables are independent. I suspect that achieving the same performance using a more flexible generative function will not be possible or at least would require much more complicated code. I think that simple vectorization of primitive distributions as done here is a sufficiently different use case to warrant a distinct implementation strategy.

Definitely interested in discussing this more.

bzinberg:
BTW, regarding the performance differences, it looks like the the difference comes not from the sampling itself but from the packaging of scalars into 0D vectors:

```
julia> @time for i=1:100_000 randn(()) end
  0.004398 seconds (100.00 k allocations: 9.155 MiB)

julia> @time for i=1:100_000 randn() end
  0.000458 seconds

julia> @time for i=1:100_000 fill(randn()) end
  0.005900 seconds (100.00 k allocations: 9.155 MiB, 38.47% gc time)
```

alex-lew:
Sounds good :) 

> Regarding longer term plans for multivariate distributions in Gen -- it is true that generative functions for certain common multivariate distributions in which random variables are dependent on other variables would be useful. Examples include multivariate normal, and exchangeable distributions like normal-inverse-gamma multivariate distribution (collapsed mean and variance of the normal). However, these are subtly different from one another in their properties (e.g. exchangeability is a special property that the multivariate normal does not possess), and they are both different from the special case handled here in which all variables are independent. I suspect that achieving the same performance using a more flexible generative function will not be possible or at least would require much more complicated code.

Interesting. Not an urgent discussion to have now, but I'd have thought that an `IndependentChoices` generative function (parameterized by a scalar-valued distribution) could use essentially the same logic (and therefore achieve the same performance) as a vector-valued distribution. But it's certainly possible I'm missing something.

I thought such a feature might be useful even in the (conditionally) independent case. Broadly, the two use cases I imagine are:

1. Although the vector of variables are themselves independent, a downstream consumer of the vector makes them conditionally _dependent_, given an observation of some downstream variable (e.g., `normal(sum(xs), 1)`).  In this case, a _proposal_ may wish to propose individually to elements of the vector, rather than have to propose to every element at once. If the vector is a distribution, this can only be done with involution MH.

2. Suppose we draw a vector of (latent) parameters from a multivariate normal, then use them to parameterize a vector of conditionally independent draws from some 1-parameter distribution. Observing only some of the variables allows us (via MH or importance sampling, e.g.) to make inferences about the others.

If it really isn't not possible to do this with no performance overhead, I agree that it fits Gen's philosophy to provide both ways of doing it--a more performant way for when you are happy to think of an entire vector as a single choice, and a slightly less performant way for when you need the elements to be separately addressed. (And again, we needn't finish this conversation now.)

marcoct:
@alex-lew Your suggestion of `IndependentChoices` generative function constructor, which takes in a distribution, is very interesting. 

Regarding performance, the GFI methods for the resulting generative function would need to be able to accept a more general class of arguments than the vectorized distribution version. For example, its `regenerate` method would need to handle arbitrary subsets of indices inside selections, and its `generate` would need to handle arbitrary constraints. A key challenge is to see whether we can dispatch to special code that is optimized for the special case handled by the vectorized distribution handles where all addresses are selected / constrained. For selections, this is possible using the `AllSelection`. For constraints, this could be plausibly implemented by introducing a special type of choice map like a `DenseVectorChoiceMap` that indicates, in its type, that it represents a dense array of addresses. That's possible, but would involve introducing another type that users would need to know about. One way around introducing another user-facing type would be to define a special behavior for choice maps that contain Julia `Array`s as their values. Then, a user could populate the constraints choice map with `:v => Float64[1.5, 2.3, 3.143]` and this would get automatically get converted to a `DenseVectorChoiceMap(Float64[1.5, 2.3, 3.143])` before being passed to the `generate` method of the`IndependentChoices` generative function. One would then also need to ensure that when simulating or proposing complete traces from an `IndependentChoices` generative function, the fact that that it a full dense set of addresses needs to be maintained. This should also be possible. But both of these require some careful thinking about the more general problem of how to annotate choice maps with type information that conveys extra information about their structure, which is also something that the Static modeling language does. These two cases should probably be considered together, in the context of finding a principled solution to this more general problem.

Another potential performance issue is that in order to efficiently implement the GFI, the trace of an `IndependentChoices` would need to be a persistent data structure (e.g. backed by a `PersistentVector` and not a simple `Array`).

alex-lew:
@marcoct Thanks for writing that out -- that all makes sense. Yes, I had imagined dispatching on whether the argument was a full vector or not.

Regarding:
> Another potential performance issue is that in order to efficiently implement the GFI, the trace of an IndependentChoices would need to be a persistent data structure (e.g. backed by a PersistentVector and not a simple Array).

Currently, are the arrays stored in traces (of `mvnormal`, e.g.) immutable? If I write an involution proposal that extracts an `mvnormal` and changes pieces of it, then propose the changed vector, might I have accidentally changed the original trace?

marcoct:
> Currently, are the arrays stored in traces (of `mvnormal`, e.g.) immutable? If I write an involution proposal that extracts an `mvnormal` and changes pieces of it, then propose the changed vector, might I have accidentally changed the original trace?

They are currently not immutable. They are currently `Vector{Float64}`s.  Yes, it would probably be better if they were wrapped in an immutable type that does not provide setter methods.

bzinberg:
Marco and I spoke about this, and agreed that I'll patch this distribution into Cora for now, and will attempt over the next 2 weeks to do the request in https://github.com/probcomp/Gen/pull/102#pullrequestreview-242751246 (depends how much time I am able to spend on it).  If it takes longer than 2 weeks then would be better to merge as-is.

alex-lew:
Discussions on broadcasting from two other Julia PPLs, if we revisit this in the future:
https://github.com/cscherrer/Soss.jl/issues/14
https://github.com/TuringLang/Turing.jl/issues/476

PR #107, opened 2019-06-15
Allow use of docstrings on gen functions

### LINKED ISSUE #105: Provide a way to attach docstrings to `@gen`ed functions
It would be nice for an expression like

```julia
"""
Increment the argument by 1.
"""
@gen function f(x)
 x + 1
end
```

to attach the given docstring to the (macro-wrapped) function.  It currently produces the sensible error

```
'@gen' not documentable. See 'Base.@__doc__' docs for details.
```

marcoct:
Fixes https://github.com/probcomp/Gen/issues/105

### DIFF
--- a/src/dsl/dynamic.jl
+++ b/src/dsl/dynamic.jl
@@ -8,19 +8,21 @@ function make_dynamic_gen_function(name, args, body, return_type, annotations)
     julia_fn_defn = Expr(:function,
         Expr(:call, esc(julia_fn_name), gf_args...),
         esc(body))
-
-    # create generator and assign it a name
     arg_types = map((arg) -> esc(arg.typ), args)
     has_argument_grads = map(
         (arg) -> (DSL_ARG_GRAD_ANNOTATION in arg.annotations), args)
     accepts_output_grad = DSL_RET_GRAD_ANNOTATION in annotations
-    Expr(:block,
-        julia_fn_defn,
-        Expr(:(=), 
-            esc(name),
-            Expr(:call, :DynamicDSLFunction,
-                quote Type[$(arg_types...)] end,
-                julia_fn_defn,
-                has_argument_grads, esc(return_type),
-                accepts_output_grad)))
+
+    quote
+        # first define the underlying Julia function
+        $julia_fn_defn
+
+        # now wrap it in a DynamicDSLFunction value
+        Core.@__doc__ $(esc(name)) = DynamicDSLFunction(
+            Type[$(arg_types...)],
+            $(esc(julia_fn_name)),
+            $has_argument_grads,
+            $(esc(return_type)),
+            $accepts_output_grad)
+    end
 end
--- a/test/dynamic_dsl.jl
+++ b/test/dynamic_dsl.jl
@@ -456,12 +456,18 @@ end
 end
 
 @testset "docstrings" begin
-    @doc "my documentation" (@gen function foo(x)
-        return x + 1
-    end)
+
+    """
+    my documentation
+    """
+    @gen function foo(x)
+            return x + 1
+        end
+
     io = IOBuffer()
     print(io, @doc foo)
     @test String(take!(io)) == "my documentation\n"
+
 end
 
 end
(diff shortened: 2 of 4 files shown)


src/dsl/dynamic.jl
```diff
@@ -8,19 +8,21 @@ function make_dynamic_gen_function(name, args, body, return_type, annotations)
...
-        julia_fn_defn,
-        Expr(:(=), 
-            esc(name),
-            Expr(:call, :DynamicDSLFunction,
-                quote Type[$(arg_types...)] end,
-                julia_fn_defn,
-                has_argument_grads, esc(return_type),
-                accepts_output_grad)))
+
+    quote
```
bzinberg, line 16: Checking my understanding: Does this quoted bit get recursively parsed, where the recursive parsing is needed because the `Expr(:call, ...)` on the left does not recursively expand macros? (And in our case we need to recursively expand the macro `Core.@__doc__`)

test/dynamic_dsl.jl
```diff
@@ -455,4 +455,13 @@ end
             logpdf(normal, x, 0, 1) + logpdf(normal, y, 0, 2))
 end
 
+@testset "docstrings" begin
+    @doc "my documentation" (@gen function foo(x)
```
bzinberg: Is the syntax
```julia
"""
my documentation
"""
@gen function foo(x)
  return x + 1
end
```
supported?

bzinberg (commented):
Thanks Marco!  Left a couple questions.

src/dsl/dynamic.jl
marcoct: The quote isn't needed. It would have been possible to construct the AST directly as before. In this commit, I shifted to using `quote` because (i) it gives more readable cleaner code, and (ii) when I tried to put `Core.@__doc__` into the AST (it involves creating a `Expr(:macrocall, <macroname>, ...)` AST node (see https://docs.julialang.org/en/v1/devdocs/ast/index.html#Macros-1) it didn't work immediately, so it just seemed simpler to use `quote .. end`, in which case it's Julia's responsibility to construct the properly formed AST.

marcoct:
Note that it is a `GenerativeFunction` value that is documented and not a Julia function or a Julia type. But Julia lets you document arbitrary values apparently, so it all still works. That is, when you write

```julia
@gen function foo() end
```
`foo` is a `DynamicDSLFunction <: GenerativeFunction`, not a Julia function.

test/dynamic_dsl.jl
marcoct: Yes. I also just committed a change to the test that documents it using that syntax instead.

src/dsl/dynamic.jl
```diff
@@ -8,19 +8,21 @@ function make_dynamic_gen_function(name, args, body, return_type, annotations)
     julia_fn_defn = Expr(:function,
         Expr(:call, esc(julia_fn_name), gf_args...),
         esc(body))
```
bzinberg: e

src/dsl/dynamic.jl
bzinberg: > it involves creating a `Expr(:macrocall, <macroname>, ...`

Ah, got it.

bzinberg:
> Note that it is a `GenerativeFunction` value that is documented and not a Julia function or a Julia type. But Julia lets you document arbitrary values apparently, so it all still works. That is, when you write
> 
> ```julia
> @gen function foo() end
> ```
> 
> `foo` is a `DynamicDSLFunction <: GenerativeFunction`, not a Julia function.

Yeah, that's how it should be i.m.o.!  The thing that ultimately gets assigned to the name `foo` is the thing that I want to attach a docstring to.

BTW, this wrinkle exists in Python too -- if you wrap your function in a decorator, its docstring disappears.  That's what [`functools.wraps`](https://docs.python.org/2/library/functools.html#functools.wraps) attempts to conveniently solve.

bzinberg (approved):
LGTM!

PR #116, opened 2019-08-01
Generate documentation for the exponential distribution.

ghost:
https://probcomp.github.io/Gen/dev/ref/distributions/

Docs for the exponential distribution seem to be missing.


alex-lew:
Thanks!

PR #120, opened 2019-08-22
Allow `broadcasted_normal` to accept `AbstractArray`s as arguments instead of requiring `Array`s

bzinberg:
This allows us to use things like `PermutedDimsArray` as arguments, which is useful for me.


zane (approved):
:shipit:

PR #130, opened 2019-10-12
Update parsing to allow returning @trace in a static DSL

### LINKED ISSUE #87: Cannot return @trace in Static DSL
```julia
julia> @gen (static) function f()
           return @trace(normal(0,1), :foo)
           end
ERROR: LoadError: Syntax error when parsing static DSL function at return #= REPL[3]:2 =# @trace(normal(0, 1), :foo). 
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] static_dsl_syntax_error(::Expr, ::String) at /scratch/fsaad/.julia/packages/Gen/n7UPH/src/dsl/static.jl:6 (repeats 2 times)
 [3] parse_static_dsl_function_body!(::Array{Expr,1}, ::Dict{Symbol,Symbol}, ::Expr) at /scratch/fsaad/.julia/packages/Gen/n7UPH/src/dsl/static.jl:235
 [4] make_static_gen_function(::Symbol, ::Array{Gen.Argument,1}, ::Expr, ::Symbol, ::Set{Symbol}) at /scratch/fsaad/.julia/packages/Gen/n7UPH/src/dsl/static.jl:254
 [5] parse_gen_function(::Expr, ::Set{Symbol}) at /scratch/fsaad/.julia/packages/Gen/n7UPH/src/dsl/dsl.jl:75
 [6] @gen(::LineNumberNode, ::Module, ::Any, ::Any) at /scratch/fsaad/.julia/packages/Gen/n7UPH/src/dsl/dsl.jl:87
in expression starting at REPL[3]:1
```

Returning an `@trace` value works in the DynamicDSL.

nmbrgts:
This PR updates how returns are parsed to allow returning a trace directly. To do this, I modified `parse_return!` to call `parse_trace_expr!` to handle the returned expression. I've tested this out with the example given in #87 and it produces comparable code. Let me know if I'm missing anything.

Closes #87

### DIFF
--- a/src/dsl/static.jl
+++ b/src/dsl/static.jl
@@ -194,7 +194,7 @@ function parse_trace_line!(stmts::Vector{Expr}, bindings, line::Expr)
     end
 end
 
-# return foo (must be a symbol)
+# return foo (must be a symbol) or return @trace(..)
 function parse_return!(stmts::Vector{Expr}, bindings, line::Expr)
     if line.head != :return
         return false
@@ -202,7 +202,9 @@ function parse_return!(stmts::Vector{Expr}, bindings, line::Expr)
     if isa(line.args[1], Expr) && line.args[1].head == :macrocall
         var = gensym()
         typ = QuoteNode(Any)
-        res = parse_trace_expr!(stmts, bindings, var, line.args[1], typ)
+        if !parse_trace_expr!(stmts, bindings, var, line.args[1], typ)
+            return false # the right-hand-side is a macro but not a valid `@trace` expression
+        end
     elseif isa(line.args[1], Symbol)
         var = line.args[1]
     else
--- a/src/dsl/static.jl
+++ b/src/dsl/static.jl
@@ -115,6 +115,7 @@ end
 choice_or_call_at(gen_fn::GenerativeFunction, addr_typ) = call_at(gen_fn, addr_typ)
 choice_or_call_at(dist::Distribution, addr_typ) = choice_at(dist, addr_typ)
 
+       # parse_trace_expr!(stmts::Vector{Expr}, bindings, name::Symbol("##XYZ"), line::Epr(not :(=)), typ)
 function parse_trace_expr!(stmts, bindings, name, addr_expr, typ)
     # NOTE: typ is unused
     if !(isa(addr_expr, Expr) && addr_expr.head == :macrocall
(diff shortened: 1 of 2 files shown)


marcoct:
Thanks for the contribution @nmbrgts ! It would be great to have a unit test for this functionality that tests the resulting intermediate representation, in `test/static_dsl.jl`. I imagine that it would be a short test, like https://github.com/probcomp/Gen/blob/master/test/static_dsl.jl#L332-L344, and it could go right after that one. Thanks again.

nmbrgts:
Thanks for the feedback @marcoct! I might not get a chance to take a look at adding the unit test until the weekend.

nmbrgts:
I took me a little longer than I expected to get back to this. I've added a test set that compares the return nodes of two static DSLs, one that directly returns a trace and one that returns a trace using a variable. Let me know if there is anything else that can be tested to cover this behavior.

src/dsl/static.jl
```diff
@@ -195,10 +196,18 @@ end
...
 # return foo (must be a symbol)
 function parse_return!(stmts::Vector{Expr}, bindings, line::Expr)
-    if line.head != :return || !isa(line.args[1], Symbol)
+    if line.head != :return
+        return false
+    end
+    if isa(line.args[1], Expr) && line.args[1].head == :macrocall
+        var = gensym()
+        typ = QuoteNode(Any)
+        res = parse_trace_expr!(stmts, bindings, var, line.args[1], typ)
```
marcoct:
```suggestion
        if !parse_trace_expr!(stmts, bindings, var, line.args[1], typ)
            return false # the right-hand-side is a macro but not a valid `@trace` expression
        end
```

src/dsl/static.jl
```diff
@@ -195,10 +196,18 @@ end
 
 # return foo (must be a symbol)
```
marcoct:
```suggestion
# return foo (must be a symbol) or return @trace(..)
```

marcoct (changes requested):
There is just one suggested change that handles the case when the a different macro besides `@trace` is used.

src/dsl/static.jl
nmbrgts: Thanks for the correction! I didn't realize that the code here was eating the false return value on failure.

nmbrgts:
Thanks! I hadn't seen that GitHub feature before. It's pretty convenient!

marcoct:
Thanks @nmbrgts!

PR #149, opened 2019-11-08
MultivariateNormal logpdf gradients w.r.t mean and covariance

SamWitty:
This pull request implements gradients of the multivariate normal distribution's log likelihood with respect to its mean vector and covariance matrix. This enables gradient-based inference in models where a MultivariateNormal's mean vector or covariance matrix is a function of some latent random variable, e.g. Gaussian Process Latent Variable Models.

### DIFF
--- a/test/modeling_library/distributions.jl
+++ b/test/modeling_library/distributions.jl
@@ -172,8 +172,8 @@ end
     @test isapprox(actual[1][2], finite_diff_vec(f, args, 1, 2, dx))
     @test isapprox(actual[2][1], finite_diff_vec(f, args, 2, 1, dx))
     @test isapprox(actual[2][2], finite_diff_vec(f, args, 2, 2, dx))
-
-    @test actual[3] === nothing # not yet implemented
+    @test isapprox(actual[3][1, 1], finite_diff_mat_sym(f, args, 3, 1, 1, dx))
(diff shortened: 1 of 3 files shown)


test/modeling_library/distributions.jl
```diff
@@ -172,8 +172,8 @@ end
     @test isapprox(actual[1][2], finite_diff_vec(f, args, 1, 2, dx))
     @test isapprox(actual[2][1], finite_diff_vec(f, args, 2, 1, dx))
     @test isapprox(actual[2][2], finite_diff_vec(f, args, 2, 2, dx))
-
-    @test actual[3] === nothing # not yet implemented
+    @test isapprox(actual[3][1, 1], finite_diff_mat_sym(f, args, 3, 1, 1, dx))
+    @test isapprox(actual[3][1, 2], finite_diff_mat_sym(f, args, 3, 1, 2, dx))
```
marcoct, line 176: Seems like it would make sense to check the other 2 entries in the matrix too?

marcoct: Thanks a lot Sam -- this is great. I just had one question about the test.

SamWitty: Good catch. Just added below.

marcoct:
Looks great to me!

PR #171, opened 2019-12-23
Removed deprecated use of lgamma(X::real) and lbeta(X::real)

SamWitty:


bzinberg (approved):
Thanks @SamWitty.

Just to confirm -- testing strategy is unit tests still pass, and you've also checked that this is the only instance of `lgamma` in the repo?

If so, looks good to merge (and according to @marcoct's recent comment, merging without his review on a small fix like this is fine)

SamWitty:
Thanks for the review @bzinberg. 

My understanding re: unit tests is that all tests are run automatically for each pull request. Is this correct/sufficient? I didn't explicitly check the test suite to make sure this method was covered.

It looks like there are 3 other instances of `lgamma` in the repo. I'll change those as well and combine them into this pull request.

Thanks!

bzinberg:
Great, thanks Sam.  Yes, the existing unit tests are run automatically for each pull request.  Whether that's sufficient is the question that I was raising -- in this case I think it obviously is, since the change is just a name update.

Because there is a general (or is it just me) temptation to rely on existing unit tests without deliberately asking, "can I meaningfully add more tests to cover these changes?", I think it's a good idea to record (or as a reviewer, ask about) some form of testing strategy in the PR description, even if the strategy is "as long as the CI passes, that's coverage enough."  That way it's harder to gloss over testing.

SamWitty:
@bzinberg, it looks like there wasn't an existing test for the inv_gamma distribution, so I added a minimal one.

On CI: I made the mistake of making these changes in the GitHub web interface (given how simple the changes were), which means that each file change is its own individual commit. The build test are not passing for the newer commits, I suspect because 0efedf0 changes the import and 3b3c44b, 93b5ae1, etc. change the usage of `lgamma`. Am I correct that this is why the builds are failing? Is there a way to run the build tests with a build that includes all of the commits together? I'm happy to just scrap the pull request and redo it fresh in a single commit if that would be easier.

Sorry this simple change is requiring so much back and forth...

bzinberg:
The back and forth is a good thing!  Thank you for helping improve Gen and, more selfishly on my part, for making my review worthwhile by engaging with me :slightly_smiling_face:

For the question of whether a bunch of tiny commits are ok: I think it's totally legit -- they are coherent units of change and have good descriptions (I wish everyone would do this...).  Having all the name changes in one commit would have also been fine.

On CI, it's certainly not necessary to have every commit pass CI, in fact, in test-driven development (e.g. [here](https://github.com/probcomp/Gen/pull/170/commits)) the strategy is to first intentionally commit a failing test, such that the bug will be fixed when the test passes.  As long as the latest commit on the branch passes CI when it's merged into master, all is well.

You can check the build logs including error message by clicking Details in the Travis box below and then following the link there to "The build" -- for example, [this one](https://travis-ci.com/probcomp/Gen/jobs/269845289).  The error seems to be that `SpecialFunctions.loggamma` doesn't exist.  I think bumping the version of that dependency will hopefully fix the problem, since the rename seems to have been introduced in 0.8.0 and we're still on 0.7.2.

bzinberg:
Welp... looks like there may be a dependency bug in some combination of SpecialFunctions or Distributions:
```
(Gen) pkg> add SpecialFunctions@0.9.0
 Resolving package versions...
ERROR: Unsatisfiable requirements detected for package SpecialFunctions [276daf66]:
 SpecialFunctions [276daf66] log:
 ├─possible versions are: [0.7.0-0.7.2, 0.8.0, 0.9.0] or uninstalled
 ├─restricted to versions 0.9.0 by an explicit requirement, leaving only versions 0.9.0
 └─restricted by compatibility requirements with Distributions [31c24e10] to versions: 0.7.0-0.7.2 — no versions left
   └─Distributions [31c24e10] log:
     ├─possible versions are: [0.16.0-0.16.4, 0.17.0, 0.18.0, 0.19.1-0.19.2, 0.20.0, 0.21.0-0.21.3, 0.21.5-0.21.11] or uninstalled
     └─restricted to versions 0.21.5 by an explicit requirement, leaving only versions 0.21.5
```

bzinberg:
Applied the fix in https://github.com/JuliaMath/SpecialFunctions.jl/issues/182 (remove Distributions, update SpecialFunctions, re-add Distributions) and now the error is fixed for me locally -- let's see how it goes on CI.

bzinberg:
@SamWitty, I think this PR is ready to merge.  Any further comments before doing so?

SamWitty:
No further comments on my end! Thanks for following up and looking into why the build tests were failing. Also, thanks for the general tips and clarification about CI and test-driven development.

bzinberg:
Sure thing - thanks for sending the PR.

PR #172, opened 2019-12-24
auxiliary state accessing for static dsl

georgematheos:
I’ve implemented the auxiliary trace accessing for the static DSL, according to the functionality specified in #153.  I copied over the functionality tests from #153 as well; everything appears to be passing.

A couple little “hygiene” decisions arose when doing this.   For the aux state to be accessible, we need to add a `get_subtrace` method and a `getindex` method; I’ve written the code so that these are accessible via `Gen.static_get_subtrace` and `Gen.static_getindex`.  I followed the Juliaism used throughout the static trace code to use `Val` objects for multiple dispatch for the new generated functions I wrote, so these take `Val`-wrapped addresses as inputs.

Currently, I am making these functions easily accessible in the Gen namespace.  Alternatively, we could use `gensym()` to create a random identifier for `static_get_subtrace` and `static_getindex`, or we could rename to `_static_get_subtrace` and `_static_getindex` to denote that they aren’t intended to be outward-facing.  My inclination was to use `gensym`, but I saw that the `static_get_submap` functions, etc., are made accessible as part of the Gen namespace, so tried to follow suit with this.  Let me know what you think would be better.

Also, as was noted in an earlier discussion, it feels a bit like implementing the choicemap code separately from the `getindex(trace, …)` code is a little bit redundant.  Not that hard to do separately so probably not a big deal, but if we think of a more elegant implementation, that would be nice.

And of course, let me know if anything looks off in the code, or there’s anything that can be improved.  Happy holidays!

### DIFF
--- a/src/modeling_library/call_at/call_at.jl
+++ b/src/modeling_library/call_at/call_at.jl
@@ -27,6 +27,22 @@ struct CallAtTrace <: Trace
     key::Any
 end
 
+function Base.getindex(trace::CallAtTrace, addr::Pair)
+    first, rest = addr
+    if first == trace.key
+        return trace.subtrace[rest]
+    else
+        error("Address prefix $addr not found.")
+    end
+end
+function Base.getindex(trace::CallAtTrace, addr)
+    if addr == trace.key
+        return trace.subtrace[]
+    else
+        error("Address $addr not found.")
+    end
+end
+
 get_args(trace::CallAtTrace) = (get_args(trace.subtrace)..., trace.key)
 get_retval(trace::CallAtTrace) = get_retval(trace.subtrace)
 get_score(trace::CallAtTrace) = get_score(trace.subtrace)
--- a/src/static_ir/trace.jl
+++ b/src/static_ir/trace.jl
@@ -36,8 +36,19 @@ get_submap(choices::StaticIRTraceAssmt, addr::Pair) = _get_submap(choices, addr)
 
 abstract type StaticIRTrace <: Trace end
 
+function static_get_subtrace(trace::StaticIRTrace, addr)
+    error("Not implemented")
+end
+
+static_haskey(trace::StaticIRTrace, ::Val) = false
+Base.haskey(trace::StaticIRTrace, key) = Gen.static_haskey(trace, Val(key))
+
 function Base.getindex(trace::StaticIRTrace, addr)
-    get_choices(trace)[addr]
+    Gen.static_getindex(trace, Val(addr))
+end
+function Base.getindex(trace::StaticIRTrace, addr::Pair)
+    first, rest = addr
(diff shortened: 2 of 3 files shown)


georgematheos:
Update: it might be worth holding off on this until we figure out #173.

marcoct:
Nice job @georgematheos :) A couple questions / comments:

You mentioned the Map combinator, but I didn't the Map-combinator-specific code being modified. Is it just the VectorTrace code that was modified? If so, would it not apply to the Unfold combinator as well, which uses the same trace type?

> A couple little “hygiene” decisions arose when doing this. For the aux state to be accessible, we need to add a get_subtrace method and a getindex method; I’ve written the code so that these are accessible via Gen.static_get_subtrace and Gen.static_getindex. I followed the Juliaism used throughout the static trace code to use Val objects for multiple dispatch for the new generated functions I wrote, so these take Val-wrapped addresses as inputs.

That makes sense. These could probably be used for code simplifications in the Static IR GFI implementations at some point (e.g. https://github.com/probcomp/Gen/blob/master/src/static_ir/update.jl#L315-L316).

> Currently, I am making these functions easily accessible in the Gen namespace. Alternatively, we could use gensym() to create a random identifier for static_get_subtrace and static_getindex, or we could rename to _static_get_subtrace and _static_getindex to denote that they aren’t intended to be outward-facing. My inclination was to use gensym, but I saw that the static_get_submap functions, etc., are made accessible as part of the Gen namespace, so tried to follow suit with this. Let me know what you think would be better.

Elsewhere in Gen, functions starting with _ are used to indicate internal functions, so that would be great here too. Also, `static_get_submap` being exported is intended to allow users to write inference programs with that make their address schemas available as part of types. But it seems like some of the need for that is obviated by this PR.

> Also, as was noted in an earlier discussion, it feels a bit like implementing the choicemap code separately from the getindex(trace, …) code is a little bit redundant. Not that hard to do separately so probably not a big deal, but if we think of a more elegant implementation, that would be nice.

Agreed.

georgematheos:
Hi @marcoct ; thanks for the review!  Yes, you are right that since I have modified the vector trace code both the Map and Unfold combinators will have the performance improvement; I misspoke.  It's only the Recurse one that will still need work.

I can certainly change to using underscored function names for `static_get_subtrace` and `static_getindex`; should I make that change on this pull request or on the performance one extending this that I pushed yesterday?  And I agree that the `static_get_subtrace` (or now `_static_get_subtrace`) could simplify code like what you linked to.

marcoct:
> I can certainly change to using underscored function names for static_get_subtrace and static_getindex; should I make that change on this pull request or on the performance one extending this that I pushed yesterday?

The other PR extends this one, so making changes to, and merging that one, makes sense right?

Issue #173, opened 2019-12-30
Static trace indexing very slow due to runtime value type dispatch


georgematheos:
I have been spending some time profiling Gen inference code, and one thing I've found is that `getindex` calls for static traces (/static choicemaps) are extremely slow.  After poking around for a bit, I have come to believe this is because the current mechanism used for indexing relies on Julia type dispatch, but due to shortcomings in the Julia compiler, it is unable to do this dispatch at compile time.

Here's an example of a little test I have run:
```julia
using Gen

@gen (static) function static()
    a = @trace(normal(0, 1), :a)
end
Gen.load_generated_functions()

@gen function dynamic()
    return @trace(normal(0, 1), :a)
end

function many_lookups(tr)
    for _=1:1e5
        x = tr[:a]
    end
end

println("dynamic:")
dynamic_tr = simulate(dynamic, ())
@time many_lookups(dynamic_tr)

println("static:")
static_tr = simulate(static, ())
@time many_lookups(static_tr)
```
This yields output:
```
dynamic:
  0.085845 seconds (402.20 k allocations: 18.410 MiB)
static:
  1.271537 seconds (419.16 k allocations: 19.311 MiB, 0.97% gc time)
```
The dynamic lookups run about 15x faster than the static ones.

Profiling the static lookup (after JIT compilation has run):
```
25218 ...erformance/trace_lookup2.jl:15; many_lookups(::Main.TraceLookup2.var"##Sta...
 25216 ...pqN/src/static_ir/trace.jl:40; getindex
  25187 ...en/VlpqN/src/choice_map.jl:75; getindex
   25184 ...pqN/src/static_ir/trace.jl:18; get_value(::Gen.StaticIRTraceAssmt{Main....
    24812 ./essentials.jl:695; Val(::Symbol)
    201   ./none:0; static_get_value
     201 ./Base.jl:20; getproperty
```
The way the static lookups currently works is that to lookup `addr`, we create `Val{addr}` and then call `static_get_value(choices, Val(addr))`.  The profiler shows that most of the time is spent right after we just wrapped the symbol in a `Val`, which I'm interpreting to mean that most of the time is spent with julia searching for the proper method of `static_get_value` for the given `Val{:a}` type.  In other words, this method dispatch is happening at runtime, not compile-time.

[...]

PR #177, opened 2020-01-08
20200107 georgematheos fast static lookup

georgematheos:
This should resolve issue #173 in most cases.  This is also an extension of #172 
which implements aux state accessing for the static DSL

As I mentioned in the issue, the static lookups are very slow when the value-type
method dispatch has to happen at runtime.  For flat lookups this is solved by inlining
`getindex` and other methods; if we have a nested structure where we
pass through a `Map` combinator,
we need the `Map` trace type to realize that the subtraces are all of the static
trace type, so the Julia compiler can properly propagate the value types at compile time.

To give a sense of the effect this PR has, we get 70x or greater speedup for static lookups (and a bit better even when passing through the map combinator).  For a more sophisticated and realistic
entity-mention model I am using for Gen profiling in general, I found that this led
to a 1.65x speedup for running gibbs sampling on the whole model, so the speedup
for the lookups makes a serious difference to performance for MCMC in general.

To give a sense of timing results, before the PR here is a timing output:
```
dynamic function lookup:
  0.118068 seconds (318.75 k allocations: 17.701 MiB)
static function lookup:
  6.757818 seconds (315.52 k allocations: 17.630 MiB)
dynamic function lookup passing through map combinator:
  0.483430 seconds (5.31 M allocations: 124.247 MiB, 12.71% gc time)
static function lookup passing through map combinator:
 13.821885 seconds (5.33 M allocations: 125.474 MiB, 0.43% gc time)
dynamic nested in static:
  7.354814 seconds (2.36 M allocations: 65.574 MiB)
static nested in static:
 13.350080 seconds (2.34 M allocations: 64.496 MiB)
static nested in dynamic:
  6.839463 seconds (3.31 M allocations: 78.605 MiB, 0.42% gc time)
```

After this change, the output is
```
dynamic function lookup:
  0.128471 seconds (336.30 k allocations: 18.604 MiB)
static function lookup:
  0.094528 seconds (317.89 k allocations: 17.739 MiB)
dynamic function lookup passing through map combinator:
  0.429125 seconds (5.37 M allocations: 127.137 MiB, 7.91% gc time)
static function lookup passing through map combinator:
  0.158622 seconds (3.38 M allocations: 97.197 MiB, 20.67% gc time)
dynamic nested in static:
  0.195995 seconds (2.37 M allocations: 65.986 MiB, 14.50% gc time)
static nested in static:
  0.132090 seconds (2.35 M allocations: 65.036 MiB, 22.17% gc time)
static nested in dynamic:
  7.036895 seconds (3.31 M allocations: 78.593 MiB, 0.39% gc time)
```

While I have updated the `Map` combinator I have not looked at the `Recurse` or `Unfold`
ones; I'll make an issue noting that we should make sure the type information
propagates correctly through these so we get the proper speedups, and to make sure
everything is inlined properly. For future combinators with subtraces they will also
need to propagate the type information properly.

This PR also does not implement a way to propagate the needed type information through
dynamic functions, so if a dynamic function calls a static function, it will remain
slow to lookup values from the static call through the dynamic trace.  I'm not sure
that there is a good way to do this since I don't see how the subtrace types under
a dynamic function could be known at compile time.  My sense is that almost all performance
critical code will need to avoid the dynamic DSL since it provides no way to propagate
argdiff information or utilize conditional independence, so probably it isn't the
end of the world if dynamic-then-static lookups are slow.  That said, if we want to improve
performance in this case we could implement some mechanism so that the static
trace lookups can use a normal hashmap lookup when it's parent function
is dynamic, and the method-dispatch technique in other cases (since this will
  be faster when it is possible to use).

### DIFF
--- a/src/choice_map.jl
+++ b/src/choice_map.jl
@@ -69,24 +69,24 @@ function Base.isempty(::ChoiceMap)
     true
 end
 
-get_submap(choices::ChoiceMap, addr) = EmptyChoiceMap()
-has_value(choices::ChoiceMap, addr) = false
-get_value(choices::ChoiceMap, addr) = throw(KeyError(addr))
-Base.getindex(choices::ChoiceMap, addr) = get_value(choices, addr)
+@inline get_submap(choices::ChoiceMap, addr) = EmptyChoiceMap()
+@inline has_value(choices::ChoiceMap, addr) = false
+@inline get_value(choices::ChoiceMap, addr) = throw(KeyError(addr))
+@inline Base.getindex(choices::ChoiceMap, addr) = get_value(choices, addr)
 
-function _has_value(choices::T, addr::Pair) where {T <: ChoiceMap}
+@inline function _has_value(choices::T, addr::Pair) where {T <: ChoiceMap}
     (first, rest) = addr
     submap = get_submap(choices, first)
     has_value(submap, rest)
 end
 
-function _get_value(choices::T, addr::Pair) where {T <: ChoiceMap}
+@inline function _get_value(choices::T, addr::Pair) where {T <: ChoiceMap}
     (first, rest) = addr
     submap = get_submap(choices, first)
     get_value(submap, rest)
 end
 
-function _get_submap(choices::T, addr::Pair) where {T <: ChoiceMap}
+@inline function _get_submap(choices::T, addr::Pair) where {T <: ChoiceMap}
     (first, rest) = addr
     submap = get_submap(choices, first)
     get_submap(submap, rest)
@@ -365,28 +365,28 @@ function get_address_schema(::Type{StaticChoiceMap{R,S,T,U}}) where {R,S,T,U}
     StaticAddressSchema(keys)
 end
 
-function Base.isempty(choices::StaticChoiceMap)
+@inline function Base.isempty(choices::StaticChoiceMap)
     choices.isempty
 end
 
 get_values_shallow(choices::StaticChoiceMap) = pairs(choices.leaf_nodes)
 get_submaps_shallow(choices::StaticChoiceMap) = pairs(choices.internal_nodes)
-has_value(choices::StaticChoiceMap, addr::Pair) = _has_value(choices, addr)
-get_value(choices::StaticChoiceMap, addr::Pair) = _get_value(choices, addr)
-get_submap(choices::StaticChoiceMap, addr::Pair) = _get_submap(choices, addr)
+@inline has_value(choices::StaticChoiceMap, addr::Pair) = _has_value(choices, addr)
+@inline get_value(choices::StaticChoiceMap, addr::Pair) = _get_value(choices, addr)
+@inline get_submap(choices::StaticChoiceMap, addr::Pair) = _get_submap(choices, addr)
 
 # NOTE: there is no static_has_value because this is known from the static
 # address schema
 
 ## has_value ##
 
-function has_value(choices::StaticChoiceMap, key::Symbol)
+@inline function has_value(choices::StaticChoiceMap, key::Symbol)
     haskey(choices.leaf_nodes, key)
 end
 
 ## get_submap ##
 
-function get_submap(choices::StaticChoiceMap, key::Symbol)
+@inline function get_submap(choices::StaticChoiceMap, key::Symbol)
     if haskey(choices.internal_nodes, key)
         choices.internal_nodes[key]
     elseif haskey(choices.leaf_nodes, key)
@@ -396,17 +396,17 @@ function get_submap(choices::StaticChoiceMap, key::Symbol)
     end
 end
 
-function static_get_submap(choices::StaticChoiceMap, ::Val{A}) where {A}
+@inline function static_get_submap(choices::StaticChoiceMap, ::Val{A}) where {A}
     choices.internal_nodes[A]
 end
 
 ## get_value ##
 
-function get_value(choices::StaticChoiceMap, key::Symbol)
+@inline function get_value(choices::StaticChoiceMap, key::Symbol)
     choices.leaf_nodes[key]
 end
 
-function static_get_value(choices::StaticChoiceMap, ::Val{A}) where {A}
+@inline function static_get_value(choices::StaticChoiceMap, ::Val{A}) where {A}
     choices.leaf_nodes[A]
 end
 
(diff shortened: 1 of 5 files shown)


georgematheos (commented):
I realize I'd left in some inlines I added from another little experiment I was running; these inlines are not relevant to this PR so I am taking them out.

marcoct:
This is great.

Do the timing results include compilation time (e.g. did you run it twice and time the second run)?

> This PR also does not implement a way to propagate the needed type information through
dynamic functions, so if a dynamic function calls a static function, it will remain
slow to lookup values from the static call through the dynamic trace. I'm not sure
that there is a good way to do this since I don't see how the subtrace types under
a dynamic function could be known at compile time. My sense is that almost all performance
critical code will need to avoid the dynamic DSL since it provides no way to propagate
argdiff information or utilize conditional independence, so probably it isn't the
end of the world if dynamic-then-static lookups are slow. 

I agree. I think it is fine if using DML limits performance. I think that ideally the SML + combinators code would become more mature and less verbose, and more like writing Julia code (e.g. by supporting Julia control flow constructs as we discussed earlier), then it could become the main starting point for users. But even when that happens, I think that keeping the DML around makes sense because its implementation is so simple that it serves as a sort of reference implementation of the GFI.

marcoct:
@georgematheos There are some code examples that can be executed by `examples/run_examples.jl` that use the SML, but that don't get run as part of CI at the moment (they probably should but they aren't currently because I thought they would take too long). I run them periodically when making major changes to make sure they still run at least. Could you check that the ones that involve the SML still run?

```
include("regression/static_mala_hmc.jl")
include("regression/static_map_optimize.jl")
include("regression/static_mh.jl")
include("regression/static_mh_collapsed.jl")
include("gp_structure/incremental.jl")
include("mle/static_mle.jl")
```

georgematheos:
> Do the timing results include compilation time (e.g. did you run it twice and time the second run)?

When I was starting my profile testing, I found that the compilation time for these lookup functions was inconsequential (vs. for running MCMC where it is huge), so I took out the "compiling" timings; I'm gonna take a lunch break but I can run my benchmarks again with this compiling in case after the changes I've made the compile-time is more significant.

> @georgematheos There are some code examples that can be executed by `examples/run_examples.jl` that use the SML, but that don't get run as part of CI at the moment (they probably should but they aren't currently because I thought they would take too long). I run them periodically when making major changes to make sure they still run at least. Could you check that the ones that involve the SML still run?

Can do! I'll run those examples soon and add the results to this thread.

georgematheos:
@marcoct I ran again with the changes, and turns out compilation does make a big difference after the speedups; new results are:

```
dynamic:
  0.170589 seconds (588.31 k allocations: 31.172 MiB, 5.97% gc time)
  0.034875 seconds (5 allocations: 176 bytes)
static:
  0.080942 seconds (317.65 k allocations: 17.724 MiB)
  0.000511 seconds (5 allocations: 176 bytes)
dynamic mappy:
  0.371547 seconds (5.37 M allocations: 127.383 MiB, 5.77% gc time)
  0.256749 seconds (5.00 M allocations: 106.812 MiB, 2.77% gc time)
static mappy:
  0.170101 seconds (3.38 M allocations: 97.136 MiB, 20.72% gc time)
  0.034129 seconds (3.00 M allocations: 76.294 MiB, 17.81% gc time)
dynamic nested:
  0.181373 seconds (2.37 M allocations: 65.924 MiB, 7.64% gc time)
  0.085597 seconds (2.00 M allocations: 45.777 MiB, 6.25% gc time)
static nested:
  0.113685 seconds (2.35 M allocations: 64.944 MiB, 12.60% gc time)
  0.017172 seconds (2.00 M allocations: 45.777 MiB, 30.67% gc time)
dynamic then static:
  6.690246 seconds (3.31 M allocations: 78.501 MiB, 0.19% gc time)
  6.650602 seconds (3.00 M allocations: 61.035 MiB, 0.08% gc time)
```

Notably, it looks like the static lookup is made a lot slower when there are 2 nested static functions.  Still significantly faster than the dynamic lookups though, so maybe this time just comes from unpacking the address pair, etc.

Running this before the improvement we have:
```
dynamic:
  0.154367 seconds (587.98 k allocations: 31.168 MiB)
  0.035434 seconds (5 allocations: 176 bytes)
static:
  5.677942 seconds (315.21 k allocations: 17.612 MiB)
  5.652189 seconds (5 allocations: 176 bytes)
dynamic mappy:
  0.437291 seconds (5.37 M allocations: 127.038 MiB, 6.33% gc time)
  0.314949 seconds (5.00 M allocations: 106.812 MiB, 6.51% gc time)
static mappy:
 11.695737 seconds (5.33 M allocations: 125.410 MiB, 0.20% gc time)
 11.634708 seconds (5.00 M allocations: 106.812 MiB, 0.11% gc time)
dynamic nested:
  5.910022 seconds (2.36 M allocations: 65.507 MiB, 0.23% gc time)
  5.785663 seconds (2.00 M allocations: 45.777 MiB, 0.13% gc time)
static nested:
 11.332550 seconds (2.34 M allocations: 64.401 MiB, 0.14% gc time)
 11.371217 seconds (2.00 M allocations: 45.777 MiB, 0.06% gc time)
dynamic then static:
  5.902021 seconds (3.31 M allocations: 78.509 MiB, 0.25% gc time)
  5.791408 seconds (3.00 M allocations: 61.035 MiB, 0.12% gc time)
```
So for flat static lookups this is actually a 10,000x speedup!  And even passing through the map combinator it's a 300x improvement.



> There are some code examples that can be executed by `examples/run_examples.jl` that use the SML, but that don't get run as part of CI at the moment (they probably should but they aren't currently because I thought they would take too long). I run them periodically when making major changes to make sure they still run at least. Could you check that the ones that involve the SML still run?

I ran all the examples and everything worked (if you want I can paste in the outputs, but there's a lot of logging so I am not planning to).  I did run into a little issue: for both `gp_structure` tests I get a `ERROR: LoadError: LoadError: setindex! not defined for CSV.Column{Float64,Float64}`.  This error is discussed here: https://github.com/JuliaData/CSV.jl/issues/438; using the suggestion from this issue I changed `examples/gp_structure/shared.jl` line 17 from `df = CSV.read("$(@__DIR__)/airline.csv")` to `df = CSV.read("$(@__DIR__)/airline.csv", copycols=true)` and then these tests passed as well.  Perhaps I have a different version of the CSV library.

PR #184, opened 2020-01-31
20200129 georgematheos staticmacros

georgematheos:
For some code I'm working on (the BLOG to Gen project), it will be important to be able to use some macros in the static DSL.

For instance, I want to be able to write code like
```julia
@type T
@gen (static) function create_and_sample_Ts()
  @num T = poisson(5)
  samples = @trace(@sample(T), :samples)
  return samples
end
```
and have it macroexpand into some generative functions that do some useful stuff.

This PR adds support for using macros in a couple places in the static DSL.

### DIFF
--- a/docs/src/ref/modeling.md
+++ b/docs/src/ref/modeling.md
@@ -405,6 +405,34 @@ The functions must also satisfy the following rules:
 
 - Julia control flow constructs (e.g. `if`, `for`, `while`) cannot be used as top-level statements in the function body. Control flow should be implemented inside Julia functions that are called, generative functions that are called such as generative functions produced using [Generative Function Combinators](@ref).
 
+#### Macros in static DSL
+
+Gen includes several macros with special meanings for use within generative functions (`@trace`, `@param`, and `@grad`).
+In addition to these, user-defined macros may be used within the dynamic DSL.  The static DSL also supports user-defined macros,
+so long as they (non-recursively) `macroexpand` into code which makes the line valid under the restrictions of the
+static DSL.  Macros may be used in the static DSL in any of the following places:
+
+- An entire line may be a macro which (non-recursively) `macroexpand`s to a valid line for the static DSL.  For example:
+```julia
+@generate_line arg1 arg2
+```
+may be used if `macroexpand(__module__, :(@num(arg1, arg2)); recursive=false)` becomes a valid static DSL line.
+
+- A macro may be called USING PARENTHESES as the first argument to an `@trace` expression.
+
+So it is valid to include
+```julia
+a = @trace(@output_distribution(arg1, arg2), :addr)
+```
+
+But it is invalid to include
+
+```julia
+a = @trace(@output_distribution arg1 arg2 , :addr)
+```
+
+(The reason macros cannot be used without parentheses is that then `:addr` is parsed as part of the macro call.)
+
 ### Loading generated functions
 Before a function with a static annotation can be used, the [`load_generated_functions`](@ref) method must be called:
 ```@docs
--- a/src/dsl/dsl.jl
+++ b/src/dsl/dsl.jl
@@ -53,7 +53,7 @@ end
 include("dynamic.jl")
 include("static.jl")
 
-function parse_gen_function(ast, annotations)
+function parse_gen_function(ast, annotations, __module__)
     if ast.head != :function
         error("syntax error at $ast in $(ast.head)")
     end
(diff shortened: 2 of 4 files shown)


georgematheos:
I realize that this implementation may not handle escapes within macros properly.  (It might; I just haven't tested it carefully.)  Worth thinking about more carefully before pushing; I'll work on it once I get a chance.

georgematheos:
Actually, I think that it might be pretty hard to handle escaping well here.  The issue is that Julia doesn't expose all of the macroexpansion backend to the Julia user (most of the code for it is written in C and femtolisp, not Julia), so I don't think we have full control over how the Julia macros end up being expanded.

The difficulty arises because we need the `@trace` macro to not get macroexpanded in the static DSL, since this is parsed as a special symbol.  As a result, when macroexpanding in the static DSL parser, we can't do a fully recursive macroexpansion, since this would destroy the `@trace` calls.  (I suppose we could allow these to macroexpand into `traceat` calls and then parse these...maybe this is what we should do?  But assuming we don't want to do this:)  Thus we need to do iterative shallow macroexpands, but this appears to result in different hygiene behavior than the recursive macroexpansion!  This means that if I write a macro in a way where it can iteratively be macroexpanded into comprehensible for the static DSL, if I try to use it in the dynamic DSL, it will not be macroexpanded properly!

The example I'm looking at is the `@num` macro I am writing (defined in the `Blog` module), where
```julia
# non recursive expansion
@macroexpand1 @num T = [() -> poisson(5)]
```
outputs
```
_T_set = #= test/tests.jl:24 =# @trace((ObjectSetCreator{$(Expr(:escape, :T))}(NumDeclaration((), poisson, (((), (arg1,))->(arg1,)))))(((), ($(Expr(:escape, 5)),))), :_T_set)
```

If I recursively expand this, I get
```julia
_T_set = Gen.traceat(Main.Tests.Blog.:(var"##state#402"), Main.Tests.Blog.ObjectSetCreator{T}(Main.Tests.Blog.NumDeclaration((), Main.Tests.Blog.poisson, (((), (Main.Tests.Blog.arg1,))->(Main.Tests.Blog.arg1,)))), (((), (5,)),), :_T_set)
```
where for some reason `var"##state#402"` ends up getting nested into the `Main.Tests.Blog` module!  We don't want this, but we do want `ObjectSetCreator` to become nested into the Blog module.

On the other hand, if I run a single macroexpansion and then macroexpand recursively from there in the Main module, I get very different behavior:
```julia
macroexpand(Main, macroexpand(Main, quote @num T = [() -> poisson(5)] end, recursive=false), recursive=true)
```
yields
```
_T_set = Gen.traceat(var"##state#402", ObjectSetCreator{$(Expr(:escape, :T))}(NumDeclaration((), poisson, (((), (arg1,))->(arg1,)))), (((), ($(Expr(:escape, 5)),)),), :_T_set)
```
Where _nothing_ is wrapped in the `Blog` module!  This is good in the sense that we don't want `var"##state#402"` to end up wrapped, but bad since we do want `ObjectSetCreator` wrapped.

I'm not sure I entirely understand the macroexpansion behavior in Julia, but I'm not aware of a way to exercise detailed enough control over hygiene rules to ensure that in both cases, the variables we want to be module-nested end up properly nested, and the ones we don't want to be nested do not.  If anyone who sees this is aware of a way to do this, please let me know!


It looks like the Julia devs have been considering trying to improve macro hygiene rules for a while now, so at some point this might become easier to deal with (https://github.com/JuliaLang/julia/pull/6910, https://github.com/JuliaLang/julia/issues/23221) but it doesn't seem that there's quite a clear path to when this will be implemented.


In terms of solutions which don't require rewriting portions of the julia macroexpansion codebase (which, to be fair, could be a fun little hackathon project!), I think probably the best solution would either be to
1. Recursively macroexpand the entire static DSL function body, including the `@trace` expressions, before parsing the function.  Then, instead of the static DSL code looking for `@trace` macro calls, it will look for `traceat` function calls.
or
2. Have the dynamic DSL also do shallow iterative macroexpands, so it's behavior mirrors the static DSL.  (Perhaps we could use the `MacroTools` lib to walk the syntax tree and expand out all macros not in the set of special Gen macros).  This certainly isn't a perfect solution (since ideally the way we expand macros in Gen would be consistent with the way the macros would be expanded on their own), so probably (1) is a better call.  But I have some feeling it might end up being more complicated than I'm guessing to get 1 working, so if we can't do it, this might be better.

Anyway, I'm probably not going to have time to implement either of these solutions for a little while and will just be using the code I included in this PR in the meantime.  I'll close this PR and make an issue referencing this to mark it as a TODO.

PR #186, opened 2020-02-06
add filtering a choice map based on selection; and complement selections

marcoct:
- Add a complement selection (that is the set complement of a given selection)

- Add filtering of choice maps based on presence in a selection

- Includes unit tests for both of these new features

### DIFF
--- a/src/address.jl
+++ b/src/address.jl
@@ -102,11 +102,34 @@ A singleton type for a selection that contains all choices at or under an addres
 """
 struct AllSelection <: Selection end
 get_address_schema(::Type{AllSelection}) = AllAddressSchema()
-Base.isempty(::AllSelection) = false
+Base.isempty(::AllSelection) = false # it is not guaranteed to be empty
 Base.in(addr, ::AllSelection) = true
 Base.getindex(::AllSelection, addr) = AllSelection()
 
 export AllSelection
+
+########################
+# complement selection #
+########################
+
+"""
+    ComplementSelection(complement::Selection)
+
+A selection that is the complement of another selection.
+
+An address is in the selection if it is not in the complement selection.
+"""
+struct ComplementSelection <: Selection
+    complement::Selection
+end
+get_address_schema(::Type{ComplementSelection}) = DynamicAddressSchema()
+Base.isempty(::ComplementSelection) = false # it is not guaranteed to be empty
+Base.in(addr, selection::ComplementSelection) = !(addr in selection.complement)
+function Base.getindex(selection::ComplementSelection, addr)
+    ComplementSelection(selection.complement[addr])
+end
+
+export ComplementSelection
 
 ####################
 # static selection #
@@ -156,7 +179,7 @@ end
 function Base.getindex(selection::StaticSelection, addr::Pair)
     (first, rest) = addr
     subselection = selection.subselections[first]
-    get_subselection(subselection, rest)
+    subselection[rest]
 end
 
 function Base.in(addr::Symbol, selection::StaticSelection{T,U}) where {T,U}
--- a/src/address.jl
+++ b/src/address.jl
@@ -112,13 +112,6 @@ export AllSelection
 # complement selection #
 ########################
 
-"""
-    ComplementSelection(complement::Selection)
-
-A selection that is the complement of another selection.
-
-An address is in the selection if it is not in the complement selection.
-"""
 struct ComplementSelection <: Selection
     complement::Selection
 end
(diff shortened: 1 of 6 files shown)


src/address.jl
```diff
@@ -156,7 +179,7 @@ end
 function Base.getindex(selection::StaticSelection, addr::Pair)
     (first, rest) = addr
     subselection = selection.subselections[first]
-    get_subselection(subselection, rest)
+    subselection[rest]
```
marcoct, line 186: This was a bug.

src/choice_map.jl
```diff
@@ -953,3 +953,24 @@ nested_view(c::ChoiceMap) = ChoiceMapNestedView(c)
...
+        if (key in selection)
+            output[key] = value
+        end
+    end
+    for (key, submap) in get_submaps_shallow(choices)
+        subselection = selection[key]
+        set_submap!(output, key, get_selected(submap, subselection))
+    end
+    output
+end
```
bzinberg, line 977: Can we add `iter_deep` from https://github.com/probcomp/Gen/issues/123 and then have this function be something like

```julia
get_selected(choices::ChoiceMap, selection::Selection) = choicemap(
    ((keypath, choices[keypath]) for keypath in iter_deep(choices)
     if matches(keypath, selection))...)
```
?

src/choice_map.jl
```diff
@@ -953,3 +953,24 @@ nested_view(c::ChoiceMap) = ChoiceMapNestedView(c)
...
 
 export nested_view
+
+#############################################
+# filtering choice maps based on selections #
+#############################################
+
+function get_selected(
+        choices::ChoiceMap, selection::Selection)
+    output = choicemap()
```
bzinberg, line 966: I think we want this to return the same type as `choices`, not always a `DynamicChoiceMap`?

src/choice_map.jl
marcoct: I agree that `iter_deep` is useful for certain use cases, but my concern with using it here (and elsewhere in the internal implementations of the choice maps API) is that `matches(keypath, selection)` is linear in the depth of the address. Simultaneously doing a walk of two trees is a common implementation technique in the Gen implementation that avoids this.

src/choice_map.jl
marcoct: I don't think so. User code does not and should not depend on the concrete type of a choice map. The exception is mutable choice maps (DynamicChoiceMaps) that the user constructs themselves. Otherwise choice maps all implement the same interface.

For example, some concrete choice map types are views into traces. The fact that the choice map is a view on a trace is a performance optimization --- there is nothing you can do with such a choice map that you cannot do with a DynamicChoiceMap.

bzinberg:
Is that still true if we require `matches` to depend only on the bottom
level key (as always to be already the case in the existing implementation)?

src/choice_map.jl
bzinberg: Is that still true if we require `matches` to depend only on the bottom
level key (as appears to be already the case in the existing implementation)?

The principle is that I want to separate the traversal logic from the matching logic, if feasible.  So I'm trying to understand if/why the separation is not possible in this case.

src/choice_map.jl
bzinberg: Ah ok.  Hmm... Does the API require that any trace type be able to interoperate with any kind of choice map?

marcoct: Yes.

bzinberg: Ok, in that case SGTM.

src/choice_map.jl
bzinberg: Oops, I just reread the code and saw that there's a traversal of keypaths in the selection too.  OK, thinking about it again.

bzinberg: Seems reasonable, up to me not knowing the motivation for having the current implementations of choice maps use the data structures it uses, instead of nested dicts.  Let's call this resolved now, and I'll ask you later for some info on the above.

src/choice_map.jl
```diff
@@ -953,3 +953,24 @@ nested_view(c::ChoiceMap) = ChoiceMapNestedView(c)
 # aux data together.
 
 export nested_view
+
+#############################################
+# filtering choice maps based on selections #
+#############################################
+
+function get_selected(
```
bzinberg, line 964: This is a new function, right?  Do we want to add it to the public API?  (If so let's add a brief docstring)

bzinberg (approved):

src/choice_map.jl
marcoct: Good point.

marcoct: I added it to the public API for choice maps, and also added a function `complement()` to the public API for selections.

bzinberg:
(Awesome, looks good to merge)

PR #187, opened 2020-02-08
Document `verbose` flag for importance sampling

ali-ramadhan:
Only found out about the `verbose` flag by looking through the code so thought it would be useful to document the option in case others find find it useful too.

This PR documents the `verbose` flag for `importance_resampling` and adds a similar verbose option to `importance_sampling`.


marcoct:
Thank you @ali-ramadhan!

PR #188, opened 2020-02-12
Improve documentation for MCMC; add MCMC kernel DSL; refactor inference docs

marcoct:
- Add an MCMC Kernel DSL with documentation and an example (but no unit tests yet)
- Add documentation for MCMC, including mention of convergence, stationarity, etc.
- Split up the inference docs from te modeling language docs


alex-lew:
@marcoct Very cool! Will do a more thorough review soon, but a first thought is that `mh`, `hmc`, and `elliptical_slice` could have variants that are already `pkern`s, with the right reversals specified. (Is there a reason this is not feasible that I'm not seeing?)

This would make the use of `@pkern` an advanced feature, used only when you wanted to write a truly custom kernel whose stationarity you've ensured yourself. Then `@ckern` could just be `@kern`, and users could directly call MH (using the `~` syntax) inside the function.

(On a totally subjective note, I wonder if we could just let the user choose an arbitrary non-macro name for the trace in a `@ckern`, instead of `(@T)`? The rule, which could be enforced by the `@ckern` macro, would just be that this `trace` variable can't be assigned to except by using `~`. I think this would make the kernel functions look a little more natural / less noisy.)

marcoct:
> but a first thought is that `mh`, `hmc`, and `elliptical_slice` could have variants that are already `pkern`s, with the right reversals specified. (Is there a reason this is not feasible that I'm not seeing?)

I implemented this in the latest commit.

> This would make the use of `@pkern` an advanced feature, used only when you wanted to write a truly custom kernel whose stationarity you've ensured yourself. Then `@ckern` could just be `@kern`, and users could directly call MH (using the `~` syntax) inside the function

Agreed. Done.

> (On a totally subjective note, I wonder if we could just let the user choose an arbitrary non-macro name for the trace in a `@ckern`, instead of `(@T)`? The rule, which could be enforced by the `@ckern` macro, would just be that this `trace` variable can't be assigned to except by using `~`. I think this would make the kernel functions look a little more natural / less noisy.)

This makes sense. But I did not have time to implement it today.

marcoct:
> This makes sense. But I did not have time to implement it today.

Okay, I guess I did have time :)

Now the example is:

```julia
@kern function my_kernel(trace)

    # cycle through the x's and do a random walk update on each one
    for i in 1:trace[:n]
        trace ~ mh(trace, random_walk_proposal, (i,))
    end

    # repeatedly pick a random x and do a random walk update on it
    if trace[:n] > 0
        for rep in 1:10
            let i ~ uniform_discrete(1, trace[:n])
                trace ~ mh(trace, random_walk_proposal, (i,))
            end
        end
    end

    # remove the last x, or add a new one, a random number of times
    let n_add_remove_reps ~ uniform_discrete(0, max_n_add_remove)
        for rep in 1:n_add_remove_reps
            trace ~ mh(trace, add_remove_proposal, (), add_remove_involution)
        end
    end
end
```

PR #190, opened 2020-02-13
Implement `length`, `keys` and `==` for `ChoiceMapNestedView`

bzinberg:
(cc @vkmvkmvkmvkm)

Tested: unit


bzinberg:
This is a non-conceptual FR fix, so in line with what Marco said earlier, I'll go ahead and merge this in over the weekend if no one objects.

PR #191, opened 2020-02-17
Add support for ~ syntax within `@gen` modeling DSLs

alex-lew:
This pull request adds the following syntactic sugar to the two `@gen` modeling DSLs:

Sugar|Desugars to
-----|-----------
`x ~ e`|`x = @trace(e, :x)`
`{a} ~ e`|`@trace(e, a)`
`{*} ~ e`|`@trace(e)`

For example:

```julia
@gen function linear_regression(xs)
  # Assign variables and give the choices addresses at the same time
  slope ~ normal(0, 1)
  intercept ~ normal(0, 1)
  prob_outliers ~ uniform(0, 0.5)

  for (i, x) in enumerate(xs)
    # A tilde expression can appear anywhere, but should be
    # surrounded by parentheses (otherwise precedence is too low).
    if ({:is_outlier => i} ~ bernoulli(prob_outliers))
      {:y => i} ~ normal(0, 10)
    else
      {:y => i} ~ normal(slope * x + intercept, 1)
    end
  end
end
```

Some notes:

The precedence of the tilde operator is relatively weak. So, for example, the parentheses in the code `if (b ~ bernoulli(0.4)); x; else; y; end` are required.

Currently, we do not allow expressions like `x[3] ~ normal(0, 1)` or `(x, y) ~ f()` or `x.y ~ normal(0, 1)`. All of these are in theory possible to support -- we just need to decide on a convention for converting the left-hand side into an address. There is a function, `address_from_expression`, for doing this -- right now it only supports symbols, but that should be easy to change by editing this function.

I wasn't sure what the best way to test this was, so I modified some of the existing tests to run twice, with and without tilde syntax. Let me know if you'd prefer a different unit testing strategy :-)

I haven't yet added/changed documentation. If people like the syntax, we may decide to shift docs and tutorials to use tilde by default; otherwise, we can just describe the sugar as being an option.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -164,7 +164,7 @@ uuid = "90014a1f-27ba-587c-ab20-58faa44d9150"
 version = "0.9.10"
 
 [[Pkg]]
-deps = ["Dates", "LibGit2", "Markdown", "Printf", "REPL", "Random", "SHA", "UUIDs"]
+deps = ["Dates", "LibGit2", "Libdl", "Logging", "Markdown", "Printf", "REPL", "Random", "SHA", "UUIDs"]
 uuid = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 
 [[Printf]]
@@ -249,7 +249,7 @@ uuid = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
 version = "0.9.3"
 
 [[SuiteSparse]]
-deps = ["Libdl", "LinearAlgebra", "SparseArrays"]
+deps = ["Libdl", "LinearAlgebra", "Serialization", "SparseArrays"]
 uuid = "4607b0f0-06f3-5cda-b6b1-a6196a1729e9"
 
 [[Test]]
--- a/src/dsl/dsl.jl
+++ b/src/dsl/dsl.jl
@@ -66,14 +66,14 @@ end
 function desugar_tildes(expr)
     MacroTools.postwalk(expr) do e
         if MacroTools.@capture(e, {*} ~ rhs_)
-          :(@trace($rhs))
-      elseif MacroTools.@capture(e, {addr_} ~ rhs_)
-          :(@trace($rhs, $(addr)))
-      elseif MacroTools.@capture(e, lhs_ ~ rhs_)
-          addr_expr = address_from_expression(lhs)
-          :($lhs = @trace($rhs, $(addr_expr)))
+            :(@trace($rhs))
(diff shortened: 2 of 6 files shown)


marcoct:
@alex-lew This is really nice! A big improvement.

Re the tests -- I like that you loop over the two syntaxes and reuse most of the test code. What do you think about just testing the conversion from `~` syntax to `@trace` syntax (e.g. by comparing the two ASTS, after some normalization e.g. removing line nodes, etc.)?

fsaad:
How does an expression such as `n = @trace(geometric(0.5),  :n) + 1` translate?

marcoct:
> How does an expression such as `n = @trace(geometric(0.5), :n) + 1` translate?

First, in this case you would probably want to use the distributions DSL to directly construct the geometric(0.5) + 1 distribution:
```julia
@dist geom_plus_one(p) = geometric(p) + 1
...
n = @trace(geom_plus_one(0.5), :n)
n ~ geom_plus_one(0.5)
```

Assuming you didn't want to use the dist DSL, then it would be:
```julia
n = ({:n} ~ geometric(0.5)) + 1
```

But it would probably be better to name it:
```julia
n = ({:n_minus_one} ~ geometric(0.5)) + 1
```

marcoct:
@alex-lew I think something like this could suffice for the tests, and would result in less code to maintain. Actually maybe this plus smoke test for each language (e.g. running simulate() on a model written in each language) to make sure the syntax is supported at all. What do you think?

```julia
using Gen
import MacroTools

normalize(ex) = MacroTools.prewalk(MacroTools.rmlines, ex)

# dynamic

expected = normalize(:(
@gen function foo()
    x = @trace(normal(0, 1), :x) 
    y = @trace(normal(0, 1), :y => :z)
    z = @trace(bar())
end))

actual = normalize(Gen.desugar_tildes(:(
@gen function foo()
    x ~ normal(0, 1)
    y = ({:y => :z} ~ normal(0, 1))
    z = ({*} ~ bar())
end)))

@assert actual == expected

# static

expected = normalize(:(
@gen (static) function foo()
    x = @trace(normal(0, 1), :x) 
    y = @trace(normal(0, 1), :y)
end))

actual = normalize(Gen.desugar_tildes(:(
@gen (static) function foo()
    x ~ normal(0, 1)
    y = ({:y} ~ normal(0, 1))
end)))

@assert actual == expected
```

alex-lew:
Looks good -- just committed those changes!

PR #195, opened 2020-02-28
Add support for optional positional arguments in the dynamic DSL.

ztangent:
Currently, the built-in modeling languages and the GFI have no support for optional trailing arguments, nor for keyword arguments. This pull request provides support for optional trailing arguments for the dynamic modeling language (but not the static modeling language), enabling the following:

```julia
@gen function foo(x, y=2, (grad)(z::Float64=3.0))
    a = @trace(normal(x+y+z, 1), :a)
    b = @trace(normal(a, 1), :b)
    return (x, y, z, x+y+z)
end
```

`foo` can now be called with optional arguments omitted:
```julia
julia> foo(5, 4)
(5, 4, 3.0, 12.0)
```

Traces produced by `generate`, `simulate`, `update`, etc., have default arguments filled in as expected:
```julia
julia>  tr, w = generate(foo, (5, 4), choicemap());
julia> get_args(tr)
(5, 4, 3.0)
julia> new_tr, _, _, _ = update(foo, (1,), (UnknownChange(),) choicemap());
julia> get_args(new_tr)
(1, 2, 3.0)
```

Combinators also work as expected:
```julia
julia> map_foo = Map(foo);
julia> tr, w = generate(map_foo, ([1, 2, 3], [4, 5, 6]), choicemap());
julia> get_retval(tr)[3]
(3, 6, 3.0, 12.0)
```

We can even do optional arguments with `GenerativeFunction`s as default values!
```julia
@gen function outer(inner::GenerativeFunction=foo)
    @trace(inner(1), :inner)
end
```
```julia
julia> foo_but_with_one_less_argument = outer;
julia> foo_but_with_one_less_argument()
(1, 2, 3.0, 6.0)
```
Test cases are provided in `test/optional_args.jl`, and the docs have been updated accordingly to characterize behavior when optional arguments are omitted.


bzinberg:
Awesome!

Is there documentation you can update, ideally within this PR?

ztangent:
@bzinberg Yup, I'll do that, and add some test cases to make sure combinators etc. are working with this change!

src/dsl/static.jl
```diff
@@ -256,11 +256,14 @@ function make_static_gen_function(name, args, body, return_type, annotations)
     push!(stmts, :(set_accepts_output_grad!(builder, $(QuoteNode(accepts_output_grad)))))
     bindings = Dict{Symbol,Symbol}() # map from variable name to node name
     for arg in args
+        if arg.default != nothing
+            @warn "Ignoring default value $(arg.name)=$(arg.default), unsupported in static DSL."
```
marcoct: @ztangent I think this should probably be an error. I don't see a reason to allow the user to leave the default argument in their code.

marcoct:
@ztangent Nice job. I just had one comment.

src/dsl/static.jl
ztangent: Done!

PR #196, opened 2020-02-28
Implemented Laplace distribution in Gen 

Kenta426:
implemented a laplace distribution in src/modeling_library/laplace.jl, and a quick testcase in test/modeling_library/distributions.jl.


marcoct:
Perfect! Thanks @Kenta426.

PR #197, opened 2020-03-03
Document Distributions DSL

### LINKED ISSUE #154: Document distributions DSL
The "distribution DSL feature" (https://github.com/probcomp/Gen/pull/103) was merged, but isn't yet added to the Gen documentation.

The PR contains a pretty detailed Markdown writeup about the feature.

We might only want to document those parts of the feature that are more stable / tested / complete, or we could explain the missing functionality that might be expected (gradients for vector-valued random choices is one).

bzinberg:
Documents the Distributions DSL by pasting in the PR description of https://github.com/probcomp/Gen/pull/103 and making light edits.

Fixes https://github.com/probcomp/Gen/issues/154.

## Tested
Previewed the doc site locally.

### DIFF
--- a/docs/src/ref/distributions.md
+++ b/docs/src/ref/distributions.md
@@ -1,5 +1,3 @@
-## Probability Distributions
-
 ## Built-In Distributions
 
 ```@docs
--- a/docs/src/ref/distributions.md
+++ b/docs/src/ref/distributions.md
@@ -70,7 +70,7 @@ the `:n_students` trace address.
 
 ### Permitted means of combination
 
-Of course, it is not possible for `@dist` to work on any arbitrary `body`.  We
+It is not possible for `@dist` to work on any arbitrary `body`.  We
 now describe which constructs are permitted inside the `body` of a `@dist`
 expression.
 
--- a/docs/src/ref/distributions.md
+++ b/docs/src/ref/distributions.md
@@ -90,8 +90,6 @@ each expression also belongs to one of three "type spaces." These are:
 **Importantly, Julia control flow constructs generally expect `CONST` values:
 the condition of an `if` or the range of a `for` loop cannot be `ARG` or
 `RND`.**
-_(Developer note: We could change this: it would not be too difficult, e.g., to
-add support for most use cases of `if`.)_
 
 The body expression as a whole must be a `RND` expression, representing a
 random variable. The behavior of the `@dist` definition is then to define a new
--- a/docs/src/ref/distributions.md
+++ b/docs/src/ref/distributions.md
@@ -20,8 +20,9 @@ laplace
 ## Creating New Distributions via the Distributions DSL
 
 In addition to using the above built-in distributions, and to defining subtypes
-of `Gen.Distribution` in plain Julia, you may combine existing distributions to
-create new ones using the `@dist` DSL.  The syntax for this DSL is
+of `Gen.Distribution` in plain Julia, you may apply deterministic
+transformations to existing distributions and thereby create new ones, using
+the `@dist` DSL.  The syntax for this DSL is
 
 ```julia
 @dist name(arg1, arg2, ..., argN) = body
@@ -68,7 +69,7 @@ and think about the more natural random variable `:n_students`.  This leads to
 more natural inference programs, which can constrain and propose directly to
 the `:n_students` trace address.
 
-### Permitted means of combination
+### Permitted constructs for the body of a `@dist`
 
 It is not possible for `@dist` to work on any arbitrary `body`.  We
 now describe which constructs are permitted inside the `body` of a `@dist`
--- a/docs/src/ref/distributions.md
+++ b/docs/src/ref/distributions.md
@@ -35,7 +35,7 @@ end
 
 Here `body` is ordinary Julia code, with the constraint that `body` must
 contain exactly one random choice.  The value of the `@dist` expression is then
-a `Gen.distribution` object called `name`, parameterized by `arg1, ..., argN`,
+a `Gen.Distribution` object called `name`, parameterized by `arg1, ..., argN`,
 representing the distribution over _return values_ of `body`.
 
 This DSL is designed to address the issue that sometimes, values stored in the
--- a/docs/src/ref/distributions.md
+++ b/docs/src/ref/distributions.md
@@ -169,7 +169,7 @@ exp(normal(x, 1)) :: RND  (by rule 6)
 ```
 
 Here's another:
-```
+```julia
 @dist function labeled_cat(labels, probs)
 	index = categorical(probs)
 	labels[index]
@@ -190,7 +190,7 @@ vectors. So, for example, it also works with Dicts.
 
 Another one (not as realistic, but it uses all the rules):
 
-```
+```julia
 @dist function weird(x)
   log(normal(exp(x), exp(x))) + (x * (2 + 3))
 end
--- a/docs/src/ref/distributions.md
+++ b/docs/src/ref/distributions.md
@@ -192,28 +192,3 @@ normal(exp(x), exp(x)) :: RND 				(by rule 3)
 log(normal(exp(x), exp(x))) :: RND 			(by rule 6)
 log(normal(exp(x), exp(x))) + (x * (2 + 3)) :: RND 	(by rule 6)
 ```
-
-## A note for the long-term future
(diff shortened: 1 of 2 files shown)


ztangent:
@bzinberg thanks for doing this! i was almost going to do it myself. while you're at it, do you think you could add `geometric` to the current list of built-in distributions? for whatever reason, it's currently missing.

also, it seems to me like the note for the future shouldn't be included in the docs, maybe? that's more like an internal design note.

bzinberg:
> @bzinberg thanks for doing this! i was almost going to do it myself. while you're at it, do you think you could add geometric to the current list of built-in distributions? for whatever reason, it's currently missing.

Let's call that a separate PR, so that submission is not delayed by (and so that I'm not on the hook for) anything beyond this one fix.

> also, it seems to me like the note for the future shouldn't be included in the docs, maybe? that's more like an internal design note.

Do we have a separate place for internal design notes (that people might ever actually read)?  I actually think it's quite reasonable to put a note like this in the docs (there are others that I deleted because they weren't as appropriate, as shown in the git diff) -- interested users may want to see and it's not secret.  WDYT?

ztangent:
@bzinberg gotcha, just made a separate pull request just to make sure `geometric` is added to the docs. I think @marcoct and @alex-lew will have more informed thoughts on your second question!

docs/src/ref/distributions.md
```diff
@@ -19,3 +17,207 @@ exponential
 laplace
 ```
 
+## Creating New Distributions via the Distributions DSL
+
+In addition to using the above built-in distributions, and to defining subtypes
+of `Gen.Distribution` in plain Julia, you may combine existing distributions to
```
alex-lew: You can't use the `@dist` DSL to combine existing distributions, unfortunately -- only one random choice per `@dist` :)

docs/src/ref/distributions.md
```diff
@@ -19,3 +17,207 @@ exponential
...
+or
+```julia
+@dist function name(arg1, arg2, ..., argN)
+    body
+end
+```
+
+Here `body` is ordinary Julia code, with the constraint that `body` must
+contain exactly one random choice.  The value of the `@dist` expression is then
+a `Gen.distribution` object called `name`, parameterized by `arg1, ..., argN`,
```
alex-lew: `Gen.Distribution` with a capital `D`?

docs/src/ref/distributions.md
```diff
@@ -19,3 +17,207 @@ exponential
...
+end
+```
+
+and think about the more natural random variable `:n_students`.  This leads to
+more natural inference programs, which can constrain and propose directly to
+the `:n_students` trace address.
+
+### Permitted means of combination
+
+Of course, it is not possible for `@dist` to work on any arbitrary `body`.  We
```
alex-lew: Maybe it is not obvious to readers of the docs -- so perhaps we can remove "Of course"?

docs/src/ref/distributions.md
```diff
@@ -19,3 +17,207 @@ exponential
...
+   All expressions representing non-random values that depend on distribution
+   arguments are `ARG` expressions.
+3. `RND`: Random variables. All expressions whose runtime values may differ
+   across multiple calls to this distribution (with the same arguments) are
+   `RND` expressions.
+
+**Importantly, Julia control flow constructs generally expect `CONST` values:
+the condition of an `if` or the range of a `for` loop cannot be `ARG` or
+`RND`.**
+_(Developer note: We could change this: it would not be too difficult, e.g., to
```
alex-lew: I'd be in favor of removing the "developer note" here. If we want to make this a priority, we can open an issue?

docs/src/ref/distributions.md
```diff
@@ -19,3 +17,207 @@ exponential
...
+a single direction of information flow -- anything that touches classified data
+is also classified. Also, Tabular v2's type system, which similarly separates
+deterministic and random values in its type system.)
+
+#### Restrictions
+Users may _not_ reassign to arguments (like `x` in the above example), and may
+not apply functions with side effects. Names bound to expressions of type `RND`
+must be used only once. e.g., `let x = normal(0, 1) in x + x` is not allowed.
+
+**TODO:** (Expand on scalar-vs.-vector problems here.)
```
alex-lew: Is it OK to leave this TODO in the docs?

docs/src/ref/distributions.md
```diff
@@ -19,3 +17,207 @@ exponential
...
+2 + 3 :: CONST 						(by rule 4)
+x :: ARG 						(by rule 2)
+x * (2 + 3) :: ARG 					(by rule 5)
+exp(x) :: ARG 						(by rule 5)
+normal(exp(x), exp(x)) :: RND 				(by rule 3)
+log(normal(exp(x), exp(x))) :: RND 			(by rule 6)
+log(normal(exp(x), exp(x))) + (x * (2 + 3)) :: RND 	(by rule 6)
+```
+
+## A note for the long-term future
```
alex-lew: What do you see as the benefit for putting this in the Gen documentation?

alex-lew:
Thanks for doing this, Ben! Left a few comments but overall looks good.

docs/src/ref/distributions.md
bzinberg: Good call, an issue is the right place for this.  It's also probably the right place for the "long-term future note" below.  Filed https://github.com/probcomp/Gen/issues/199.

docs/src/ref/distributions.md
bzinberg: Issue would be better -- filed https://github.com/probcomp/Gen/issues/200, thanks.

docs/src/ref/distributions.md
bzinberg: The benefit was that users (who may be researchers in this area) can be aware of the discussion and participate if they want.  But as you pointed out above, an issue is a better place for this.  Filed https://github.com/probcomp/Gen/issues/201.

docs/src/ref/distributions.md
bzinberg: Done

docs/src/ref/distributions.md
bzinberg: Done

docs/src/ref/distributions.md
bzinberg: Fixed.

bzinberg (commented):
Good calls on all fronts, I think I've addressed them all

marcoct:
@bzinberg This looks great, but I have one comment:

Users may implement new distributions by implementing the Distribution interface directly, or with the Distributions DSL. Currently, implementing new distributions by implementing the Distribution interface directly is documented under 'Extending Gen' here: 

https://probcomp.github.io/Gen/dev/ref/extending/#Custom-distributions-1

The title of the section ' ## Creating New Distributions via the Distributions DSL' might lead some readers of the 'Probability Distributions' section to believe that the distributions DSL is the only way to create a new distribution. It would be good to either add a link to the 'Extending Gen' section from this 'Probability Dsitributions' section, or to move the Distributions DSL documentation to Extending Gen.

marcoct:
@bzinberg Ah I see you mentioned it -- maybe just include a link in that first sentence to https://probcomp.github.io/Gen/dev/ref/extending/#Custom-distributions-1.

alex-lew:
@marcoct @bzinberg Maybe it makes sense to cross-reference in both places? Or, as you suggested, put them booth in the same place. 

We could also be more explicit about the different use cases? "Gen can be extended with new probability distributions, which can then be used to make random choices within generative functions.  Simple transformations of existing distributions can be created using the [Distributions DSL]. Users can construct more complex custom distributions by implementing Gen's `Distribution` interface directly."

bzinberg:
@marcoct @alex-lew Yeah, that totally makes sense.  I took a crack at incorporating your suggestions, WDYT?

marcoct:
@bzinberg Looks great!

bzinberg:
Thanks @alex-lew for writing [basically everything](https://github.com/probcomp/Gen/pull/197/commits/ba26d2bf57f79b1a9efd1bedf9abb8d524b0b4b1), so that all we had to do here were minor [touch-ups](https://github.com/probcomp/Gen/pull/197/files/ba26d2bf57f79b1a9efd1bedf9abb8d524b0b4b1..862c32ed9aeb5a86660c3b6c5e4a8168839b762a).

PR #198, opened 2020-03-03
Add geometric to list of built-in distributions in the docs.

ztangent:
For whatever reason, `geometric` is missing from the docs. This actually led me to implement `geometric` as a series of `bernoulli` flips during my first attempt to use Gen, because I thought Gen just didn't support geometric distributions.


bzinberg (approved):
didn't realize it was such a small change :)

PR #203, opened 2020-03-08
Document a definition of "sample space", "support", "disciplined generative function", and adjust the behavior of `logpdf` accordingly

ztangent:
Currently, `logpdf` has inconsistent behavior across distributions as to what it returns for values outside the support of the distribution. For example, distributions that are implemented by wrapping the corresponding distribution in `Distributions.jl` return `-Inf` outside their support, but `categorical` and `beta` throw errors.

This pull request addresses this inconsistent behavior, ensuring that all distributions return -Inf when queried outside of their support. A summary of changes:

- Works for both built-in and `@dist` distributions.
- Test cases added accordingly (including tests for distributions which had lacked tests altogether).
- `logsumexp` implementation adjusted to handle `-Inf`.
- `logpdf_grad` implemented for `Poisson` because it was missing.

This change should also guarantee that importance sampling doesn't lead to errors when proposal distributions have wider support than the corresponding target distributions. This is useful in a number of cases, such as using a normal as a proposal distribution to a variable sampled from a distribution with support only over the positive reals. Consider, for example, trying to infer a latent variable `x ~ exponential(1)`  from a noisy Gaussian observation of `y ~ normal(x, 0.5)`:

```
@gen function model()
    x ~ exponential(1)
    y ~ normal(x, 0.5)
end
```

If `y` is observed to have a large value, e.g. `y=5`, importance sampling using the prior as the proposal is going to lead to a high variance estimate. We could instead have a proposal that proposes `x` to be centered around `y`:

```
@gen function proposal(y)
    x ~ normal(y, 0.5)
end
```

This will lead to traces being proposed that are much more likely given the constraints, leading to a good estimate of `x` with less traces. But very occasionally, `x` might be proposed to be less than 0, which is impossible under the model. Still, we don't want importance sampling to just throw an error if this happens. Rather, it seems better if we just assign that trace `-Inf` log weight, and carry on sampling. The changes in this pull request should ensure that those sorts of error doesn't occur.

(Note that I haven't tested the specific example given above -- I just believe it should work given the current implementations of `importance_sampling` and `importance_resampling`, which rely upon `logsumexp`, which has been adjusted in this pull request to correctly handle `-Inf` weights instead of returning `NaN`s.)

### DIFF
--- a/docs/src/ref/modeling.md
+++ b/docs/src/ref/modeling.md
@@ -121,7 +121,8 @@ Gen's built in modeling languages require that a address is associated with a fi
 end
 ```
 
-A generative function can be **disciplined** or not. In a disciplined generative function, the support of random choices at each address must be fixed. That is, there exists a set S that is a subset of the sample space such that for all executions of the generative function, if an address appears in the execution, its support must be S. Violating this discipline will cause NaNs, errors, or undefined behavior in certain inference programs. However, there are many inference programs where violating this discipline does not result in undefined behavior and is convenient.
+A generative function can be **disciplined** or not. In a disciplined generative function, the support of random choices at each address must be fixed. That is, for each address `a` there exists a set S that is a subset of the sample space such that for all executions of the generative function, if `a` occurs as the address of a choice in the execution, the support of that choice must be exactly S. Violating this discipline will cause NaNs, errors, or undefined behavior in some inference programs. However, there are many inference programs where violating this discipline does not result in undefined behavior and is convenient.
+Authors of inference programs should document which kinds of undisciplined models their inference algorithms allow or expect to see.
 
 If the support of a random choice needs to change, a disciplined generative function can represent this by using a different address for each distinct value of the support. For example, consider the following generative function:
 ```julia
--- a/docs/src/ref/modeling.md
+++ b/docs/src/ref/modeling.md
@@ -97,11 +97,33 @@ val::Bool = @trace(bernoulli(0.5), :z)
 Not all random choices need to be given addresses.
 An address is required if the random choice will be observed, or will be referenced by a custom inference algorithm (e.g. if it will be proposed to by a custom proposal distribution).
 
-### Choices should have constant support
-The support of a random choice at a given address (the set of values with nonzero probability or probability density) must be constant across all possible executions of the `@gen` function.
-Violating this discipline will cause errors in certain cases.
-If the support of a random choice needs to change, use a different address for each distinct value of the support.
-For example, consider the following generative function:
+
+### Sample space and support of random choices
+
+Different probability distributions produce different types of values for their random choices. For example, the [`Bernoulli`](@ref) distribution results in `Bool` values (either `true` or `false`), the [`Normal`](@ref) distribution results in `Real` values that may be positive or negative, and the [`Beta`](@ref) distributions result in `Real` values that are always in the unit interval (0, 1).
+
+Each [`Distribution`](@ref) is associated with two sets of values:
+
+- The **sample space** of the distribution, which does not depend on the arguments.
+
+- The **support** of the distribution, which may depend on the arguments, and is the set of values that has nonzero probability (or probability density). It may be the entire sample space, or it may be a subset of the sample space.
+
+For example, the sample space of [`Bernoulli`](@ref) is `Bool` and its support is either `{true}`, `{false}`, or `{true, false}`. The sample space of [`Normal`](@ref) is `Real` and its support is the set of all values on the real line. The sample space of [`Beta`](@ref) is `Real` and its support is the set of values in the interval (0, 1).
+
+Gen's built in modeling languages require that a address is associated with a fixed sample space. For example, it is not permitted to use a `Bernoulli` distribution to sample at addresss `:a` in one execution, and a `Normal` distribution to sample at address `:a` in a different execution, because their sample spaces differ (`Bool` vs `Real`):
+```julia
+@gen function foo()
+    if @trace(bernoulli(0.5), :branch)
+        @trace(bernoulli(0.5), :x)
+    else
+        @trace(normal(0, 1), :x)
+    end
+end
+```
+
+A generative function can be **disciplined** or not. In a disciplined generative function, the support of random choices at each address must be fixed. That is, there exists a set S that is a subset of the sample space such that for all executions of the generative function, if an address appears in the execution, its support must be S. Violating this discipline will cause NaNs, errors, or undefined behavior in certain inference programs. However, there are many inference programs where violating this discipline does not result in undefined behavior and is convenient.
+
+If the support of a random choice needs to change, a disciplined generative function can represent this by using a different address for each distinct value of the support. For example, consider the following generative function:
 ```julia
 @gen function foo()
     n = @trace(categorical([0.5, 0.5]), :n) + 1
@@ -109,19 +131,12 @@ For example, consider the following generative function:
 end
 ```
 The support of the random choice with address `:x` is either the set ``\{1, 2\}`` or ``\{1, 2, 3\}``.
-Therefore, this random choice does satisfy our condition above.
-This would cause an error with the following, in which the `:n` address is modified, which could result in a change to the domain of the `:x` variable:
+Therefore, this random choice does not have constant support, and the generative function `foo` is not 'disciplined'.
+Specifically, this could result in undefined behavior for the following inference program:
 ```julia
-tr, _ = generate(foo, (), choicemap((:n, 2), (:x, 3)))
-tr, _ = mh(tr, select(:n))
+tr, _ = importance_resampling(foo, (), choicemap((:x, 3)))
 ```
-We can modify the address to satisfy the condition by including the domain in the address:
-```julia
-@gen function foo()
-    n = @trace(categorical([0.5, 0.5]), :n) + 1
-    @trace(categorical(ones(n) / n), (:x, n))
-end
-```
+It is recommended to write disciplined generative functions when possible.
 
 ## Calling generative functions
 
--- a/docs/src/ref/modeling.md
+++ b/docs/src/ref/modeling.md
@@ -121,8 +121,8 @@ Gen's built in modeling languages require that a address is associated with a fi
 end
 ```
 
-A generative function can be **disciplined** or not. In a disciplined generative function, the support of random choices at each address must be fixed. That is, for each address `a` there exists a set S that is a subset of the sample space such that for all executions of the generative function, if `a` occurs as the address of a choice in the execution, the support of that choice must be exactly S. Violating this discipline will cause NaNs, errors, or undefined behavior in some inference programs. However, there are many inference programs where violating this discipline does not result in undefined behavior and is convenient.
-Authors who want their inference code to be reusable should consider documenting which kinds of undisciplined models their inference algorithms allow or expect to see.
+A generative function can be **disciplined** or not. In a disciplined generative function, the support of random choices at each address must be fixed. That is, for each address `a` there must exist a set S that is a subset of the sample space such that for all executions of the generative function, if `a` occurs as the address of a choice in the execution, then the support of that choice is exactly S. Violating this discipline will cause NaNs, errors, or undefined behavior in some inference programs. However, in many cases it is convenient to write an inference program that operates correctly and efficiently on some specialized class of undisciplined models.
+In these cases, authors who want their inference code to be reusable should consider documenting which kinds of undisciplined models their inference algorithms allow or expect to see.
 
 If the support of a random choice needs to change, a disciplined generative function can represent this by using a different address for each distinct value of the support. For example, consider the following generative function:
 ```julia
--- a/docs/src/ref/modeling.md
+++ b/docs/src/ref/modeling.md
@@ -122,7 +122,7 @@ end
 ```
 
 A generative function can be **disciplined** or not. In a disciplined generative function, the support of random choices at each address must be fixed. That is, for each address `a` there exists a set S that is a subset of the sample space such that for all executions of the generative function, if `a` occurs as the address of a choice in the execution, the support of that choice must be exactly S. Violating this discipline will cause NaNs, errors, or undefined behavior in some inference programs. However, there are many inference programs where violating this discipline does not result in undefined behavior and is convenient.
-Authors of inference programs should document which kinds of undisciplined models their inference algorithms allow or expect to see.
+Authors who want their inference code to be reusable should consider documenting which kinds of undisciplined models their inference algorithms allow or expect to see.
 
 If the support of a random choice needs to change, a disciplined generative function can represent this by using a different address for each distinct value of the support. For example, consider the following generative function:
 ```julia
--- a/src/modeling_library/dist/relabeled_distribution.jl
+++ b/src/modeling_library/dist/relabeled_distribution.jl
@@ -12,7 +12,7 @@ function logpdf(d::WithLabelArg{T, U}, x::T, collection, base_args...) where {T,
                 push!(logprobs, logpdf(d.base, index, base_args...))
             end
         end
-        logsumexp(logprobs)
+        isempty(logprobs) ? -Inf : logsumexp(logprobs)
     else
         error("Cannot relabel a continuous distribution")
     end
@@ -66,7 +66,7 @@ function logpdf(d::RelabeledDistribution{T, U}, x::T, base_args...) where {T, U}
                 push!(logprobs, logpdf(d.base, index, base_args...))
             end
         end
-        logsumexp(logprobs)
+        isempty(logprobs) ? -Inf : logsumexp(logprobs)
     else
         error("Cannot relabel a continuous distribution")
     end
--- a/src/inference/inference.jl
+++ b/src/inference/inference.jl
@@ -2,12 +2,12 @@ import Random
 
 function logsumexp(arr::AbstractArray{T}) where {T <: Real}
     max_arr = maximum(arr)
-    max_arr + log(sum(exp.(arr .- max_arr)))
+    max_arr == -Inf ? -Inf : max_arr + log(sum(exp.(arr .- max_arr)))
 end
 
 function logsumexp(x1::Real, x2::Real)
-    max_arr = max(x1, x2)
-    max_arr + log(exp(x1 - max_arr) + exp(x2 - max_arr))
+    m = max(x1, x2)
+    m == -Inf ? m : m + log(exp(x1 - m) + exp(x2 - m))
 end
 
 export logsumexp
--- a/src/modeling_library/beta.jl
+++ b/src/modeling_library/beta.jl
@@ -11,7 +11,8 @@ Sample a `Float64` from a beta distribution.
 const beta = Beta()
 
 function logpdf(::Beta, x::Real, alpha::Real, beta::Real)
-    (alpha - 1) * log(x) + (beta - 1) * log1p(-x) - logbeta(alpha, beta)
+    (x < 0 || x > 1 ? -Inf :
+    (alpha - 1) * log(x) + (beta - 1) * log1p(-x) - logbeta(alpha, beta) )
 end
 
 function logpdf_grad(::Beta, x::Real, alpha::Real, beta::Real)
--- a/src/modeling_library/beta_uniform.jl
+++ b/src/modeling_library/beta_uniform.jl
@@ -10,9 +10,13 @@ Samples a `Float64` value from a mixture of a uniform distribution on [0, 1] wit
 const beta_uniform = BetaUniformMixture()
 
 function logpdf(::BetaUniformMixture, x::Real, theta::Real, alpha::Real, beta::Real)
-    lbeta = log(theta) + logpdf(Beta(), x, alpha, beta)
-    luniform = log(1.0 - theta)
-    logsumexp(lbeta, luniform)
+    if x < 0 || x > 1
+        -Inf
+    else
+        lbeta = log(theta) + logpdf(Beta(), x, alpha, beta)
+        luniform = log(1.0 - theta)
+        logsumexp(lbeta, luniform)
+    end
 end
 
 function logpdf_grad(::BetaUniformMixture, x::Real, theta::Real, alpha::Real, beta::Real)
--- a/src/modeling_library/categorical.jl
+++ b/src/modeling_library/categorical.jl
@@ -8,7 +8,7 @@ Given a vector of probabilities `probs` where `sum(probs) = 1`, sample an `Int` 
 const categorical = Categorical()
 
 function logpdf(::Categorical, x::Int, probs::AbstractArray{U,1}) where {U <: Real}
-    log(probs[x])
+    (x > 0 && x <= length(probs)) ? log(probs[x]) : -Inf
 end
 
 function logpdf_grad(::Categorical, x::Int, probs::AbstractArray{U,1})  where {U <: Real}
--- a/src/modeling_library/mvnormal.jl
+++ b/src/modeling_library/mvnormal.jl
@@ -8,13 +8,13 @@ Samples a `Vector{Float64}` value from a multivariate normal distribution.
 const mvnormal = MultivariateNormal()
 
 function logpdf(::MultivariateNormal, x::AbstractVector{T}, mu::AbstractVector{U},
-                cov::AbstractMatrix{V}) where {T,U,V}
+                cov::AbstractMatrix{V}) where {T <: Real, U <: Real, V <: Real}
     dist = Distributions.MvNormal(mu, cov)
     Distributions.logpdf(dist, x)
 end
 
 function logpdf_grad(::MultivariateNormal, x::AbstractVector{T}, mu::AbstractVector{U},
-                cov::AbstractMatrix{V}) where {T,U,V}
+                cov::AbstractMatrix{V}) where {T <: Real,U <: Real, V <: Real}
     dist = Distributions.MvNormal(mu, cov)
     inv_cov = Distributions.invcov(dist)
 
@@ -26,7 +26,7 @@ function logpdf_grad(::MultivariateNormal, x::AbstractVector{T}, mu::AbstractVec
 end
 
 function random(::MultivariateNormal, mu::AbstractVector{U},
-                cov::AbstractMatrix{V}) where {T,U,V}
+                cov::AbstractMatrix{V}) where {U <: Real, V <: Real}
     rand(Distributions.MvNormal(mu, cov))
 end
 
--- a/src/modeling_library/piecewise_uniform.jl
+++ b/src/modeling_library/piecewise_uniform.jl
@@ -36,9 +36,10 @@ function logpdf(::PiecewiseUniform, x::Real, bounds::AbstractVector{T},
     # |    probs[1]  |  probs[2]         | probs[3]     |
     if x <= bounds[1] || x >= bounds[end]
         -Inf
+    else
+        bin = get_bin(bounds, x)
+        log(probs[bin]) - log(bounds[bin+1] - bounds[bin])
     end
-    bin = get_bin(bounds, x)
-    log(probs[bin]) - log(bounds[bin+1] - bounds[bin])
 end
 
 function random(::PiecewiseUniform, bounds::Vector{T},
--- a/src/modeling_library/poisson.jl
+++ b/src/modeling_library/poisson.jl
@@ -8,12 +8,11 @@ Sample an `Int` from the Poisson distribution with rate `lambda`.
 const poisson = Poisson()
 
 function logpdf(::Poisson, x::Integer, lambda::Real)
-    x * log(lambda) - lambda - loggamma(x+1)
+    x < 0 ? -Inf : x * log(lambda) - lambda - loggamma(x+1)
 end
 
 function logpdf_grad(::Poisson, x::Integer, lambda::Real)
-    error("Not implemented")
-    (nothing, nothing)
+    (nothing, x/lambda - 1)
 end
 
 
--- a/src/modeling_library/poisson.jl
+++ b/src/modeling_library/poisson.jl
@@ -7,11 +7,11 @@ Sample an `Int` from the Poisson distribution with rate `lambda`.
 """
 const poisson = Poisson()
 
-function logpdf(::Poisson, x::Integer, lambda::Real)
+function logpdf(::Poisson, x::Int, lambda::Real)
     x < 0 ? -Inf : x * log(lambda) - lambda - loggamma(x+1)
 end
 
-function logpdf_grad(::Poisson, x::Integer, lambda::Real)
+function logpdf_grad(::Poisson, x::Int, lambda::Real)
     (nothing, x/lambda - 1)
 end
 
--- a/test/modeling_library/dist.jl
+++ b/test/modeling_library/dist.jl
@@ -4,12 +4,18 @@
 
   @dist labeled_cat(labels, probs) = labels[categorical(probs)]
   @test labeled_cat([:a, :b], [0., 1.]) == :b
+  @test isapprox(logpdf(labeled_cat, :b, [:a, :b], [0.5, 0.5]), log(0.5))
+  @test logpdf(labeled_cat, :c, [:a, :b], [0.5, 0.5]) == -Inf
 
   dict = Dict(1 => :a, 2 => :b)
   @dist dict_cat(probs) = dict[categorical(probs)]
   @test dict_cat([0., 1.]) == :b
+  @test isapprox(logpdf(dict_cat, :b, [0.5, 0.5]), log(0.5))
+  @test logpdf(dict_cat, :c, [0.5, 0.5]) == -Inf
 
   @enum Fruit apple orange
   @dist enum_cat(probs) = Fruit(categorical(probs) - 1)
   @test enum_cat([0., 1.]) == orange
+  @test isapprox(logpdf(enum_cat, orange, [0.5, 0.5]), log(0.5))
+  @test logpdf(enum_cat, orange, [1.0]) == -Inf
 end
--- a/test/modeling_library/distributions.jl
+++ b/test/modeling_library/distributions.jl
@@ -1,10 +1,10 @@
 import DataStructures: OrderedDict
 
 @testset "bernoulli" begin
-    
+
     # random
     x = bernoulli(0.5)
-    
+
     # logpdf_grad
     f = (x::Bool, prob::Float64) -> logpdf(bernoulli, x, prob)
     args = (false, 0.3,)
@@ -21,6 +21,9 @@ end
     x = beta(0.5, 0.5)
     @test 0 < x < 1
 
+    # out of support
+    @test logpdf(beta, -1, 0.5, 0.5) == -Inf
+
     # logpdf_grad
     f = (x, alpha, beta_param) -> logpdf(beta, x, alpha, beta_param)
     args = (0.4, 0.2, 0.3)
@@ -35,6 +38,9 @@ end
     # random
     x = categorical([0.2, 0.3, 0.5])
     @test 0 < x < 4
+
+    # out of support
+    @test logpdf(categorical, -1, [0.2, 0.3, 0.5]) == -Inf
 
     # logpdf_grad
     f = (x, probs) -> logpdf(categorical, x, probs)
@@ -53,6 +59,9 @@ end
     x = gamma(1, 1)
     @test 0 < x
 
+    # out of support
+    @test logpdf(gamma, -1, 1, 1) == -Inf
+
     # logpdf_grad
     f = (x, shape, scale) -> logpdf(gamma, x, shape, scale)
     args = (0.4, 0.2, 0.3)
@@ -63,11 +72,14 @@ end
 end
 
 @testset "inv_gamma" begin
-    
+
     # random
     x = inv_gamma(1, 1)
     @test 0 < x
-    
+
+    # out of support
+    @test logpdf(inv_gamma, -1, 1, 1) == -Inf
+
     # logpdf_grad not implemented
 
 end
@@ -188,11 +200,47 @@ end
     @test isapprox(actual[3][2, 2], finite_diff_mat_sym(f, args, 3, 2, 2, dx))
 end
 
+@testset "uniform" begin
+
+    # random
+    x = uniform(-0.5, 0.5)
+    @test -0.5 < x < 0.5
+
+    # out of support
+    @test logpdf(uniform, -1, -0.5, 0.5) == -Inf
+
+    # logpdf_grad
+    f = (x, low, high) -> logpdf(uniform, x, low, high)
+    args = (0.0, -0.5, 0.5)
+    actual = logpdf_grad(uniform, args...)
+    @test isapprox(actual[1], finite_diff(f, args, 1, dx))
+    @test isapprox(actual[2], finite_diff(f, args, 2, dx))
+    @test isapprox(actual[3], finite_diff(f, args, 3, dx))
+end
+
+@testset "uniform_discrete" begin
+
+    # random
+    x = uniform_discrete(10, 20)
+    @test 10 <= x <= 20
+
+    # out of support
+    @test logpdf(uniform_discrete, -1, 10, 20) == -Inf
+
+    # logpdf_grad
+    args = (1, 1, 5)
+    actual = logpdf_grad(uniform_discrete, args...)
+    @test actual == (nothing, nothing, nothing)
+end
+
 @testset "piecewise_uniform" begin
-   
+
     # random
     x = piecewise_uniform([-0.5, 0.5], [1.0])
     @test -0.5 < x < 0.5
+
+    # out of support
+    @test logpdf(piecewise_uniform, -1, [-0.5, 0.5], [1.0]) == -Inf
 
     # logpdf_grad
     f = (x, bounds, probs) -> logpdf(piecewise_uniform, x, bounds, probs)
@@ -211,7 +259,10 @@ end
     # random
     x = beta_uniform(0.5, 0.5, 0.5)
     @test 0 < x < 1
-    
+
+    # out of support
+    @test logpdf(beta_uniform, -1, 0.5, 0.5, 0.5) == -Inf
+
     # logpdf_grad
     f = (x, theta, alpha, beta) -> logpdf(beta_uniform, x, theta, alpha, beta)
     args = (0.5, 0.4, 10., 2.)
@@ -227,6 +278,9 @@ end
     # random
     @test geometric(0.5) >= 0
 
+    # out of support
+    @test logpdf(geometric, -1, 0.5) == -Inf
+
     # logpdf_grad
     f = (x, p) -> logpdf(geometric, x, p)
     args = (4, 0.3)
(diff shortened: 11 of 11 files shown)


ztangent:
Tested `importance_sampling` with the model and proposal described above, and things work as expected even when there are `-Inf` traces!

```julia
# Estimate x given y=2 using the default proposal
julia> traces, weights, _ = importance_sampling(model, (), choicemap(:y => 2), 100000);
julia> est_x(traces, weights)
1.7506215606278905
# Estimate x given y=2 using the custom proposal
julia> traces, weights, _ = importance_sampling(model, (), choicemap(:y => 2), proposal, (2,), 100000);
julia> est_x(traces, weights)
1.7502471915184172
# Did any of the proposed traces have -Inf log weight? (i.e. zero probability)
julia> any(weights .== -Inf)
true
```

I don't think this particular example leads to errors for `importance_sampling` in the current master branch of Gen, because the `logpdf` implementation for `exponential` already correctly handles values outside of the support. I think it will lead to (very rare) errors for `importance_resampling` though, in the (very rare) event that the first two proposed traces have `-Inf` weight, because calling `logsumexp(-Inf, -Inf)` returns `NaN` in the current master branch.

Even though these errors are rare for this particular example, it'd be good to guarantee that they *never* occur. I can also think of other natural examples where errors would be more likely, e.g., using a Gaussian proposal to a value of `theta ~ uniform(0, 2*pi)`, when your observed variables are `x ~ normal(cos(theta), 0.2)` and `y ~ normal(sin(theta), 0.2)`. A very natural proposal is `theta ~ normal(atan2(x, y), 0.2`, but this will lead to `theta` being sampled outside of the original support when when `y` is very close to 0.

bzinberg:
Do we have a notion of "support" in Gen that is different from "sample space"?  Pasting from [Slack](https://probcomp.slack.com/archives/C5T7HNLRY/p1583689543101000?thread_ts=1583553050.085600&cid=C5T7HNLRY):

>>  For example, suppose there's a choice in the model that is ordinarily a categorical between 1 and 5. But based on the constraints you know it can't be 3.
>>
>> One way of solving this without auxiliary randomness is just proposing from `categorical([0.25, 0.25, 0, 0.25, 0.25])`, zeroing out the third choice.
>
> A question that comes up here is, do we distinguish between a value like 3 (in the sample space but not in the support) vs a value like 10 (not in the sample space)?  I think @marcoct's fail-fast strategy is kind of trying to say "throw an error if the constraint lies outside of the sample space," but Gen doesn't distinguish between support and sample space.  If the distinction did exist, then I would think "times when logprob being `-Inf` is ok" === "times when the constraint is outside (or on the topological boundary of) the support but inside the sample space"

ztangent:
For the implementation of `categorical` in particular, the current master branch of Gen will throw an error if you call `logpdf` with a value that's outside of the sample space, but return `-Inf` if it's within the sample space but outside the support (because it will just do `log(probs[i]) = log(0) = -Inf` ). However, this is inconsistent with the implementation of `logpdf` of other discrete distributions, e.g.,  `uniform_discrete` will return `-Inf` for values outside of its support (which is equal to the sample space, in this case). So will `geometric` for negative integers. This pull request addresses these inconsistencies by ensuring `-Inf` is always returned for anything outside the support, including anything outside of the sample space.

I personally think this behavior is preferable, because there are going to be times when the sample space of a categorical distribution you're sampling from might depend on previous random choices, and you might not know in advance how large the sample space is going to be, so you can't set some maximum limit and zero everything else out.

src/modeling_library/dist/relabeled_distribution.jl
```diff
@@ -66,7 +66,7 @@ function logpdf(d::RelabeledDistribution{T, U}, x::T, base_args...) where {T, U}
                 push!(logprobs, logpdf(d.base, index, base_args...))
             end
         end
-        logsumexp(logprobs)
+        isempty(logprobs) ? -Inf : logsumexp(logprobs)
```
marcoct, line 69: @ztangent Are there cases where logprobs being empty is intended?

marcoct:
@ztangent This looks good, nice job filling in more tests as well.

I had one question about a case which I think could be an error instead of -Inf., but generally I agree that allowing -Inf is convenient. I've also used the Gaussian random walk proposals on a non-negative random variable. And it makes sense to have consistency.

Some more general comments about the general idea of support and strictness, etc.:

Re returning -Inf vs erroring, I think a key distinction is between whether an out-of-support value  is intended by the user or is the result of a bug. The distribution of these two events depends on users' programming style and their expectations on how the system should be used, and these are things that Gen has some control over, so it can encourage people to use a more disciplined (if more tedious) style.

Some of my concern with allowing -Inf instead of errors everywhere is because of the undefined behavior that can appear randomly (and with very low probability). Like in importance sampling as you mentioned. Those sort of errors can be really bad because they are rare, and this is the sort of thing that being more strict (dynamic or static) checks on support will detect. Our POPL trace types paper (https://dl.acm.org/ft_gateway.cfm?id=3371087) is also along these lines.

Another design principle behind Gen is provisioning for future performance optimization and static compilation by ensuring that properties of the computational processes are mad explicit statically. For example, if we know that -Infs are not possible, and proposal/model supports are always matched, then we might be able to generate more efficient code that doesn't include lots of branches and tests for Inf and NaN, and that guarantees that it won't crash.

in the future, it might make sense to have an "unsafe" annotation for the built-in modeling languages. If this flag is on, then zero probability choices result in -Inf instead of an error. If you are in safe mode, then, we guarantee that -Inf never appears. This approach would admit that there are multiple regimes in which a user would like to use Gen, and allow both regimes. But I don't think that needs to be part of this PR.

alex-lew:
@marcoct Are there any worries re: incremental computation here? "Subtracting -Inf and adding 2.0," e.g.

I realize that in general, we won't ever be updating a -Inf-scoring trace in MCMC (we should never move to a zero-probability region). But maybe we want to handle this case anyway, either with an error or with a slower code path that reruns the entire generative function rather than trying to incrementally recompute?

src/modeling_library/dist/relabeled_distribution.jl
ztangent: Good question! I think this hinges on whether we want to allow users to evaluate the logpdf values for `labeled_cat` or `labeled_unif` distributions which may not have constant support, or if we want an error to be thrown. For clarity, by those two distributions, I mean:
```julia
@dist labeled_cat(labels, probs) = labels[categorical(probs)]
@dist labeled_unif(labels) = labels[uniform_discrete(1, length(labels))]
```
I have at least one use case where allowing choices with non-constant support might be justified, and where having `logpdf` return `-Inf` for those choices instead of erroring would be convenient for importance sampling. I'll put this in a separate comment below.

ztangent: Also, I think if we make `logpdf` throw errors for `labeled_cat` and `labeled_unif` when we query values outside their support, then I think we should make `categorical` and `uniform_discrete` do the same for consistency. We might also then want to do the same for all the other discrete distributions (`geometric` and `poisson`). But I wonder if this would be too restrictive...?

ztangent: Okay @alex-lew has a more concrete example of when the support of a `labeled_cat` distribution might be not constant, will let him add it below instead of mine! (Mine was a random graph search example.)

bzinberg: My first thought on this is that it makes sense for `labeled_cat(labels, probs)`'s `logpdf` to throw an error when the input is not an element of `labels`, as @marcoct seems to also be suggesting above.

I think we need to be more precise about what "support" means.  See [Wikipedia](https://en.wikipedia.org/wiki/Support_(measure_theory)).  For `labeled_cat`, I would say that `labels` is the sample space, not necessarily the support.  The support is
```julia
[label for (label, prob) in zip(labels, probs) if prob > 0]
```
(Note that support is a bit more complicated in the continuous case, as there is a topological closure, which has no effect in the discrete case.)

I would say that `labeled_cat` defines a measure on the finite set `labels`, and the only things that measure should know how to talk about are subsets of `labels`.  Hence I think evaluating `logpdf(..., x)` where `x ∉ labels`, which is the same as evaluating `μ({x})` where `{x} ⊄ labels`, should give an error.

@ztangent, if I understand your above comments correctly, it sounds like you want an implicit conversion of measures via "extension by zero," where the Gen interpreter automatically converts μ to a measure ũ which has sample space `labels ∪ {x}`, with `ũ({x}) = 0` and agrees with μ on subsets of `labels` (these conditions uniquely determine ũ).  Perhaps what I'm trying to say is, I don't think that conversion should be implicit -- I suspect it would down the road suffer from the usual gotchas of generous implicit conversions, namely confusing behaviors and silent failures.

On the other hand, an ability to do _explicit_ conversion sounds totally fine, and may achieve the best of both worlds.  Concretely this would mean having a function in the GFI that implements
```
logpdf_extended(x) := (x in sample_space ? logpdf(x) : 0)
```

@ztangent, @marcoct, does the above make sense?

alex-lew: @bzinberg Why not define the sample space of `labeled_cat` to be all Julia values (with the discrete sigma-algebra)? This involves no implicit conversions: it's just how it's defined.

The example that Xuan mentioned above is this:

Suppose I have a generative function that samples a state in the US, and then a city within that state:

```Julia
@dist labeled_cat(labels, frequencies) = labels[categorical(frequencies ./ sum(frequencies))]
@gen function city_and_state()
  state ~ labeled_cat(all_states, state_populations)
  city ~ labeled_cat(cities_within_state[state], city_populations_in_state[state])
end
```

I'd like to observe the `city` and infer the `state`.

Now, technically, the above generative function is "incorrect" in that it uses a single address, `city`, for choices that can have different supports. But if I did use a different address for `city` depending on the state, I wouldn't be able to _observe_ `city` without knowing the state.

bzinberg: A different way of implementing `logpdf_extended`, which actually might be better, is to just implement a proper extension-by-zero of measures.  In this approach, the measure μ and a set A ⊇ sample_space(μ) are specified, and the output is a new measure,
```
ũ = extend_by_zero(μ, A)
```
such that ũ agrees with μ on subsets of `sample_space(μ)` and ũ(B) = 0 for all B ⊆ A.

I think this formulation does both what @marcoct wants and what @ztangent wants in a principled way:

* It still catches the kind of errors @marcoct wants it to catch
* It allows the user to bypass the check by explicitly declaring that `logpdf` of their given value should be computed via extension by zero if necessary.

Note also that it's possible to construct a categorical with big sample space and small support ahead of time, instead of extending it at runtime.  For example, if you want a discrete uniform distribution whose support is not known until runtime, you could have its sample space in every code path be all of ℕ, and have its `logpdf(x)` return 0 whenever `x > length(probs)`.

The analogous thing for `labeled_cat`, or any other discrete distribution, would be to have the set A above be specified by a predicate φ, that is, `φ(x) == true` iff x ∈ A.  Then the user gets to control how permissive the error checking is.  In @ztangent's case, `φ(x)` could return true for every `x`, or could return true whenever `x` looks reasonably close to having the right format.  Making φ stricter results in more error checking, at the cost of having to write a stricter upper bound on A (which is trivial in many cases, esp. when A is finite or integer-valued, but is nontrivial for many cases where A is an infinite discrete set of structured objects).

bzinberg: @alex-lew, didn't see your response before writing my follow-on above, but I think the follow-on addresses your question :slightly_smiling_face:

alex-lew: @bzinberg I think that's an interesting proposal. What about the converse, `restrict_sample_space`? This could be like an assertion, telling Gen that one would like to receive errors for values that fail a certain predicate.

Mostly, I think we need to think carefully about what the default behavior should be. And it's not obvious to me that erroring should be the default behavior. Why shouldn't the implicit sample space of `labeled_cat` be all objects, given that its argument could be changing from execution to execution?

We've looked at a pattern above of "real-world" usage in which the zeroing behavior is desirable. What are realistic errors someone might make that we want to prevent against? Mis-spelled observations? And in practice, would 0-logpdf not serve as a red flag?

One error I think is worth surfacing:

If someone performs importance_resampling when every weight is `-Inf`, I think an error should be raised. This means inference would produce meaningless results anyway.

bzinberg: > @bzinberg I think that's an interesting proposal. What about the converse, `restrict_sample_space`? This could be like an assertion, telling Gen that one would like to receive errors for values that fail a certain predicate.

If we want `restrict_sample_space(μ)` to still be a probability measure, then this is the same thing as suggesting that we add the ability to condition on arbitrary predicates.  If we don't require that, i.e. we keep the values of `logpdf` just as they are, then I think it should literally, as you say, be written as an assertion, rather than building it into an object that has mathematical semantics of "measure that is not a probability measure" (so that we avoid including arbitrary measures into the conceptual machinery of Gen).

bzinberg: > Mostly, I think we need to think carefully about what the default behavior should be.

Definitely -- and more strongly, I think it's important that we _document_ the default behavior, ideally also describing the mathematical formalism it corresponds to.

bzinberg: > Why shouldn't the implicit sample space of `labeled_cat` be all objects, given that its argument could be changing from execution to execution?

I think you answered this question in the trace types paper?  We're trying to allow upper bounds on the argument data type of `logpdf` and friends.

alex-lew: > If we want restrict_sample_space(μ) to still be a probability measure, then this is the same thing as suggesting that we add the ability to condition on arbitrary predicates.

Sure -- we could say that `restrict_sample_space(predicate)` actually restricts the sample space to the _union_ of the support and the predicate, which doesn't then require any conditioning. Or we could use a more accessible name like `add_support_assertion`. 

In general, I think anything that makes reasonable sense to us as an implementation strategy can be given a reasonable mathematical description. The most important bit to me is that we should think about what default behavior would be most useful for users.

bzinberg: > And in practice, would 0-logpdf not serve as a red flag?

I don't think it always should -- e.g. why should it be considered bad practice for the user to declare a categorical random variable where some of the weights (perhaps computed from other runtime quantities, or via a neural net) are zero?

alex-lew: > I think you answered this question in the trace types paper? We're trying to allow upper bounds on the argument data type of logpdf and friends.
 
In the trace types paper, we required that the support matched the sample space (we made no distinction).  For example, the trace types paper actually forbids a categorical variable with some of the weights being zero. Gen has typically forbidden it too, unless you make sure that the address of the random choice clearly indicates the support (in which case you should still never hit an `-Inf` score). 

The trace types paper is one especially restrictive point in the design space. Some of this was because the space restrictions in the paper didn't allow us to go into detail about designs that were less restrictive. I think this discussion is related to the trace types paper, but that Gen shouldn't necessarily take its cues from STPL.

bzinberg: > What are realistic errors someone might make that we want to prevent against? Mis-spelled observations?

Misspelled is one case.  Improperly formatted data set is another.  Another class of cases is bugs in the code that computes an `x` and evaluates `logpdf(x)`.  Similar to how we may want to hear complaints if we try to evaluate `[1, 2, 3] + 4`, we may want to hear complaints if we evaluate `logpdf(x)` where `x` is a string and the categorical distribution is defined over integers.

marcoct:
> I don't think it should -- e.g. why should be considered bad practice for
the user to declare a categorical random variable where some of the weights
(perhaps computed from other runtime quantities, or via a neural net) are
zero?

If the probabilities are zero I would guess that is usually due to
underflow. Also, in general zero probabilities can result in NaNs and
undefined behavior in inference algorithms. So I generally try to avoid
using zero probabilities unless they are actually really necessary (which I
think they often are not because you can e.g. smooth the distribution a
tiny bit without substantially changing it's semantics). Smoothing the
probabilities is good because it doesn't change the semantics of the model
(much) while simplifying the structure of the computational process.

bzinberg:
> If the probabilities are zero I would guess that is usually due to underflow. Also, in general zero probabilities can result in NaNs and undefined behavior in inference algorithms. So I generally try to avoid using zero probabilities unless they are actually really necessary (which I think they often are not because you can e.g. smooth the distribution a tiny bit without substantially changing it's semantics). Smoothing the probabilities is good because it doesn't change the semantics of the model (much) while simplifying the structure of the computational process.

I think weights for a categorical should always be specified by logprobs, never by probs, and any numerical issues resulting from using probs (not including issues that would have still shown up even if logprobs had been used) are a user error.  @marcoct, do you agree with this?  And if so, how much of the comment above still applies?

marcoct:
> I think weights for a categorical should always be specified by logprobs,
never by probs ...

I don't think that's right. I think that requiring users to specify log
probabilities is very onerous. Versions of distributions that accept
logspace params are useful, because they are more numerically stable, but I
don't think they should exclusively be used.

bzinberg:
> Versions of distributions that accept logspace params are useful, because they are more numerically stable, but I don't think they should exclusively be used.

For a one-off script, I agree it's nice to be able to specify probs, especially for pedagogical purposes.  But I think Gen should not take on responsibility for putting band-aids on issues that wouldn't have happened if the user had passed in logprobs.  It is a slew of potentially cumbersome-to-track-down numeric issues that in many cases are confined to user code that Gen literally can't fix.

Perhaps less controversial, I think Gen's core libraries and "non-user code" should use logprobs everywhere.

marcoct:
> Perhaps less controversial, I think Gen's core libraries and "non-user code" should use logprobs everywhere.

Agreed, with the exception that in some cases other representations of a probability are more numerically accurate, like for probabilities close to 1, and in cases where the probabilities are not small (like tests).

marcoct:
Since this PR makes Gen's behavior more consistent, and significantly expands the tests for distributions, I think we can merge it.

We can open separate issues to discuss the bigger design decisions around support.

src/modeling_library/dist/relabeled_distribution.jl
ztangent: @bzinberg it's worth noting that `logpdf` for `categorical` currently only accepts `Integer` values for `x`, so a type error will be thrown if you pass in a string!

Currently, user-defined distributions defined by the `@dist` DSL don't support this sort of type checking, but it looks like @alex-lew nicely implemented things so that it *should* be possible with a minor modification for type annotations to be supported when defining `@dist` distributions, e.g.:

```julia
@dist typed_labeled_cat(labels::String, probs) = labels[categorical(probs)]
```

I'll create an issue to note that typing is not possible right now.

bzinberg (changes requested):
@marcoct, @ztangent, here are the things that I think should be addressed before merging:

1. This line:
> We can open separate issues to discuss the bigger design decisions around support.

I agree with the fact that it's good to have those issues.  Key question is, who will open them?  In this case we are (now) discussing it explicitly.  Process-wise, I suspect in many cases, the important work of moving the unresolved discussion points into issues would fall through the cracks.  What guidelines should we set for the project to make sure issues raised during a PR discussion that aren't addressed by the eventual submission don't get lost when the PR closes?

2. @ztangent, I appreciate the attention you're paying to the current behavior of `categorical`, and the reasoning you're doing about what would be a better behavior.  For me, *the key point is that the current behavior is undocumented, and thus unspecified.*  Therefore I think this PR should not change from one unspecified behavior to another unspecified behavior, with notes about what the new behavior is occurring only in the PR discussion between maintainers.  Instead I think the PR should document any behavior that we want users to be able to use.  We should not force users to rely on unspecified behavior.

I have a pretty strong feeling about the above point, so would appreciate if we can discuss it before merging the PR.

marcoct:
>  What guidelines should we set for the project to make sure issues raised during a PR discussion that aren't addressed by the eventual submission don't get lost when the PR closes?

@bzinberg Thanks for being careful about this. How about we ensure that issues are created before merging? In this case there were three reviewers added, so the responsibility for this is unclear. But I am happy to do it in this case. (Update: I see you have already done this).

@bzinberg  @ztangent @alex-lew Regarding support of distributions and sample space:

There is only a distinction between sample space and support when the support is not the "full" sample space. In the existing Gen usage and documentation, a "disciplined" modeling approach involves associating a fixed support with each address. That guarantees that 0 probability densities never show up. However, that makes some things inconvenient, which is why previously, 0 probability densities were sometimes allowed, and why, in this PR, they are allowed more consistently, instead of triggering an error.

I think it makes sense to admit a separate notion of sample space and support into Gen, where by "sample space" we mean the Julia type for the value of the random choice accepted by the `logpdf` Julia function, and by support we mean the subset of values of that type that have nonzero density. There are some meaningful differences in how they interact with the implementation -- e.g. the "sample space" is a static property of the distribution, and the support is not (it depends on the arguments).

Then, a "disciplined" generative function will associate addresses with fixed support. That is encouraged but not required. Gen will not enforce this by throwing errors when zero probability densities are encountered.

The documentation can distinguish between "disciplined" models and models that don't have the static support property. I think this would reflect the existing behavior of Gen, which is also the behavior of Gen in this PR (but now it is more consistent ).

For reference, the relevant piece of existing documentation is [here](https://probcomp.github.io/Gen/dev/ref/modeling/#Choices-should-have-constant-support-1). It is a somewhat ambiguous about whether disciplined models are required or not (it uses the word "must" and then says that if you don't do it, some things might not work). We can improve it for this PR:

marcoct:
How about this for initial documentation that is a slight improvement on the existing documentation?

### Sample space and support of random choices

Each address is associated with a fixed **sample space**, that is the Julia type associated with the the Probability Distribution sampled at the choice. If different distributions are used with the same address, they must share the same sample space. (Technically, the sample space is a 'measure space'). The sample space is represented by the Julia type of the first argument to the distribution's [`logpdf`](@ref) method.

Given specific arguments to the distribution, **support** of a random choice is the set of values with nonzero probability (density). This may include the entire sample space, or it may be a subset.

A generative function can be **disciplined** or not. In a disciplined generative function, the support of random choices at each address must be fixed. That is, there exists a set S that is a subset of the sample space such that for all executions of the generative function, if an address appears in the execution, its support must be S. Violating this discipline will cause NaNs, errors, or undefined behavior in certain inference programs. However, there are many inference programs where violating this discipline does not result in undefined behavior and is convenient.

If the support of a random choice needs to change, a disciplined generative function can represent this by using a different address for each distinct value of the support. For example, consider the following generative function:
```julia
@gen function foo()
    n = @trace(categorical([0.5, 0.5]), :n) + 1
    @trace(categorical(ones(n) / n), :x)
end
```
The support of the random choice with address `:x` is either the set ``\{1, 2\}`` or ``\{1, 2, 3\}``.
Therefore, this random choice does satisfy our condition above.
This would cause an error with the following, in which the `:n` address is modified, which could result in a change to the support of the `:x` variable:
```julia
tr, _ = generate(foo, (), choicemap((:n, 2), (:x, 3)))
tr, _ = mh(tr, select(:n))
```
We can modify the address to satisfy the condition by including the support in the address:
```julia
@gen function foo()
    n = @trace(categorical([0.5, 0.5]), :n) + 1
    @trace(categorical(ones(n) / n), (:x, n))
end
```

marcoct:
To be clear, in the version of behavior I am proposing above, logpdf should never error. The only errors should be type errors if the value is of the not type expected by logpdf (actually I guess these will be no method found errors, which is unfortunate; it is probably straightforward to write a defaut logpdf method that has an error message that indicates that the issue is with the type of a random choice being wrong).

marcoct:
@bzinberg @alex-lew @ztangent Upon thinking more, and reading above comments again, it probably makes more sense to allow the sample space to be a subset of the Julia type expected by logpdf. That would allow additional user control over their "sample space", without having to define a complicated Julia type.

So then, there would be three sets of decreasing size (each is a subset of the previous):

- The set of values for which the Julia type matches. (This is a static property of the Distribution).

- The set of values for which logpdf does not raise an runtime error (this is the sample space / domain; which technically is defined by the distribution's measure space). (I think ideally this would be a static property of the distribution).

- The support.  (This can be dynamic i.e. it can depend on the arguments)

We can discuss this more in https://github.com/probcomp/Gen/issues/206, but for the purposes of this PR, I would be fine with documenting a version where the logpdf is allowed to error. Then, users can choose to use distributions with different behavior depending on their needs -- i.e. that are more or less strict regarding how they define the measure space and whether they automatically extend it  or not. I think that allowing for multiple modes of use (more strict and less strict) is appropriate at this point.

bzinberg:
> @bzinberg Thanks for being careful about this. How about we ensure that issues are created before merging?

Sounds good!

bzinberg:
Thanks @marcoct for writing those docs, they look great!  I have a couple of minor comments, could you push the docs to this branch so I can drop them in?

ztangent:
@bzinberg should I add the docs to this PR? I don't think @marcoct can do it since it's on my fork of Gen.

@marcoct I'm confused by "I would be fine with documenting a version where the logpdf is allowed to error." -- do you mean that `logpdf` will return `-Inf` for out-of-support values and only throw an error if there's a type error? Or do you mean that `logpdf` will throw an error instead of returning `-Inf`? Or do you mean that in the documentation, we just tell users that errors may occur if the support changes, even though in some cases, no error will in fact be thrown?

I also wanted to note that, with the changes made in this pull request, the following example from the docs will no longer throw an error.

```julia
tr, _ = generate(foo, (), choicemap((:n, 2), (:x, 3)))
tr, _ = mh(tr, select(:n))
```

Instead, `mh` will always reject the newly proposed trace, because its weight will be `-Inf`. And so `tr` will be always be stuck with its initial choicemap (because there are no other possible values of `:n` that could make `:x = 3`.

marcoct:
>  @bzinberg should I add the docs to this PR? I don't think @marcoct can do it since it's on my fork of Gen.

@ztangent Can I draft a version and send it to you? Or you can give me push access to your fork.

bzinberg:
@ztangent

> @bzinberg should I add the docs to this PR? I don't think @marcoct can do it since it's on my fork of Gen.

I believe that by default, when you send a PR, the assigned reviewers get permission to push to the branch.

marcoct:
> @marcoct I'm confused by "I would be fine with documenting a version where the logpdf is allowed to error." -- do you mean that logpdf will return -Inf for out-of-support values and only throw an error if there's a type error? Or do you mean that logpdf will throw an error instead of returning -Inf? Or do you mean that in the documentation, we just tell users that errors may occur if the support changes, even though in some cases, no error will in fact be thrown?

The docs will need to distinguish between sample space and support, and "disciplined" v "undisciplined" generative functions. It boils down to three pieces

1. If you give a value that is in the sample space but not in the support then logpdf will return -Inf.

2. If you give a value that is not in the sample space, then sometimes this will result in a type error, and sometimes it will results in a runtime error (`error(..)`).

3. If you want to write a "disciplined" generative function then you should ensure addresses have constant support. But the built-in modeling languages will not prevent you from writing and using undisciplined generative functions. (Some user inference programs may crash on these functions -- better characterizing what inference programing patterns require disciplined functions is for a future PR).

A `Distribution` will need to have a well-defined sample space and support in their documentation (the support can be a function of the arguments to the distribution, and maybe the sample space can be as well?). Currently, these are not explicitly represented, but they could be in the future.

ztangent:
@marcoct just invited you as a collaborator to my fork! in the future I'm happy to develop on branches within the probcomp repo itself, I just don't have the access rights for that right now.

bzinberg:
@ztangent, did Marco not automatically get push access to this branch when he became an assigned reviewer on the PR?  [Looks like I did](https://github.com/probcomp/Gen/pull/203/commits/83b8b33da7707949bbf6d8439ec487bac9d5b43c) :slightly_smiling_face:

ztangent:
> A `Distribution` will need to have a well-defined sample space and support in their documentation (the support can be a function of the arguments to the distribution, and maybe the sample space can be as well?). Currently, these are not explicitly represented, but they could be in the future.

Cool, this makes sense. If this is how we're going forward, I think we should also document the sample spaces of the current built-in distributions. With the changes in this PR, all the 1D continuous built-in distributions will have `Real` as their sample space, discrete built-in distributions will have `Int` as their sample space, and multivariate continuous distributions will have `AbstractVector{Real}` as their sample space.

@bzinberg nope, no automatic push access, as far as I can tell. EDIT: oh nvm, I see that you managed to push!

docs/src/ref/modeling.md
```diff
@@ -97,31 +97,46 @@ val::Bool = @trace(bernoulli(0.5), :z)
...
     n = @trace(categorical([0.5, 0.5]), :n) + 1
-    @trace(categorical(ones(n) / n), (:x, n))
+    @trace(categorical(ones(n) / n), :x)
 end
 ```
+The support of the random choice with address `:x` is either the set ``\{1, 2\}`` or ``\{1, 2, 3\}``.
+Therefore, this random choice does not have constant support, and the generative function `foo` is not 'disciplined'.
+Specifically, this could result in undefined behavior for the following inference program:
+```julia
+tr, _ = importance_resampling(foo, (), choicemap((:x, 3)))
```
ztangent, line 138: Thanks for writing up these docs @marcoct! They provide a lot of clarity. My only issue is maybe with this example of possible undefined behavior. With the changes made in this pull request, I tested this exact example, and `importance_resampling` correctly does rejection sampling and returns only traces with with `tr[:n] = 2` (which is the only way that `tr[:x]` could have been `3`:
```julia
julia> traces = [importance_resampling(foo, (), choicemap((:x, 3)), 100)[1] for i=1:1000];
julia> all(tr -> tr[:n] == 2, traces)
true
```
I'm confident that `importance_resampling` will correctly reject `-Inf` traces in other cases as well, based on how the current implementation is written. Specifically, it will never enter this branch of the code for a `-Inf` trace:
```
# Line 102 of src/inference/importance.jl
        if bernoulli(exp(log_weight - log_total_weight))
            model_trace = cand_model_trace
        end
```
So it seems overly conservative to say that `importance_resampling` will result in undefined behavior. Unless you're referring to the fact that very (very) rarely, `importance_resampling` will return a `-Inf` trace, because it never manages to propose a trace with non-zero probability.

marcoct: > Unless you're referring to the fact that very (very) rarely, importance_resampling will return a -Inf trace, because it never manages to propose a trace with non-zero probability.

@ztangent Yes, that's what I'm referring to.

marcoct: And it's not clear that it should return anything in that case -- I'm not sure what the behavior is right now, but I think maybe it should just crash. We could define the behavior to just return a random trace if it receives all -Inf weights but that seems not obvious, and the same sort of decision would need to be made in various other inference primitives (e.g. primitives SMC) as well as by user inference programs (which may not be as careful as the built-in primitives in their handling of -Infs and NaNs).

ztangent: I am a fan of removing the original suggested fix though, where we include the support in the trace address:
```julia
@gen function foo()
    n = @trace(categorical([0.5, 0.5]), :n) + 1
    @trace(categorical(ones(n) / n), (:x, n))
end
```
The reason is that if you define, `foo` this way, and then do:
```julia
tr, _ = importance_resampling(foo, (), choicemap((:x, 3) => 3), 100)
```
what ends up happening is that the constraint `(:x, 3) => 3` ends up being totally ignored when :n is sampled to be 1. Effectively, this means that `:n=1` ends up being sampled with relative probability `0.5`, while `:n=2` ends up being sampled with relative probability `0.5 * 1/3` (because all traces with `:n=2` are constrained to satisfy `(:x, 3)=3`, and end up being down-weighted by 1/3).

IMO this is very unexpected behavior! It certainly was to me when I first tried to do something like this in Gen. And I don't think this sort of behavior is well documented either. What I *would* have expected is that all traces with `:n=1` be rejected, because they fail to ensure that `(:x, 3)` exists in the trace.  But Gen currently doesn't have logic that checks this*. So to me, the undisciplined version of `foo` actually exhibits more well-defined behavior than the disciplined version, because `importance_resampling` correctly does rejection sampling (with the changes in this PR).

*And I understand that adding the logic to check this might mess up some other things in Gen, because sometimes Gen allows unsampled addresses to be in the choicemap, as the current implementation of `regenerate` allows for.

docs/src/ref/modeling.md
```diff
@@ -97,31 +97,46 @@ val::Bool = @trace(bernoulli(0.5), :z)
...
 
-### Choices should have constant support
-The support of a random choice at a given address (the set of values with nonzero probability or probability density) must be constant across all possible executions of the `@gen` function.
-Violating this discipline will cause errors in certain cases.
-If the support of a random choice needs to change, use a different address for each distinct value of the support.
-For example, consider the following generative function:
+
+### Sample space and support of random choices
+
+Different probability distributions produce different types of values for their random choices. For example, the [`Bernoulli`](@ref) distribution results in `Bool` values (either `true` or `false`), the [`Normal`](@ref) distribution results in `Real` values that may be positive or negative, and the [`Beta`](@ref) distributions result in `Real` values that are always in the unit interval (0, 1).
```
bzinberg, line 103:
```suggestion
Different probability distributions produce different types of values for their random choices. For example, the [`Bernoulli`](@ref) distribution results in `Bool` values (either `true` or `false`), the [`Normal`](@ref) distribution results in `Real` values that may be positive or negative, and the [`Beta`](@ref) distributions result in `Real` values that are always in the unit interval [0, 1].
```

docs/src/ref/modeling.md
```diff
@@ -97,31 +97,46 @@ val::Bool = @trace(bernoulli(0.5), :z)
...
+
+Each [`Distribution`](@ref) is associated with two sets of values:
+
+- The **sample space** of the distribution, which does not depend on the arguments.
+
+- The **support** of the distribution, which may depend on the arguments, and is the set of values that has nonzero probability (or probability density). It may be the entire sample space, or it may be a subset of the sample space.
```
bzinberg, line 109: AFAIK, the usual definition of "support" is the topological closure of this set -- i.e., this set along with its boundary.  As one possible motivation for including the boundary, note that e.g. for the beta distribution, any neighborhood of a boundary point has positive measure -- this is not true of points outside the support.

docs/src/ref/modeling.md
marcoct: > IMO this is very unexpected behavior! It certainly was to me when I first tried to do something like this in Gen. And I don't think this sort of behavior is well documented either.

Yes. This is the behavior defined incredibly tersely in the current GFI documentation (in the subtle definition of the behavior of q), but it could be much better documented, at both the GFI level and the inference library level (like what importance resampling does).  But I think this is a question for a different issue (https://github.com/probcomp/Gen/issues/208). But briefly,  I think that importance resampling should probably crash if not all the constraints are visited, but that there should be a flag to enable or disable this check (Gen's current approach to not always performing dynamic checks of this form is to allow for improved performance.).  I think that the GFI should keep its behavior of ignoring irrelevant constraints (or have an option for being more stringent), and that inference primitives should have an option for enabling dynamic checks for this sort of thing -- this is the approach taken by the various MCMC primitives -- e.g. you can enable dynamic checks that your observations aren't being over-written.

docs/src/ref/modeling.md
```diff
@@ -97,31 +97,46 @@ val::Bool = @trace(bernoulli(0.5), :z)
...
-The support of the random choice with address `:x` is either the set ``\{1, 2\}`` or ``\{1, 2, 3\}``.
-Therefore, this random choice does satisfy our condition above.
-This would cause an error with the following, in which the `:n` address is modified, which could result in a change to the domain of the `:x` variable:
-```julia
-tr, _ = generate(foo, (), choicemap((:n, 2), (:x, 3)))
-tr, _ = mh(tr, select(:n))
-```
-We can modify the address to satisfy the condition by including the domain in the address:
+
+A generative function can be **disciplined** or not. In a disciplined generative function, the support of random choices at each address must be fixed. That is, there exists a set S that is a subset of the sample space such that for all executions of the generative function, if an address appears in the execution, its support must be S. Violating this discipline will cause NaNs, errors, or undefined behavior in certain inference programs. However, there are many inference programs where violating this discipline does not result in undefined behavior and is convenient.
```
bzinberg:
```suggestion
A generative function can be **disciplined** or not. In a disciplined generative function, the support of random choices at each address must be fixed. That is, for each address `a` there exists a set S that is a subset of the sample space such that for all executions of the generative function, if `a` occurs as the address of a choice in the execution, the support of that choice must be exactly S. Violating this discipline will cause NaNs, errors, or undefined behavior in some inference programs. However, there are many inference programs where violating this discipline does not result in undefined behavior and is convenient.
Authors of inference programs should document which kinds of undisciplined models their inference algorithms allow or expect to see.
```

docs/src/ref/modeling.md
ztangent: > And it's not clear that it should return anything in that case -- I'm not sure what the behavior is right now, but I think maybe it should just crash. We could define the behavior to just return a random trace if it receives all -Inf weights but that seems not obvious, and the same sort of decision would need to be made in various other inference primitives (e.g. primitives SMC) as well as by user inference programs (which may not be as careful as the built-in primitives in their handling of -Infs and NaNs).

So the behavior right now (with this PR) is that it'll return a trace that is inconsistent with the observations, with weight `-Inf`:
```julia
@gen function foo()
    n = @trace(categorical([1.0, 0.0]), :n) + 1
    @trace(categorical(ones(n) / n), :x)
end
```
```julia
julia> tr, w = importance_resampling(foo, (), choicemap((:x, 3)), 10);
julia> tr[:n], w
1, -Inf
```
I think one possibility is to catch `-Inf` and throw an error. Another possibility is to just do a `@warn`ing? I feel like one reason why you might want to avoid throwing an error is that external code that calls `importance_sampling` or other inference algorithms might want to detect when `-Inf` traces are returned then automatically increase the number of samples without having to do a `try/catch` block. But anyway, none of this has to go into this PR, just thought I'd confirm what the current behavior is and have it documented here.

bzinberg: @ztangent:

> So it seems overly conservative to say that importance_resampling will result in undefined behavior.

Aha, but (in my opinion) undefined is not about what the code will do, it's about whether the behavior is specified!  So if the spec says behavior for a certain kind of input is undefined, then it is truly undefined, even if a particular implementation always handles that input in a knowable way.

docs/src/ref/modeling.md
marcoct: OK. I this is an instance where if we are super precise, it may become unreadable to many users if all the detail is given. I think that the details can be put in parentheses in some cases, or in separate "mathematical details" sections. This sort of tension between readability and precision shows up everywhere in Gen's documentation. I think that a reasonable approach is to have a "mathematical details" subsection in various sections of documentation that gives all the precise details for potential users who want to know or expect to see these details.

docs/src/ref/modeling.md
marcoct: I don't think we should prescribe what authors of inference programs should do, since inference programs are user code, and are not necessarily documented at all.

marcoct: We could qualify this with "authors of reusable inference code should" ..

docs/src/ref/modeling.md
bzinberg: That sounds good to me.  I get that there is a trade-off between readability and precision.  I still think you should say the support of the beta distribution is [0, 1] though, not (0, 1).  Wdyt?

marcoct: I'm not sure. I just looked up on [Wikipedia](https://en.wikipedia.org/wiki/Beta_distribution) and saw this:

> However, the inclusion of x = 0 and x = 1 does not work for α, β < 1; accordingly, several other authors, including W. Feller,[5][6][7] choose to exclude the ends x = 0 and x = 1, (so that the two ends are not actually part of the domain of the density function) and consider instead 0 < x < 1. 

In general, I prefer to avoid boundary cases in the definitions of things when we can, to encourage users to avoid writing models that are unecessarily deterministic. For example, when the value sampled from a Beta distribution is used as a probability, it would be nice to know that it won't be exactly 0 or 1. But I chose (0, 1) in this case mainly as a nudge to encourage a way of thinking about models -- whether the distinction manifests in practice is not obvious (i.e. being exactly 0 or 1 is zero probability) and how the distinction is implemented is complex and involves characterizing behavior in floating points, which is a big discussion of its own.

bzinberg: Ah, I didn't realize that some authors say the support of the beta distribution is (0, 1).  Cool, fine with me to keep it that way then.

There is of course a trickiness in trying to make precise the intuitions you're advocating above -- for example, the value of a density function at any particular point can be changed and the result will still be a valid density function for the same distribution.  I agree that it makes sense to intentionally not go into that level of detail in the Gen docs, unless the issue makes a real difference in what is or isn't possible to do in Gen.

docs/src/ref/modeling.md
bzinberg: Sure.
```suggestion
A generative function can be **disciplined** or not. In a disciplined generative function, the support of random choices at each address must be fixed. That is, for each address `a` there exists a set S that is a subset of the sample space such that for all executions of the generative function, if `a` occurs as the address of a choice in the execution, the support of that choice must be exactly S. Violating this discipline will cause NaNs, errors, or undefined behavior in some inference programs. However, there are many inference programs where violating this discipline does not result in undefined behavior and is convenient.
Authors who want their inference code to be reusable should consider documenting which kinds of undisciplined models their inference algorithms allow or expect to see.
```

docs/src/ref/modeling.md
```diff
@@ -97,31 +97,47 @@ val::Bool = @trace(bernoulli(0.5), :z)
...
-This would cause an error with the following, in which the `:n` address is modified, which could result in a change to the domain of the `:x` variable:
-```julia
-tr, _ = generate(foo, (), choicemap((:n, 2), (:x, 3)))
-tr, _ = mh(tr, select(:n))
-```
-We can modify the address to satisfy the condition by including the domain in the address:
+
+A generative function can be **disciplined** or not. In a disciplined generative function, the support of random choices at each address must be fixed. That is, for each address `a` there exists a set S that is a subset of the sample space such that for all executions of the generative function, if `a` occurs as the address of a choice in the execution, the support of that choice must be exactly S. Violating this discipline will cause NaNs, errors, or undefined behavior in some inference programs. However, there are many inference programs where violating this discipline does not result in undefined behavior and is convenient.
+Authors who want their inference code to be reusable should consider documenting which kinds of undisciplined models their inference algorithms allow or expect to see.
```
bzinberg: Clarifying the phrasing a bit (note that the first para has phrasing changes too, but github doesn't know how to highlight the diff)
```suggestion
A generative function can be **disciplined** or not. In a disciplined generative function, the support of random choices at each address must be fixed. That is, for each address `a` there must exist a set S that is a subset of the sample space such that for all executions of the generative function, if `a` occurs as the address of a choice in the execution, then the support of that choice is exactly S. Violating this discipline will cause NaNs, errors, or undefined behavior in some inference programs. However, in many cases it is convenient to write an inference program that operates correctly and efficiently on some specialized class of undisciplined models.
In these cases, authors who want their inference code to be reusable should consider documenting which kinds of undisciplined models their inference algorithms allow or expect to see.
```

bzinberg (approved):
The changes I asked for before merging the PR have been satisfied, so I'm no longer requesting to delay the merge.  Left a couple of small rephrasing suggestions, then I'm good with this.  Note, I only reviewed the spec and the comments on the discussion page, haven't looked at the code at all :slightly_smiling_face:

marcoct (approved):

PR #207, opened 2020-03-10
Add contributor guidelines to make sure issues raised in PR discussion don't get buried

bzinberg:
As discussed in https://github.com/probcomp/Gen/pull/203#issuecomment-597263077

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,3 +13,20 @@
 ```sh
 ./docs/build_docs_locally.sh
 ```
+## Submitting Pull Requests (PRs)
+
+- Include documentation in the same PR as the code.
+
+- Make sure that important info doesn't stay siloed in a PR discussion, since
+  users won't see that.
+
+  + If a reviewer asks for clarification about what the code does or is
+    intended to do, this often reveals that some info that future users may
+    want to know, is missing.  Rather than just respond to the PR discussion
+    comment, put that information in the docs.
+
+  + Often, concerns are raised in a PR discussion that lead to a down-scoping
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -13,3 +13,20 @@
...
+
+- Include documentation in the same PR as the code.
+
+- Make sure that important info doesn't stay siloed in a PR discussion, since
+  users won't see that.
+
+  + If a reviewer asks for clarification about what the code does or is
+    intended to do, this often reveals that some info that future users may
+    want to know, is missing.  Rather than just respond to the PR discussion
+    comment, put that information in the docs.
```
marcoct: I would revise this to:

> Rather than just respond to the PR discussion comment, consider putting that information in the docs.

I think there some cases where clarification questions might be based on presumed conceptual or mathematical frameworks or misconceptions that are not generally held by the audience. This could be due to e.g. the reviewer being more familiar with engineering than modeling and inference, or vice versa, depending on different combinations of reviewer and content, or the reviewer expecting a clarification using mathematical terminology that is not accessible to the audience of Gen. Introducing these terms into the documentation might not be appropriate, because the documentation could end up being heavily tailored to the reviewer's particular way of thinking. I think we should try to make a best effort to document in response to clarifications, but there should be some flexibility regarding what actually ends up in the docs.

marcoct: Put another way, information shared on a PR is a discussion between individuals, and what is put on the documentation is something tailored for the general audience of Gen. Those can be different.

bzinberg: makes sense

marcoct (approved):

Issue #206, opened 2020-03-10
Decide on a semantics of "support", "sample space", "extend by zero", and "restrict", and their implications for where/whether `logprob` should return `-Inf` or throw error


bzinberg:
Continuation thread for the discussion started in https://github.com/probcomp/Gen/pull/203#discussion_r389749694.

fsaad:
If the main character is "logprob" i.e., the behavior of the probability density function p : X → [0, ∞), then one standard terminology would be:
- domain: the set of values (X) on which the function p is defined;
- support: {x | p(x) > 0}.

"Sample space" usually refers to an abstract set Ω (e.g., the [0,1] or {0,1}^∞ is a canonical choice, endowed with the Lebesgue measure), on which every random variable Z : Ω → X is defined.  In Gen's case this space might be the sequence of bits generated by the (P)RNG.  There is rarely any interest in considering a sample space Ω other than [0,1], see e.g., Kallenberg Thm 3.19, Lemma 3.22, Thm 6.14, 6.16, and 6.17.

It seems logical at minimum to permit calling p(x) for any element x in the domain X, including those outside of the support.  Or you might chose to restrict the domain to the support, which also is sensible provided it is documented clearly, but results in a less general strange semantics of the CDF as being the the integral of the PDF for continuous distributions.

Whether calling p(x) for an x not in the domain, e.g., p("hello world") for p Gaussian or p(1.123) for p Categorical({"hello":0.4, "world":0.6}), results in "Error" or "0" is perhaps more debatable, since one can extend p to the universal set in the natural way if they have a good reason to do that.

marcoct:
Copying and adapting from https://github.com/probcomp/Gen/pull/203

Each address is associated with a fixed measure space, which defines the 'base measure' for the random choices. We can decide to call the set of values in this space the "sample space" or the "domain" (maybe domain is more standard?). A `Distribution` type is associated with a base measure. The only non-trivial commitment here is to require that addresses are associated with fixed base measures, which is important.

It has also been helpful in Gen to have a notion of a "disciplined" generative function for which the **support** at an address is constant. But it is becoming more clear that is probably too restrictive to require this (or enforce it by throwing errors). However, I think that the notion of a disciplined generative function with constant support at addreses should be kept.

[...]

PR #212, opened 2020-03-13
Support type annotations in the @dist DSL.

ztangent:
Currently the `@dist` DSL has limited support for type checking. With the changes in #203, this makes it slightly more dangerous to use user-defined distributions such as:
```julia
@dist labeled_cat(labels, probs) = labels[categorical(probs)]
```
because `labels` could be vector of anything, so the sample space of `labeled_cat` ends up being `Any`, meaning something like `logpdf(labeled_cat, "c", [:a, :b, :c], [0.3, 0.3, 0.4])` will return `-Inf`, rather than erroring.

Sometimes behavior like the above may be useful, if we genuinely do want to sample from a categorical that may have mixed data types. But in lieu of solving the issues raised in #206 (which may require us to distinguish the sample space from the values that are of the correct type), we can still allow users of the `@dist` DSL to be program more defensively by supporting type annotations. This PR enables such functionality, allowing us to write, for e.g.:
```julia
@dist symbol_cat(labels::Vector{Symbol}, probs) = labels[categorical(probs)]
```
`MethodError`s will then be thrown if we try to do either of the following:
```julia
symbol_cat(["a, "b"], [0.5, 0.5]) #Wrong parameter type
logpdf(symbol_cat, "a", [:a, :b], [0.5, 0.5]) #Wrong output type
```

Implementation-wise, type annotations are first captured and stored as a type parameter when `SimpleArg`s are constructed by the `@dist` macro. Type checking is then enforced in two places -- the first in calls to `eval_arg` for `SimpleArg`s, and second in how operators such as `getindex` propagate type information to the distributions they construct.

This PR also reformats the `@dist` DSL code to be a little more readable by trying to stick to the 80-char limit.

### DIFF
--- a/src/modeling_library/dist/dist_dsl.jl
+++ b/src/modeling_library/dist/dist_dsl.jl
@@ -5,16 +5,15 @@ include("relabeled_distribution.jl")
 
 #TODO: Remove Distribution{T} from structs; use a type U instead
 
-
 # `Arg` types: represent arguments to distributions
-abstract type Arg end
+abstract type Arg{T} end
 
 # Represents a argument at a certain index
-struct SimpleArg <: Arg
+struct SimpleArg{T} <: Arg{T}
     i :: Int8
 end
 
-struct TransformedArg <: Arg
+struct TransformedArg{T} <: Arg{T}
     f_args :: Vector{Arg} # Is it better to do Union{TransformedArg, SimpleArg}?
     orig_f :: Function
     arg_passer :: Function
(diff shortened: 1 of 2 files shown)


src/modeling_library/dist/dist_dsl.jl
```diff
@@ -145,24 +143,20 @@ end
...
   arguments = fndef[:args]
   body = fndef[:body]
 
   name_to_index = Dict{Symbol, Int8}()
+  name_to_type = Dict{Symbol, Type}()
   for (i, arg) in enumerate(arguments)
-      (argname, _, _, _) = splitarg(arg)
+      (argname, argtype, _, _) = splitarg(arg)
       name_to_index[argname] = i
+      name_to_type[argname] = eval(argtype)
```
ztangent: I just realized that one potential issue here is that this may perform the `eval` with respect to Gen, and not the calling module...? And so user-defined types may be ignored...?

alex-lew: Yeah -- since you're just embedding this type into generated code anyway, why not just use `argtype` directly (rather than `eval(argtype)`)?

ztangent: Just changed this line to `__module__.eval(argtype)`, which works as expected! I think the reason I can't use `argtype` directly is because `SimpleArg`s are constructed within the `@dist` macro, and so I need to provide the type parameter then and there.

src/modeling_library/dist/dist_dsl.jl
```diff
@@ -145,24 +143,20 @@ end
...
   end
 
   function process_node(node)
       if typeof(node) == Symbol && haskey(name_to_index, node)
-          return :($(SimpleArg(name_to_index[node])))
+          return :($(SimpleArg{name_to_type[node]}(name_to_index[node])))
```
alex-lew, line 159: Right -- I think I meant that you could do `:(SimpleArg{type}($(name_to_index[node])))` -- so that the SimpleArg is created in the generated code, rather than at compile-time. But if `__module__.eval` works, that's OK too!

ztangent: Mm right, I tried doing that then got errors, but I just realized it's because I didn't do `Gen.SimpleArg` instead of just `SimpleArg`!

marcoct (approved):

Issue #213, opened 2020-03-13
Extend static modeling language syntax


marcoct:
The implementation of the SML (static modeling language) is currently broken down into stages:

- an intermediate representation (IR) based on a DAG (https://github.com/probcomp/Gen/blob/master/src/static_ir/static_ir.jl)

- a surface syntax that resembles Julia function definition and generates this IR (https://github.com/probcomp/Gen/blob/master/src/dsl/static.jl)

There are many extensions to the surface syntax that would make it less restricted, that would not require changes to the underlying IR. This includes:

- supporting more general types of expressions including `y = @trace(normal(0, 1), :x) + 1` as suggested by @ztangent in https://github.com/probcomp/Gen/issues/90, as well as expressions that include multiple `@trace` calls on the right-hand-side, and let expressions.

- restricted classes of for loops and list comprehensions (could be converted to a map combinator). (supporting if-else-end branching in this way would require introducing a branching combinator; which should be a separate issue)

- relaxing the static single assignment restriction as suggested by @alex-lew (the language implementation would convert to SSA when generating the IR)

- possibly, supporting macros as suggested in https://github.com/probcomp/Gen/issues/185

I am putting these extensions into one issue, because I think that the SML surface syntax code could probably be redesigned with this set of capabilities in mind, while provisioning for yet additional extensions in the future.

marcoct:
A revisiting of the SML code should also address the poor error messages that are currently thrown during syntax errors involving `@trace`: https://github.com/probcomp/Gen/issues/90

AriMKatz:
Would loops based on `foldl` help, by preserving more structure? see discussion here https://github.com/JuliaLang/julia/pull/35074#issuecomment-597803494

ztangent:
I've started taking a stab at this in [`ztangent:sml-refactor`](https://github.com/ztangent/Gen/tree/sml-refactor)! Right now I've just switched over some of the pattern-matching logic to use `@capture` from `MacroTools` so it's a little neater, but I've a good sense of how the parsing code is like now and how it might need to change.

ztangent:
Partially addressed by #214! Still unaddressed:
- Support for restricted classes of for loops and list comprehensions
- Support for macro calls

femtomc:
One comment I noticed about the extensions outlined above: it looks like they could be addressed by an SSA form analysis and transformation using the IRTools representation (before constructing the static DAG representation). 

[...]

Issue #90, opened 2019-03-26
Using `@trace` incorrectly in a static gen function gives confusing error message.


marcoct:
```julia
using Gen

@gen (static) function foo()
    y = @trace(normal(0, 1), :x) + 1
    return y
end

load_generated_functions()

println(foo())
```

The issue is that the statement `y = @trace(normal(0, 1), :x) + 1` fails to parse as a `@trace` statement because of the `+ 1` at the end. Therefore, we attempt to parse it as a Julia statement, which succeeds (because we do not check the expression for `@trace`). Then, at run-time, the `@trace` macro gets expanded into a call to `traceat` that whose first argument is the gensym `##state#360`, which is only provided in the context of a dynamic `@gen` function, and is undefined in regular Julia code.

Some thought about how to handle macros in Julia code, and how to handle `@trace` in particular, is needed. We could disallow all macros, but that would seem overly restrictive. We could search the AST for `@trace` and throw a syntax error, but that seems like a special case and brittle.

Error message: 
```
ERROR: LoadError: UndefVarError: ##state#360 not defined
Stacktrace:
 [1] (::getfield(Main, Symbol("##3#4")))() at ./none:0
 [2] macro expansion at /home/marcoct/dev/Gen/src/static_ir/simulate.jl:87 [inlined]
 [3] simulate at /home/marcoct/dev/Gen/src/static_ir/simulate.jl:87 [inlined]
 [4] propose at /home/marcoct/dev/Gen/src/gen_fn_interface.jl:202 [inlined]
 [5] (::getfield(Main, Symbol("##StaticGenFunction_foo#364")))() at /home/marcoct/dev/Gen/src/static_ir/static_ir.jl:54
 [6] top-level scope at none:0
 [7] include at ./boot.jl:317 [inlined]
 [8] include_relative(::Module, ::String) at ./loading.jl:1044
 [9] include(::Module, ::String) at ./sysimg.jl:29
 [10] exec_options(::Base.JLOptions) at ./client.jl:231
 [11] _start() at ./client.jl:425
in expression starting at /home/marcoct/dev/Gen/test.jl:10
```

This issue was encountered by @fsaad

ztangent:
Perhaps a less brittle approach would be to lift the restriction that `@trace` be the outermost expression on the RHS of an assignment, and instead create a new variable in the IR for every function call involving `@trace` as an argument? This would mean the following two definitions lead to the same static computation graph (up to variable renaming):

```julia
[...]

PR #214, opened 2020-03-17
Refactor parsing for static modeling language and extend supported syntax.

ztangent:
Addresses #90 and parts of #213. Opening this pull request for purposes of discussion and collaboration -- I don't think this is quite ready to be merged yet.

With the changes in this PR,  the static modeling language can now:
- Handle expressions that include multiple `@trace` calls on the right-hand-side:
```julia
@gen (static) function circle(r, t, std)
    pt = (@trace(normal(r*cos(t), std), :x), @trace(normal(r*sin(t), std), :y))
    return pt
end
@test circle(1, 0, 0) == (1, 0)
```

- Support re-assignment by relaxing the static single assignment restriction:
```julia
@gen (static) function foo(x)
    x = x + 1
    return x
end
@test foo(0) == 1
```

- Support tuple assignments:
```julia
@gen (static) function foo(params::Tuple)
    p, (mu1, sigma1), (mu2, sigma2) = params
    mu, sigma = @trace(bernoulli(p), :x) ? (mu1, sigma1) : (mu2, sigma2)
    return @trace(normal(mu, sigma), :y)
end
@test foo((1, (1, 0), (0, 0)) == 1
```
- Return arbitrary purely functional Julia expressions (including those with `@trace` calls)
- Catch and throw errors at parse time for a number of unsupported constructs (e.g. `@trace` calls which are non-addressed, argument splatting within `@trace` calls).

Most of these changes were implemented be refactoring the code to process ASTs bottom-up using `MacroTools.postwalk`, instead of the previous top-down approach. This allows us to match all `@trace` calls that are nested in other expressions and add a node to the Static IR for each of them. The `@trace` calls are then replaced by fresh variable names which are bound to those nodes*.

(*This changes the naming convention for Static IR nodes -- `@trace` nodes were previously named according to the variables they were assigned to, but now are named according to their addresses.)

In order to support more of the syntax extensions suggested in #213 (e.g. for loops and let expressions), it seems like the bottom-up parsing approach used in this PR will be insufficient, because the meaning and variable scope of (for e.g.) an `@trace` call in a for-loop is context-dependent. One possible approach to addressing this is to do an initial top-down pass (i.e. `MacroTools.prewalk`) through the AST as a preprocessing step.

For example, the top-down pass could rewrite for-loops with trace calls using some specialized syntax, which can then be matched and handled appropriately during the bottom-up pass. Or perhaps some other way of passing contextual information downwards could be devised. It'd be great to have some thoughts and ideas on how best to do this!

### DIFF
--- a/src/dsl/static.jl
+++ b/src/dsl/static.jl
@@ -53,7 +53,7 @@ function parse_julia_expr!(stmts, bindings, name::Symbol, expr::Expr,
     input_vars = map((x) -> esc(x[1]), inputs)
     input_nodes = map((x) -> esc(x[2]), inputs)
     fn = Expr(:function, Expr(:tuple, input_vars...), esc(expr))
-    node = gensym()
+    node = gensym(expr.head)
     push!(stmts, :($(esc(node)) = add_julia_node!(
         builder, $fn, inputs=[$(input_nodes...)], name=$(QuoteNode(name)),
         typ=$(QuoteNode(typ)))))
@@ -72,7 +72,7 @@ end
 function parse_julia_expr!(stmts, bindings, name::Symbol, var::QuoteNode,
                            typ::Union{Symbol,Expr,QuoteNode})
     fn = Expr(:function, Expr(:tuple), var)
-    node = gensym()
+    node = gensym(var.value)
     push!(stmts, :($(esc(node)) = add_julia_node!(
         builder, $fn, inputs=[], name=$(QuoteNode(name)),
         typ=$(QuoteNode(typ)))))
@@ -82,7 +82,7 @@ end
 function parse_julia_expr!(stmts, bindings, name::Symbol, value,
                            typ::Union{Symbol,Expr,QuoteNode})
     fn = Expr(:function, Expr(:tuple), QuoteNode(value))
-    node = gensym()
+    node = gensym(repr(value))
     push!(stmts, :($(esc(node)) = add_julia_node!(
         builder, $fn, inputs=[], name=$(QuoteNode(name)),
         typ=$(QuoteNode(typ)))))
@@ -90,13 +90,9 @@ function parse_julia_expr!(stmts, bindings, name::Symbol, value,
 end
 
 function parse_julia_assignment!(stmts, bindings, line::Expr)
-    if line.head != :(=)
-        return false
-    end
-    @assert length(line.args) == 2
-    (lhs, expr) = line.args
+    if !MacroTools.@capture(line, lhs_ = rhs_) return false end
     (name::Symbol, typ) = parse_lhs(lhs)
-    node = parse_julia_expr!(stmts, bindings, name, expr, typ)
+    node = parse_julia_expr!(stmts, bindings, name, rhs, typ)
     bindings[name] = node
     true
 end
@@ -105,38 +101,37 @@ split_addr!(keys, addr_expr::QuoteNode) = push!(keys, addr_expr)
 split_addr!(keys, addr_expr::Symbol) = push!(keys, addr_expr)
 
 function split_addr!(keys, addr_expr::Expr)
-    @assert addr_expr.head == :call
-    @assert length(addr_expr.args) == 3
-    @assert addr_expr.args[1] == :(=>)
-    push!(keys, addr_expr.args[2])
-    split_addr!(keys, addr_expr.args[3])
+    @assert @capture(addr_expr, fst_ => snd_)
+    push!(keys, fst)
+    split_addr!(keys, snd)
 end
 
 choice_or_call_at(gen_fn::GenerativeFunction, addr_typ) = call_at(gen_fn, addr_typ)
 choice_or_call_at(dist::Distribution, addr_typ) = choice_at(dist, addr_typ)
 
+gen_arg_name(arg::Symbol) = gensym(arg)
+gen_arg_name(arg::QuoteNode) = gensym(arg.value)
+gen_arg_name(arg::Expr) = gensym(arg.head)
+gen_arg_name(arg::Any) = gensym(repr(arg))
+
        # parse_trace_expr!(stmts::Vector{Expr}, bindings, name::Symbol("##XYZ"), line::Epr(not :(=)), typ)
 function parse_trace_expr!(stmts, bindings, name, addr_expr, typ)
     # NOTE: typ is unused
-    if !(isa(addr_expr, Expr) && addr_expr.head == :macrocall
-        && length(addr_expr.args) >= 4 && addr_expr.args[1] == STATIC_DSL_TRACE)
-        return false
-    end
-    @assert isa(addr_expr.args[2], LineNumberNode)
-    call = addr_expr.args[3]
+    if !@capture(addr_expr, @macroname_(call_, site_)) return false end
+    if macroname != STATIC_DSL_TRACE return false end
     if !isa(call, Expr) || call.head != :call
         return false
     end
     local addr::Symbol
-    gen_fn_or_dist = gensym()
+    gen_fn_or_dist = gensym(call.args[1])
     push!(stmts, :($(esc(gen_fn_or_dist)) = $(esc(call.args[1]))))
-    if isa(addr_expr.args[4], QuoteNode) && isa(addr_expr.args[4].value, Symbol)
-        addr = addr_expr.args[4].value
+    if isa(site, QuoteNode) && isa(site.value, Symbol)
+        addr = site.value
         args = call.args[2:end]
     else
         # multi-part address syntactic sugar
         keys = []
-        split_addr!(keys, addr_expr.args[4])
+        split_addr!(keys, site)
         if !isa(keys[1], QuoteNode) || !isa(keys[1].value, Symbol)
             return false
         end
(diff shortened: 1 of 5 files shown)


src/dsl/static.jl
```diff
@@ -66,184 +66,172 @@ function parse_julia_expr!(stmts, bindings, name::Symbol, var::Symbol,
...
+            parse_assignment!(stmts, bindings, lhs_i, rhs_i)
+        end
+    else
+        # Handle single variable assignment (base case)
+        (name::Symbol, typ) = parse_typed_var(lhs)
+        # Generate new node name if name is already bound
+        node_name = haskey(bindings, name) ? gensym(name) : name
+        node = parse_julia_expr!(stmts, bindings, node_name, rhs, typ)
+        # Old bindings are overwritten with new nodes
+        bindings[name] = node
```
marcoct, line 185: @ztangent Is this the code that allows you to reassign a variable?

src/dsl/static.jl
```diff
@@ -66,184 +66,172 @@ function parse_julia_expr!(stmts, bindings, name::Symbol, var::Symbol,
...
+
+    # Add addr node
     push!(stmts, :($(esc(node)) = add_addr_node!(
-        builder, $(esc(gen_fn_or_dist)), inputs=[$(map(esc, inputs)...)], addr=$(QuoteNode(addr)),
-        name=$(QuoteNode(name)))))
-    true
+        builder, $(esc(gen_fn_or_dist)), inputs=[$(map(esc, inputs)...)],
+        addr=$(QuoteNode(addr)), name=$(QuoteNode(name)))))
+    # Return the name of the newly created node
+    return name
```
marcoct: @ztangent And this line, combined with the bottom-up traversal and replacement, allows you to nest `@trace` expressions inside Julia expressions?

src/dsl/static.jl
```diff
@@ -66,184 +66,172 @@ function parse_julia_expr!(stmts, bindings, name::Symbol, var::Symbol,
...
-    end
-    if isa(line.args[1], Expr) && line.args[1].head == :macrocall
-        var = gensym()
-        typ = QuoteNode(Any)
-        if !parse_trace_expr!(stmts, bindings, var, line.args[1], typ)
-            return false # the right-hand-side is a macro but not a valid `@trace` expression
-        end
-    elseif isa(line.args[1], Symbol)
-        var = line.args[1]
+function parse_expr!(stmts, bindings, expr)
```
marcoct: @ztangent It might be good to add a comment emphasizing that this function returns a new expression, as well as adds to stmts and bindings. Or maybe rename the function to `parse_and_collapse_expr!`.

src/dsl/static.jl
```diff
@@ -259,7 +247,7 @@ function make_static_gen_function(name, args, body, return_type, annotations)
         if arg.default != nothing
             error("Default argument values not supported in the static DSL.")
         end
-        node = gensym()
+        node = gensym(arg.name)
```
marcoct, line 274: Nice.

test/static_dsl.jl
```diff
@@ -97,7 +103,7 @@ params = ir.arg_nodes[2]
 
 # is_outlier
 is_outlier = ir.choice_nodes[1]
-@test is_outlier.name == :is_outlier
+# @test is_outlier.name == :is_outlier
```
marcoct: @ztangent The name is now a gensym right? If so, it's fine to delete these checks. We should probably clarify in the IR code what the purpose of the name actually is (for variable binding).

test/static_dsl.jl
```diff
@@ -232,7 +238,7 @@ end
 # at_choice_example_1
 ir = Gen.get_ir(typeof(at_choice_example_1))
 i = get_node_by_name(ir, :i)
-ret = get_node_by_name(ir, :ret)
+ret = get_node_by_addr(ir, :x)
```
marcoct, line 234: @ztangent Does the call to get_node_by_name raise an error here and in other similar lines in this file?

marcoct (approved):
@ztangent This is a major improvement!

I just had a couple clarification questions and one suggested function name change.

I think it makes the code easier to read and follow, as well as expands its functionality, so I think it can actually be merged soon, and the other questions about how to add loops can be moved back to the issue https://github.com/probcomp/Gen/issues/213, which we can keep open. 

It's nice that you identified the changes that you were able to make, and finish those, and start a discussion about future work -- that style will allow us to keep PRs small and not hanging around indefinitely.

Also, really nice job keeping up the standard of comprehensive unit tests.

src/dsl/static.jl
ztangent: Yup! What's happening here is that the first time the variable is assigned, a `JuliaNode` gets created with the same node name as the name of the variable. But when it gets reassigned, a new `JuliaNode` is created with a gensym-ed name, and the binding for that variable gets replaced by the newly created node, since all future references to that variable will refer to the new node* .

(*Assuming straight-line code, which is a restriction of the static DSL. If you do something weird like an assignment within a let block, the parser doesn't know that the assignment is within a different scope, so errors will occur.)

src/dsl/static.jl
ztangent: Yup, exactly, `return name` does that.

src/dsl/static.jl
ztangent: Makes sense! I'm happy with that, or maybe `process_expr!` to make clear that it's also modifying the expression in place? (@alex-lew did something similar with the `dist` DSL, and named the function `process_node`.)

test/static_dsl.jl
ztangent: Yup, it's a gensym now -- will delete these checks!

test/static_dsl.jl
ztangent: Yeah, because with the new changes, the node named `:ret` is now a `JuliaNode` which basically acts as a wrapper. To get the node corresponding to the `@trace` call (which has a gensym-ed name), you need to look it up by the trace address instead (or you can look it up by the gensym-ed name, but we can't know what that is in advance).

ztangent:
Awesome to hear!! :) And that makes sense about the additional extensions suggested in #213.

I'll make the suggested changes, and also update the docs to reflect the lessened restrictions on the SML.

One additional thing I was hoping to add as part of this pull-request is more stringent checks at parse-time, so that users don't accidentally exceed the limitations of the SML. For example, it wouldn't be hard to detect and disallow loops and let-blocks.

My only worry with this is that there are a lot of Julia constructs to potentially detect and disallow, and maybe it's not so great to inconsistently check for only some of them? The alternative is to rely on the user to read the docs carefully and adhere to the restrictions.

alex-lew:
This looks great!

I'll raise one concern, either for consideration now or something we can paste into an issue after merging.

Mostly, I'm curious how using these features might make Gen track dependencies more coarsely. 

For example: 

When a user writes:

```
@gen (static) function f()
  x, y = @trace(normal(0, 1), :x), @trace(normal(0, 1), :y)
  return x
```

does Gen know that the return value depends only on the random choice `:x`, not `:y`?

Or, similarly:

```
@gen (static) function f()
  pair = @trace(normal(0, 1), :x), @trace(normal(0, 1), :y)
  return pair[1]
```

Not all of these are possible to catch statically (and that's true in the current static DSL as well), but I wonder if we should either:

(1) in the docs, avoid encouraging users to write in a style that might inadvertently get them slower inference (due to less accurate dependency tracking) -- perhaps by explaining best practices;
(2) make an issue for this and outline ways we might better track dependencies in the static DSL?

For #2, I think ultimately some form of partial evaluation might be necessary. There's a section on using partial evaluation for "compiling probabilistic programs to graphical models" in the arXiv "Introduction to Probabilistic Programming" (https://arxiv.org/pdf/1809.10756.pdf, section 3.2.2) that we could look at.

ztangent:
@alex-lew you're right, I didn't think of that! The ability to write statements like
```julia
x, y = @trace(normal(0, 1), :x), @trace(normal(0, 1), :y)
```
actually comes from the intersection of the new support for tuple assignments, and the new support for nesting trace expressions with Julia expressions. So to explicitly prohibit it would look require adding this pretty ugly edge-case check in the parsing code that I don't think can be done elegantly...

My guess is that for now we should either avoid including such examples in the docs so that users don't use this pattern, or to explicitly warn them away from it. Because otherwise, I think there's a good case for both of those new features separately. Tuple assignment is useful because some generative functions return tuples, so it's nice to unpack them directly. And nesting @trace expressions in Julia expressions just makes code generally more concise.

ztangent:
Also, I just realized I unfortunately introduced a (mild) regression with this PR. On the master branch, list comprehensions without trace expressions work fine:
```julia
@gen (static) function foo(n)
       x = [i for i=1:n]
       return x
end
@assert foo(5) == [1, 2, 3, 4, 5]
```

But the bottom-up parsing introduced by this PR ends up parsing and rewriting the assignment like expression in the list comprehension, and also wrongly creates a Julia node for the assignment that it thinks occurs (though in the end the function ends up not being parsed as a whole). 

I think probably the best way to fix this is to preserve the original restriction that only top-level assignments will be -- it's quite rare that people write code with nested assignments anyway (maybe only in `begin ... end` blocks). Since expressions are already parsed line-by-line in `parse_static_dsl_function_body!`, it should be easy to rewrite the logic a little so that assignments are only pattern-matched at that level, rather than recursively matched using `postwalk`. Adding top-level line-parsing logic should also make it slightly easier in the future to support custom parsing of list comprehensions and loops. Will make this change along with the others suggested!

ztangent:
Made the changes I said I would, and updated the docs!

With the most recent changes, I think the execution model for the parser is both clearer and more extensible. The function body is parsed line by line with `parse_static_dsl_line!`.  Within `parse_static_dsl_line!`, we first do `postwalk` to parse and rewrite all nested `@trace` expressions in the line. Then we match the line / top-level expression to one of the allowed top-level constructs: params, assignments, or returns, and parse accordingly. If, in the future, we want to add support for other top-level constructs like for-loops, we can add it as a case within `parse_static_dsl_line!`.

I've also renamed functions to better reflect whether they are meant to parse only lines (e.g. `parse_param_line!`), or to parse expressions anywhere in the code. The parsing functions now have short docstrings, so that in the future everything should be more understandable to folks who want to make extensions.

src/dsl/static.jl
```diff
@@ -203,46 +204,69 @@ function parse_return!(stmts::Vector{Expr}, bindings, expr)
...
+        # Parse "lhs = rhs"
+        parse_assignment_line!(stmts, bindings, lhs, rhs)
+    elseif MacroTools.@capture(line, return expr_)
+        # Parse "return expr"
+        parse_return_line!(stmts, bindings, expr)
+    elseif line isa LineNumberNode
+        # Skip line number nodes
+    else
+        # Disallow all other top-level constructs
+        static_dsl_syntax_error(line, "Unsupported top-level construct.")
```
marcoct, line 244: @ztangent Great, this error message should save people some pain.

marcoct:
This looks good to me.

PR #216, opened 2020-03-26
Add involution MCMC with automated Jacobian calculation

marcoct:
This pull request adds a version of involution MCMC (i.e. fully programmable reversible jump MCMC) that automates computation of the Jacobian determinant correction, unlike the Gen's involution MCMC, which does not automate the computation of the Jacobian correction (automated computation of the Jacobian correction was a feature early on but was removed to reduce maintenance burden and system complexity).

Very briefly, involution MCMC requires the user to specify a family of proposal distributions (represented as a generative whose first argument is a trace of the model), and a (deterministic) function that is an involution from tuples (t, u) to tuples (t', u'), where t and t' are choice maps produced by the model generative function, and u and u' are choice maps produced by the proposal generative function. This involution is composed from: (i) an involution on the discrete choices e.g. `(t_{disc}, u_{disc})` to `(t'_{disc}, u'_{disc})`, and (ii) a family of bijections on the continuous choices, that is indexed by `(t_{disc}, u_{disc})`. It is this bijection whose Jacobian determinant is used to correct the MH acceptance probability.

Typically, the Jacobian is extremely sparse. This implementation automatically exploits this sparsity, to avoid filling in most entries in the Jacobian in most cases, and to reduce the complexity of the determinant calculation. Automatically computing the Jacobian determinant, and doing in an efficient way, is made possible by providing a DSL for users to specify the involution. The DSL makes reads and writes and copies involving the random choices in t, t', u, and u' explicit, so that the structure can be exploited. For example, copying the value of a random choice is equivalent to a read followed by a write, but using the copy language construct instead of a read followed by a write improves efficiency because an entire row / column of the Jacobian need not be computed.

Note that this feature is not yet documented. I will add documentation to this PR, hopefully soon, and before merging. I am also writing up math notes that describe how the implementation exploits sparsity structure to reduce the complexity of the Jacobian calculation and the Jacobian determinant calculation.

The PR includes four examples:

- examples/coal/coal.jl -- An adaptation of the example from Green's original reversible jump MCMC paper.

- examples/involution_mh_minimal_example.jl -- A minimal example.

- examples/gp_structure/involution_mh.jl -- Gaussian process structure learning using a flat address space for the random choices that determine the covariance function

- examples/gp_structure/involution_mh_tree.jl -- Gaussian process structure learning using a tree-structured address space for the random choices that determine the covariance function


docs/src/ref/mcmc.md
```diff
@@ -201,6 +201,207 @@ new_trace = permute_move(trace, 2)
...
+This is wasteful, since if we have inferred an accurate value for `:m`, we expect the values for `:m1` and `:m2` to be near this value.
+The same is true when proposing a structure change in the opposite direction.
+That means it will take many more steps to get an accurate estimate of the posterior probability distribution on the two structures.
+
+We would like to use inferred values for `:m1` and `:m2` to inform our proposal for the value of `:m`.
+For example, we could take the harmonic mean:
+```julia
+m = sqrt(m1 * m2)
+```
+However, there are many combinations of `m1` and `m2` that have the same harmonic mean.
```
bzinberg: This is the geometric mean.

docs/src/ref/mcmc.md
```diff
@@ -201,6 +201,207 @@ new_trace = permute_move(trace, 2)
...
+(y1, y2) = (1.0, 1.3)
+trace, = generate(model, (), choicemap((:y1, y1), (:y2, y2), (:z, false), (:m, 1.)))
+for iter=1:100
+    trace = split_merge_kernel(trace)
+    trace = fixed_structure_kernel(trace)
+end
+```
+We can then compare the results to the results from the Markov chain that used the selection-based structure-changing kernel:
+
+![rjmcmc plot](images/rjmcmc.png)
```
bzinberg, line 353: I think this image is not checked in?

bzinberg (commented):
I just read the documentation you added here in `docs/`.  This is awesome!!!

1. Adding the doc with the PR was super useful, as I'm much more efficiently able to do a high-level review.
2. Great ideas in the design of this involution DSL -- letting the user specify which parts of the trace count as continuous choices and which of those should be included in the Jacobian.
3. What is the difference under the hood between reading and copying?  Which of them (if any) corresponds to saying "treat these as constants with their current value when computing the Jacobian"?
4. I have thoughts about extending this framework to allow for functions `h` that are [covering maps](https://en.wikipedia.org/wiki/Covering_space) such that `h ∘ h` is not necessarily the identity but is a [deck transformation](https://en.wikipedia.org/wiki/Covering_space#Deck_(covering)_transformation_group,_regular_covers).  This could help with situations where we can't globally define a canonical branch of the inverse, for example mapping an angle `theta` to a point `(cos(theta), sin(theta))` on the unit circle where we have to branch-cut at some point and wrap around by 2pi.  Definitely future work though.

docs/src/ref/mcmc.md
marcoct: Fixed.

docs/src/ref/mcmc.md
marcoct: Fixed

docs/src/ref/mcmc.md
bzinberg: Great.  Small suggestions on presentation for the image (GitHub doesn't allow commenting on the binary file):
* On the bottom plots, vertical axis, change the ticks from 0 and 1 to F and T.
* On the bottom plots, horizontal axis, add a label like "# inference steps".

docs/src/ref/mcmc.md
marcoct: > 3. What is the difference under the hood between reading and copying?  Which of them (if any) corresponds to saying "treat these as constants with their current value when computing the Jacobian"?

A value can appear in the output pair of choice maps in one of three ways:
- explicit write
- explicit copy (copying values from input choice maps to output choice maps, potentially at different addresses)
- implicit copy (this can only occur for model address, and is done implicitly by 'update'; the addresses that are implicitly copied are those that are (i) not deleted from the model choice map, and (ii) not the target of an explicit write, and (iii) not the target of an explicit copy).

So, "treat these as constants with their current value, at their current address" is handled by the "implicit copy". But explicit copies also afford simplification of the Jacobian that needs to be computed, because like implicit copies, the explicit copies also produce columns in the Jacobian that contain all 0s except for a single 1 (which allows them to be ignored for the purposes of computing the determinant).

Under the hood, we first run the involution, and record each read, write, and explicit copy (address). Then, we determine the subset of these commands for which we actually need to compute their corresponding row or column in the Jacobian, and produce a version of the involution (f_array) that takes as input an array containing the values of just those special input addresses, and returns as output an array containing the values of just those special output addresses (the fact that the number of  such outputs and inputs is equal is not obvious, and a proof of this is given in the LaTex notes I sent).

> 4. I have thoughts about extending this framework to allow for functions `h` that are [covering maps](https://en.wikipedia.org/wiki/Covering_space) such that `h ∘ h` is not necessarily the identity but is a [deck transformation](https://en.wikipedia.org/wiki/Covering_space#Deck_(covering)_transformation_group,_regular_covers).  This could help with situations where we can't globally define a canonical branch of the inverse, for example mapping an angle `theta` to a point `(cos(theta), sin(theta))` on the unit circle where we have to branch-cut at some point and wrap around by 2pi.  Definitely future work though.

Interesting. On a related note, I haven't looked into applying this DSL this for quaternion-valued random choices yet. Let me know if you want to discuss that some time. Maybe your generalization is useful there.

src/inference/mh.jl
```diff
@@ -62,33 +62,34 @@ end
...
     (fwd_choices, fwd_score, fwd_ret) = propose(proposal, (trace, proposal_args...,))
     (new_trace, bwd_choices, weight) = involution(trace, fwd_choices, fwd_ret, proposal_args)
     (bwd_score, bwd_ret) = assess(proposal, (new_trace, proposal_args...), bwd_choices)
     check && check_observations(get_choices(new_trace), observations)
     if check
-        (trace_rt, fwd_choices_rt, weight_rt) = involution(new_trace, bwd_choices, bwd_ret, proposal_args)
+        (trace_rt, fwd_choices_rt, weight_rt) = involution(new_trace, bwd_choices, bwd_ret, proposal_args; check=check)
         if !isapprox(fwd_choices_rt, fwd_choices)
             println("fwd_choices:")
             println(fwd_choices)
```
bzinberg, line 97: I'd like to start moving from `println` to use of the logging macros, e.g. here:
```suggestion
            @error "fwd_choices: $fwd_choices"
```
and similarly below (but I'm not allowed to suggest below since it's not close to the changed lines).

src/inference/mh.jl
```diff
@@ -120,7 +121,7 @@ end
 """
     (new_trace, accepted) = mh(trace, selection::Selection; ..)
     (new_trace, accepted) = mh(trace, proposal::GenerativeFunction, proposal_args::Tuple; ..)
-    (new_trace, accepted) = mh(trace, proposal::GenerativeFunction, proposal_args::Tuple, involution::Function; ..)
```
bzinberg, line 123: What is `..` here?

src/inference/mh.jl
```diff
@@ -62,33 +62,34 @@ end
...
 
 The generative function `proposal` is executed on arguments `(trace, proposal_args...)`, producing a choice map `fwd_choices` and return value `fwd_ret`.
 For each value of model arguments (contained in `trace`) and `proposal_args`, the `involution` function applies an involution that maps the tuple `(get_choices(trace), fwd_choices)` to the tuple `(get_choices(new_trace), bwd_choices)`.
 Note that `fwd_ret` is a deterministic function of `fwd_choices` and `proposal_args`.
 When only discrete random choices are used, the `weight` must be equal to `get_score(new_trace) - get_score(trace)`.
-
-**Including Continuous Random Choices**
-When continuous random choices are used, the `weight` must include an additive term that is the determinant of the the Jacobian of the bijection on the continuous random choices that is obtained by currying the involution on the discrete random choices.
+When continuous random choices are used, the `weight` returned by the involution must include an additive term that is the determinant of the the Jacobian of the bijection on the continuous random choices that is obtained by currying the involution on the discrete random choices.
```
bzinberg: Perhaps worth explicitly noting here that the `@involution` DSL computes this weight automatically (it does, right?)

src/inference/involution_dsl.jl
```diff
@@ -0,0 +1,440 @@
...
+# TODO add more checks that we don't visit the same address twice (these can be
+# only enabled in 'check' mode)
+
+struct InvolutionDSLProgram
+    fn!::Function
+end
+
+struct FirstPassState
+    trace
+    u::ChoiceMap
```
bzinberg, line 18:
```suggestion
    """Choicemap of the proposal."""
    u::ChoiceMap
```

src/inference/involution_dsl.jl
```diff
@@ -0,0 +1,440 @@
...
+    constraints::ChoiceMap
+    u_back::ChoiceMap
```
bzinberg, line 24: add docstring for `u_back` -- is it the choicemap of the state that would be mapped to `u` by the involution?

src/inference/involution_dsl.jl
```diff
@@ -0,0 +1,440 @@
...
+    t_cont_reads::Dict
+    u_cont_reads::Dict
+    t_cont_writes::Dict
+    u_cont_writes::Dict
+    t_copy_reads::DynamicSelection
+    u_copy_reads::DynamicSelection
+    marked_as_retained::Set
```
bzinberg: brief docstring about what "retained" means (could be "see such-and-such place for more info") might help here

src/inference/involution_dsl.jl
```diff
@@ -0,0 +1,440 @@
...
+
+function FirstPassState(trace, u::ChoiceMap)
+    FirstPassState(
+        trace, u, choicemap(), choicemap(),
+        Dict(), Dict(), Dict(), Dict(),
+        DynamicSelection(), DynamicSelection(),
+        Set())
+end
+
+struct JacobianPassState{T}
```
bzinberg: is this `where T<:Real`?

src/inference/involution_dsl.jl
```diff
@@ -0,0 +1,440 @@
...
+
+    quote
+
+    # mutates the state
+    function $fn!($(esc(inv_state))::Union{FirstPassState,JacobianPassState}, $(map(esc, args)...))
+        $(esc(body))
+        nothing
+    end
+
+    $(esc(f)) = InvolutionDSLProgram($fn!)
```
bzinberg: Would it be easy to tweak this so that `f` picks up any docstring provided via
```julia
"""My docstring"""
@involution my_involution ...
```
?  (Asking because I think you did similarly before with the `@gen` macro)

src/inference/involution_dsl.jl
```diff
@@ -0,0 +1,440 @@
...
+
+# write_continuous_to_model
+
+macro write_continuous_to_model(addr, value)
+    quote write_continuous_to_model($(esc(inv_state)), $(esc(addr)), $(esc(value))) end
+end
+
+function write_continuous_to_model(state::FirstPassState, addr, value)
+    has_value(state.constraints, addr) && error("Model address $addr already written to")
+    state.constraints[addr] = value
```
bzinberg, line 194: My understanding is a bit shaky on this, but want to ask: if this `constraints` is the same `constraints` as in the API for `Gen.generate`, then will adding constraints here mess with the score that would be computed by `Gen.get_score`?

bzinberg (commented):
A few small comments.  Sending now, before I dive into the implementation of `apply_involution` (which I won't have time to do until later).

docs/tex/mcmc.tex
```diff
@@ -0,0 +1,208 @@
...
+In general, both $t$ and $u$ (and by necessity $t'$ and $u'$) may contain the values of continuous random choices, in which case the involution must be constructed from a family of differentiable bijections on the values of these continuous random choices.
+When applying a particular RJMCMC kernel, the determinant of the Jacobian of one of these bijections is required for the acceptance probability calculation.
+This document describes how to implement this Jacobian determinant computation efficiently using automatic differentiation, while exploiting the sparsity of the Jacobian.
+The approach uses a domain-specific language (DSL) for specifying families of bijections that makes the sparsity structure of the Jacobian explicit.
+
+\section{Example}
+
+We consider a variant of the changepoint model that was used to introduce RJMCMC in~\citep{green1995reversible}.
+Briefly, we put a prior on a class of piecewise constant functions on an interval $[0, x_{max}]$ where the number of segments is random and Poisson distributed.
```
bzinberg: num segments is Poisson + 1, num changepoints is Poisson?

docs/tex/mcmc.tex
```diff
@@ -0,0 +1,208 @@
...
+
+This section describes how to automatically perform the optimization recognized above in a general setting for arbitrary generative models and RJMCMC kernels, while using automatic differentiation.
+Let $t \in \mathbb{R}^K$ denote the vector of values of continuous random choices in the previous model trace.
+Let $u \in \mathbb{R}^L$ denote the vector of values of continuous random choices in the proposal trace.
+Similarly, let $t' \in \mathbb{R}^{K'}$ and $u' \in \mathbb{R}^{L'}$ denote the vector of values of continuous random choices in the new model trace, and new proposal trace.
+We assume that $K + L = K' + L'$ (this is the \emph{dimension matching} requirement), and that we are given a bijection $h : \mathbb{R}^{K + L} \to \mathbb{R}^{K + L}$ that is differentiable and whose Jacobian has nonzero determinant everwhere.\footnote{It may be possible to generalize and allow functions that are not differentiable at a measure zero set of points.}
+This section discusses how to efficiently compute the absolute value of the determinant of the Jacobian of $h$, which we denote by $J \in \mathbb{R}^{K + L \times K + L}$.
```
bzinberg:
```suggestion
This section discusses how to efficiently compute the absolute value of the determinant of the Jacobian of $h$, which we denote by $J \in \mathbb{R}^{(K + L) \times (K + L)}$.
```

docs/tex/mcmc.tex
```diff
@@ -0,0 +1,208 @@
...
+\left[
+\begin{array}{cc}
+J_{33}  &   J_{34}\\
+J_{43} &   J_{44}
+\end{array}
+\right]
+\]
+
+\begin{prop}
+The submatrix $[J_{11} \; J_{12}; J_{21} \; J_{22}]$ is square.
```
bzinberg, line 161: Doesn't this follow from the fact that each "copy" takes one element of `t_c ∪ u_c` to one element of `t'_c ∪ u'_c`, and this is a one-to-one correspondence?  In other words, each "output of a copy" corresponds to exactly one "input to a copy."  Or am I misunderstanding?

bzinberg (commented):
A couple of small comments on the Latex -- still going on the rest

docs/tex/mcmc.tex
marcoct: That's the intuition.

But syntactically, you are allowed to copy the same address to multiple destinations, so there the one-to-one correspondence is not an assumption. Instead, it derives from the assumption that the Jacobian determinant is nonzero (which is invoked in the proof of this proposition).

The argument could probably be simplified to:
(i) the assumption that the Jacobian determinant is nonzero implies that no two rows are the same
(ii) therefore the same address is not copied to multiple destination addresses, either explicitly or implicitly
(iii) then, t_c and u_c are elements that are copied (exactly once), and they are in one-to-one correspondence with elements of t_c' and u_c'
(iv) therefore, the submatrix is square

src/inference/involution_dsl.jl
```diff
@@ -0,0 +1,440 @@
...
+    end
+
+    u_key_to_index = Dict()
+    for (addr, v) in first_pass_state.u_cont_reads
+        if !(addr in first_pass_state.u_copy_reads) 
+            u_key_to_index[addr] = next_input_index
+            next_input_index += 1
+            push!(input_arr, v)
+        end
+    end
```
bzinberg: WDYT of this rephrasing to make it a bit more declarative?
```suggestion
    t_keys = setdiff(keys(first_pass_state.t_cont_reads),
                     keys(first_pass_state.t_copy_reads)) |> collect
    t_key_to_index = Dict(key => i for (i, key) in enumerate(t_keys))

    u_keys = setdiff(keys(first_pass_state.u_cont_reads),
                     keys(first_pass_state.u_copy_reads)) |> collect
    u_key_to_index = Dict(key => i for (i, key) in enumerate(u_keys))

    input_arr = vcat([first_pass_state.t_cont_reads[k] for k in t_keys],
                     [first_pass_state.u_cont_reads[k] for k in u_keys])
```

src/inference/involution_dsl.jl
```diff
@@ -0,0 +1,440 @@
...
+        next_output_index += 1
+    end
+
+    cont_u_back_key_to_index = Dict()
+    for (addr, v) in first_pass_state.u_cont_writes
+        cont_u_back_key_to_index[addr] = next_output_index
+        next_output_index += 1
+    end
+
+    function f_array(input_arr::AbstractArray{T}) where {T <: Real}
```
bzinberg, line 350: Could you add a docstring or comment for this function, explaining which inputs and outputs it has?  IIUC, it is a partial-application of the continuous part of the involution, with inputs corresponding to a particular superset of the columns of the reduced J.

src/inference/involution_dsl.jl
```diff
@@ -0,0 +1,440 @@
...
+
+    # apply the involution, and get metadata back about it, including the
+    # function f_array from arrays to arrays, that is the restriction of the
+    # involution to just a subset of the relevant continuous random choices
+    (constraints, u_back, marked_as_retained, t_key_to_index, input_arr, f_array) = apply_involution(
+        involution, trace, u, proposal_args, proposal_retval)
+
+    # update model trace
+    (new_trace, model_weight, _, discard) = update(
+        trace, get_args(trace), map((_) -> NoChange(), get_args(trace)), constraints)
```
bzinberg: Nit: perhaps finding a way to write this as a `fill` rather than `map` would be clearer.  E.g.,
```julia
fill(NoChange(), length(get_args(trace)))
```
or if you need tuples to stay tuples, could do
```julia
# in util file
fill_like(a::AbstractArray, val::T) where T = fill(val, size(a)...)
fill_like(t::Tuple, val::T) where T = Tuple(fill(val, length(t)))

# and then,
fill_like(get_args(trace), NoChange())
```

bzinberg (commented):
@marcoct, to help me review further, could you please explain a bit what the "first pass" and the "Jacobian pass" are, and how they differ?

docs/tex/mcmc.tex
bzinberg: Ah, got it.  Agreed, arguing directly that the correspondence is one-to-one makes things very clear.  Thanks!

src/inference/involution_dsl.jl
```diff
@@ -0,0 +1,440 @@
...
+            keep[index] = false
+        end
+    end
+    J = J[:,keep]
+    if size(J) != (num_outputs, num_outputs)
+        error("Jacobian was not square (size was $(size(J)); the function may not be an involution")
+    end
+
+    # log absolute value of Jacobian determinant
+    correction = LinearAlgebra.logabsdet(J)[1]
```
bzinberg, line 396:
```suggestion
    correction = LinearAlgebra.logabsdet(J)[1]
    if isinf(correction)
        @error "Weight correction is infinite; the function may not be an involution"
    end
```

src/inference/involution_dsl.jl
```diff
@@ -0,0 +1,440 @@
...
+# for the proposal, when calling another involution function with call
+
+# TODO proposal_retval is a way that continuous data can leak unaccounted for..
+# for now, the requirement is that the return value of f_disc cannot depend on
+# continuous addresses in the model or proposal although in the future, we
+# could do AD through its return value as well using choice_gradients() and add
+# these to the Jacobian...?
+
+# TODO add more checks that we don't visit the same address twice (these can be
+# only enabled in 'check' mode)
```
bzinberg: Could you file issues for these, and link to them from the comments?

bzinberg (commented):
Tested this out on my own use case (scene graphs), and it works!

@marcoct, please take a look at the above comments, and I expect there won't be too much more back-and-forth before I'm happy to merge this.

docs/src/ref/mcmc.md
```diff
@@ -201,6 +201,207 @@ new_trace = permute_move(trace, 2)
...
+
+An involution MCMC kernel in Gen takes as input a previous trace of the model (whose choice map we will denote by ``t``), and performs three phases to obtain a new trace of the model:
+
+- First, it traces the execution of a **proposal**, which is a generative function that takes the previous trace of the model as its first argument. Mathematically, we will denote the choice map associated with the trace of the proposal by ``u``. The proposal can of course be defined using the [Built-In Modeling Languages](@ref), just like the model itself. However, unlike many other uses of proposals in Gen, these proposals *can make random choices at addresses that the model does not*.
+
+- Next, it takes the tuple ``(t, u)`` and passes it into an **involution** (denoted mathematically by ``h``), which is a function that returns a new tuple ``(t', u')``, where ``t'`` is the choice map for a new proposed trace of the model, and ``u'`` are random choices for a new trace of the proposal. The defining property of the involution is that *it is invertible*, and *it is its own inverse*; i.e. ``(t, u) = h(h(t, u))``.
```
alex-lew: I wonder if it is worth adding some intuition here that `u'` is, conceptually, a description of a way that the proposal could be reversed, taking t' to t. But maybe the examples below are enough.

In general, I think that this is a good summary of what involution MH does, but am unsure what audience exactly will be prepared to make sense of it without some orientation (though again, maybe the example mitigates this). I wonder if there are similarly complex features in packages like SciPy or PyTorch that we could look at to get a sense of what their documentation looks like -- how to strike the balance between documenting and explaining/teaching?

docs/src/ref/mcmc.md
```diff
@@ -201,6 +201,207 @@ new_trace = permute_move(trace, 2)
...
+However, this algorithm will not be very efficient, because the internal proposal used by the selection variant of MH is not specialized to the model.
+In particular, when switching from the model with a single mean to the model with two means, the values of the new addresses `:m1` and `:m2` will be proposed from the prior distribution.
+This is wasteful, since if we have inferred an accurate value for `:m`, we expect the values for `:m1` and `:m2` to be near this value.
+The same is true when proposing a structure change in the opposite direction.
+That means it will take many more steps to get an accurate estimate of the posterior probability distribution on the two structures.
+
+We would like to use inferred values for `:m1` and `:m2` to inform our proposal for the value of `:m`.
+For example, we could take the geometric mean:
+```julia
+m = sqrt(m1 * m2)
```
alex-lew, line 276: Is there an inference or pedagogical reason to prefer this to the arithmetic mean here? (Just curious.)

docs/src/ref/mcmc.md
```diff
@@ -201,6 +201,207 @@ new_trace = permute_move(trace, 2)
...
+```
+However, there are many combinations of `m1` and `m2` that have the same geometric mean.
+In other words, the geometric mean is not *invertible*.
+However, if we return the additional degree of freedom alongside the geometric mean (`dof`), then we do have an invertible function:
```
alex-lew, line 280: I like this example, and don't think anything here needs to change for this PR. 

In a future revision, I wonder if this part could be more motivated by an inference concern (the way that line 273
nicely motivates the "merge" proposal). Readers may not understand why invertibility is required
here; connecting it directly to the "split" branch of the proposal (and the requirement that proposals
be reversible) might help.

docs/src/ref/mcmc.md
```diff
@@ -201,6 +201,207 @@ new_trace = permute_move(trace, 2)
...
+end
+```
+We can then compare the results to the results from the Markov chain that used the selection-based structure-changing kernel:
+
+![rjmcmc plot](images/rjmcmc.png)
+
+We see that if we initialize the Markov chains from the same state with a single mean (`:z` is `false`) then the selection-based kernel fails to accept any moves to the two-mean structure within 100 iterations, whereas the split-merge kernel transitions back and forth many times,
+If we repeated the selection-based kernel for enough iterations, it would eventually transition back and forth at the same rate as the split-merge.
+The split-merge kernel gives a much more efficient inference algorithm for estimating the posterior probability on the two structures.
+
```
alex-lew, line 358: Great minimal demo!

docs/src/ref/mcmc.md
```diff
@@ -201,6 +201,207 @@ new_trace = permute_move(trace, 2)
...
+### Involution DSL
+
+To define an involution using the involution DSL, use the [`@involution`](@ref) macro in front of a Julia function definition.
+The function must take three arguments, representing the arguments to the model, the arguments to the proposal (not including the trace), and the return value of the proposal.
```
alex-lew, line 362: Is it possible that the return value from the proposal might contain information about the proposal's choices, that could then be used to influence the involution without being properly tracked? If so,  we might want to warn users against using the proposal return value in certain ways.

docs/src/ref/mcmc.md
```diff
@@ -201,6 +201,207 @@ new_trace = permute_move(trace, 2)
...
+- `@copy_model_to_proposal(from_addr, to_addr)`: Copy the value (discrete or continuous) or an entire sub-map of choices under an address namespace from the input model choice map (``t``) to the output proposal choice map (``u'``).
+
+- `@copy_proposal_to_proposal(from_addr, to_addr)`: Copy the value (discrete or continuous) or an entire sub-map of choices under an address namespace from the input proposal choice map (``u``) to the output proposal choice map (``u'``).
+
+- `@copy_proposal_to_model(from_addr, to_addr)`: Copy the value (discrete or continuous) or an entire sub-map of choices under an address namespace from the input proposal choice map (``u``) to the output model choice map (``t'``).
+
+It is not necessary to explicitly copy values from the previous model choice map (``t``) to the new model choice map (``t'``) at the same address.
+These values will be copied automatically by the system.
+
+Finally, When reading a continuous value from the input model choice map that is expected to be implicitly copied by the system, the system will be more efficient if the following command is used instead of `@read_continuous_from_model`:
```
alex-lew: Lowercase `when` :-)

docs/src/ref/mcmc.md
```diff
@@ -201,6 +201,207 @@ new_trace = permute_move(trace, 2)
...
+- `@copy_model_to_model(from_addr, to_addr)`: Copy the value (discrete or continuous) or an entire sub-map of choices under an address namespace from the input model choice map (``t``) to the output model choice map (``t'``).
+
+- `@copy_model_to_proposal(from_addr, to_addr)`: Copy the value (discrete or continuous) or an entire sub-map of choices under an address namespace from the input model choice map (``t``) to the output proposal choice map (``u'``).
+
+- `@copy_proposal_to_proposal(from_addr, to_addr)`: Copy the value (discrete or continuous) or an entire sub-map of choices under an address namespace from the input proposal choice map (``u``) to the output proposal choice map (``u'``).
+
+- `@copy_proposal_to_model(from_addr, to_addr)`: Copy the value (discrete or continuous) or an entire sub-map of choices under an address namespace from the input proposal choice map (``u``) to the output model choice map (``t'``).
+
+It is not necessary to explicitly copy values from the previous model choice map (``t``) to the new model choice map (``t'``) at the same address.
+These values will be copied automatically by the system.
```
alex-lew, line 396: Perhaps we could make more explicit what triggers this implicit copying? "If, using the proposed constraints, the model visits an address that was not explicitly copied or written to, the old value will automatically be copied by the system."

docs/tex/mcmc.tex
alex-lew: 👍 At some point, it could be good to include something to the effect of, "If you find yourself copying the same source address to multiple locations, it probably means your involution is not valid. (+ some discussion of why.)"

docs/tex/references.bib
```diff
@@ -0,0 +1,39 @@
...
+@inproceedings{cusumano2019gen,
+  title={Gen: a general-purpose probabilistic programming system with programmable inference},
+  author={Cusumano-Towner, Marco F and Saad, Feras A and Lew, Alexander K and Mansinghka, Vikash K},
+  booktitle={Proceedings of the 40th ACM SIGPLAN Conference on Programming Language Design and Implementation},
+  pages={221--236},
+  year={2019},
+  organization={ACM}
+}
+
+
```
alex-lew, line 39: [Depending on the purpose / ultimate audience of this TeX document, it could be good, eventually, to cite the "Reversible Jump Probabilistic Programming" paper and describe how this is different: much more general, but also  the involution is not automatically constructed.]

examples/gp_structure/involution_mh.jl
```diff
@@ -0,0 +1,217 @@
...
+        @copy_model_to_proposal(:tree => (cur, :param), :subtree => (cur, :param))
+    elseif node_type == SQUARED_EXP
+        @copy_model_to_proposal(:tree => (cur, :length_scale), :subtree => (cur, :length_scale))
+    elseif node_type == PERIODIC
+        @copy_model_to_proposal(:tree => (cur, :scale), :subtree => (cur, :scale))
+        @copy_model_to_proposal(:tree => (cur, :period), :subtree => (cur, :period))
+    elseif (node_type == PLUS) || (node_type == TIMES)
+        child1 = get_child(cur, 1, max_branch)
+        child2 = get_child(cur, 2, max_branch)
+        @call(walk_previous_subtree(child1)) # use same namespace
```
alex-lew: I believe `@call` is undocumented? But great that it exists!

Do you think there's ever a need to make a call to an involution and "focus" a subtrace (of the model and/or proposal)? 

I'm thinking of how you would write an involution proposal that calls another involution proposal to handle a particular call in the model.

examples/gp_structure/involution_mh.jl
```diff
@@ -0,0 +1,217 @@
+using Gen
+
+include("../gp_structure/shared.jl")
+
+@gen function covariance_prior(cur::Int)
```
alex-lew, line 5: (The version we used for the class actually makes recursive calls at the addresses `:left` and `:right`, which makes the involution a bit easier -- we could port the code here some point.)

src/inference/involution_dsl.jl
```diff
@@ -0,0 +1,440 @@
+import ForwardDiff
+import MacroTools
+import LinearAlgebra
+
+# TODO allow providing a pair of address namespaces, one for the model and one
+# for the proposal, when calling another involution function with call
```
alex-lew: Ah, I see this already part of the plan :-)

(We could still document `@call`, even before this feature is added.)

src/inference/involution_dsl.jl
```diff
@@ -0,0 +1,440 @@
...
+
+# TODO proposal_retval is a way that continuous data can leak unaccounted for..
+# for now, the requirement is that the return value of f_disc cannot depend on
+# continuous addresses in the model or proposal although in the future, we
+# could do AD through its return value as well using choice_gradients() and add
+# these to the Jacobian...?
```
alex-lew: Nice -- this resolves one of my comments above. (Assuming `f_disc` means the return value of the proposal?). I do think we should document this requirement, in the docs, not just in an issue.

src/inference/involution_dsl.jl
```diff
@@ -0,0 +1,440 @@
...
+
+# TODO add more checks that we don't visit the same address twice (these can be
```
alex-lew: What does "visit" mean here? 

Is it invalid to read twice? copy and read? Or do you mean write to the same address twice?

src/inference/involution_dsl.jl
```diff
@@ -0,0 +1,440 @@
...
+
+"""
+    @involution function f(...)
+        ..
+    end
+
+Write a program in the [Involution DSL](@ref).
+"""
+macro involution(ex)
+    MacroTools.@capture(ex, function f_(args__) body_ end) || error("expected syntax: function f(..) .. end")
```
alex-lew, line 63: Not too important -- but I think `MacroTools` has plug-and-play utilities for matching function definitions no matter how they are phrased (Xuan recently changed `@gen` to use them, and the `@dist` DSL does too). Docs here: http://mikeinnes.github.io/MacroTools.jl/stable/utilities/#Function-definitions-1

It might be good to make the one-line change here, so that people could write `@involution f(...) = begin ... end` too (if that happens to be their style).

src/inference/involution_dsl.jl
```diff
@@ -0,0 +1,440 @@
...
+    quote read_continuous_from_model_retained($(esc(inv_state)), $(esc(addr))) end
+end
+
+function read_continuous_from_model_retained(state::FirstPassState, addr)
+    push!(state.marked_as_retained, addr)
+    state.trace[addr]
+end
+
+function read_continuous_from_model_retained(state::JacobianPassState, addr)
+    state.trace[addr] # read directly from the trace, instead of the array
```
alex-lew, line 182: Does the value exist in the array?

src/inference/involution_dsl.jl
alex-lew: @bzinberg I think this `constraints` is just `t'` from the TeX doc -- it's where you're keeping track of the things that have been proposed to the model. It will ultimately be passed to `update`.  (I think it shouldn't contain any of the addresses that were originally passed to `generate` when creating an initial MCMC trace, because your proposal shouldn't be changing any of those values.)

src/inference/involution_dsl.jl
```diff
@@ -0,0 +1,440 @@
...
+    involution.fn!(first_pass_state, get_args(trace), proposal_args, proposal_retval)
+
+    # create input array and mappings input addresses that are needed for Jacobian
+    # exclude addresses that were moved to another address
+    input_arr = Vector{Float64}()
+    next_input_index = 1
+
+    t_key_to_index = Dict()
+    for (addr, v) in first_pass_state.t_cont_reads
+        if !(addr in first_pass_state.t_copy_reads)
```
alex-lew, line 323: This seems to resolve my question above about whether I'm allowed to read and copy (yes).

src/inference/involution_dsl.jl
```diff
@@ -0,0 +1,440 @@
...
+end
+
+# apply 
+
+function apply_involution(involution::InvolutionDSLProgram, trace, u, proposal_args, proposal_retval)
+
+    # take the first pass through the involution, mutating first_pass_state
+    first_pass_state = FirstPassState(trace, u)
+    involution.fn!(first_pass_state, get_args(trace), proposal_args, proposal_retval)
+
```
alex-lew, line 299: @marcoct If we call `update` here, rather than after `apply_involution` returns, can't we use the `discard` to automatically determine which continuous reads were `retained`, and construct `f_array` accordingly? That way, we wouldn't need to expose any special `retained` functionality to the user. Or maybe I'm missing something?

src/inference/involution_dsl.jl
```diff
@@ -0,0 +1,440 @@
...
+    # remove columns for inputs from the trace that were retained
+    # NOTE: these columns did not have to be computed in the first place, if
+    # the user had marked the associated random choices as retained
+    keep = fill(true, length(input_arr))
+    for (addr, index) in t_key_to_index
+        if !has_value(discard, addr)
+            keep[index] = false
+        end
+    end
+    J = J[:,keep]
```
alex-lew: I think I see what's going on here. 

So: if the user writes `read` but doesn't make the corresponding `write` for a continuous choice they want "implicitly copied," that's fine -- we can just detect and remove the relevant rows from the Jacobian before computing the determinant. 

But see above re: whether we might call `update` from within `apply_involution`, and remove the need for the user to do this manually (though again, I could totally be missing something!).

alex-lew (commented):
@marcoct This is fantastic!! Awesome work.

I've left a number of comments throughout the review, and am happy to merge. My two (edit: three) most substantive comments are:

1. I think that user-marked `retained` reads may not be necessary -- if `update` is called between the first pass and the Jacobian pass, I think we can just read the `discard` to see what was (not) retained? But perhaps there's something I'm missing here.

2. We should add a bit of documentation for `@call`.

3. (**Edit to add:**) How does this play with vector- or matrix-shaped random choices?

Thanks for this awesome feature!!

docs/tex/mcmc.tex
bzinberg: s/probably/definitely

src/inference/involution_dsl.jl
bzinberg: Oh good.  I think when I originally read that name I thought `Gen.generate` would be used to construct the new trace, and in that context these things would become "constraints" or something like that... was going to suggest changing the name.

Actually... I think I'd still like to suggest that :)  Would a name like `new_choices` make sense?

docs/src/ref/mcmc.md
bzinberg: [yes](https://github.com/probcomp/Gen/pull/216#discussion_r402120720)

docs/src/ref/mcmc.md
bzinberg: @alex-lew For typos, you can make a Suggestion in your comment and then apply that Suggestion by point-and-click.  About to verify that what I said is true by doing so.
```suggestion
Finally, when reading a continuous value from the input model choice map that is expected to be implicitly copied by the system, the system will be more efficient if the following command is used instead of `@read_continuous_from_model`:
```

src/inference/involution_dsl.jl
bzinberg: +1 to this.  I think a lot of users (including myself) sometimes see this error when we make mistakes in our code:
https://github.com/probcomp/Gen/blob/ef06cf4745536aa0f8a8ddabd086b8f8aa28f18c/src/dynamic/update.jl#L192
is that the same notion of "visit"?

src/inference/involution_dsl.jl
bzinberg: +1 to using the `MacroTools` util, but @alex-lew I gotta say, if you come up with a case where someone can reasonably define their involution in one line, let me know.

src/inference/involution_dsl.jl
bzinberg: Is there some way we could just not expose the retval at all?  And then we can change that once we are able to handle the retval differentiably.

I think if feasible, making it impossible to do the wrong thing is preferable to saying in a doc "don't do this."

src/inference/involution_dsl.jl
alex-lew: :+1: for `new_choices` :-)

alex-lew:
@marcoct Would a check for whether an argument is a Float/Array of Floats be enough to determine which choices are continuous? Or do you think that would fail in some cases? I am wondering if `@copy`, `@read`, `@write` would be enough.

src/inference/involution_dsl.jl
```diff
@@ -0,0 +1,441 @@
...
+
+function apply_involution(involution::InvolutionDSLProgram, trace, u, proposal_args, proposal_retval)
+
+    # take the first pass through the involution, mutating first_pass_state
+    first_pass_state = FirstPassState(trace, u)
+    involution.fn!(first_pass_state, get_args(trace), proposal_args, proposal_retval)
+
+    # create input array and mappings input addresses that are needed for Jacobian
+    # exclude addresses that were moved to another address
+    input_arr = Vector{Float64}()
```
alex-lew, line 306: @marcoct What's your thought on multivariate choices (e.g., sample a multivariate normal and multiply it by 2)? 

I'm also remembering a meeting with Luke and Maddie where we were discussing multivariate normal proposals that would only sample some of the dimensions, where the involution replaces a sub-vector of the model's original choice. I guess we could handle that by treating it as a discrete choice in this DSL?

alex-lew:
A convenience feature that might be useful to have eventually: checking whether an address exists in the proposal / model trace. You should always be able to figure out the same information by doing some computation on the choices that *must* exist, so this isn't urgent.

docs/src/ref/mcmc.md
marcoct: Done

src/inference/mh.jl
marcoct: @bzinberg This code was already there. Let's create a separate issue for this.

src/inference/mh.jl
marcoct: The `..` stands for the keyword arguments, which are:
`check=false, observations=EmptyChoiceMap()`

src/inference/mh.jl
marcoct: Done

src/inference/involution_dsl.jl
marcoct: Most of these fields are hard to describe succinctly. I added a reference to the LaTex document.

src/inference/involution_dsl.jl
marcoct: Done

src/inference/involution_dsl.jl
marcoct: Done

src/inference/involution_dsl.jl
marcoct: Constraints contains a subset of the choices in t' from the Tex doc. It does not contain those elements of t' that are going to be copied implicitly from t.

docs/tex/mcmc.tex
marcoct: Done

src/inference/involution_dsl.jl
marcoct: Done

src/inference/involution_dsl.jl
marcoct: Thanks for the suggestion, this `map(..)` code appears everywhere in Gen, we can fix it more uniformly in a separate code cleanup pass. Most uses set the tuple of all NoChange() values. I have actually been thinking it makes more sense to make argdiff an optional keyword argument, that defaults to being the tuple of NoChanges(). New Issue: https://github.com/probcomp/Gen/issues/229

marcoct:
> What's your thought on multivariate choices (e.g., sample a multivariate normal and multiply it by 2)?

@alex-lew I think that should be supported. I made a separate issue for it: https://github.com/probcomp/Gen/issues/230

src/inference/mh.jl
bzinberg: No need -- this change was included in https://github.com/probcomp/Gen/commit/5d3938572fe5f0e9ab468729a394c237695f2bfd and I don't think there are any more instances of it.

src/inference/involution_dsl.jl
marcoct: @alex-lew That's interesting. I will look more closely into that.

docs/src/ref/mcmc.md
marcoct: @alex-lew Not sure. I took this from Green's model and did not think about why it is done this way.

src/inference/involution_dsl.jl
alex-lew: @marcoct Maybe here is where the `retained` calculation could be done automatically, if `update` has already been called? If the address does not appear in the set of `t_cont_writes`, and is also not in the discard, then it must have been implicitly copied (and therefore we can skip adding it to the array).

src/inference/involution_dsl.jl
marcoct: I meant it is invalid to write/copy to the same address twice. But it is also invalid to copy from the same address twice, and we could add an optional check for that too. I'll expand upon this comment.

src/inference/involution_dsl.jl
marcoct: @alex-lew Yeah, you're right. We can just run `update` first. I made this change. Nice find!

src/inference/involution_dsl.jl
marcoct: This code is now removed

marcoct:
> @marcoct Would a check for whether an argument is a Float/Array of Floats be enough to determine which choices are continuous? Or do you think that would fail in some cases? I am wondering if `@copy`, `@read`, `@write` would be enough.

@alex-lew We could do that for read and write (there is already no distinction for copy), but I'm hesitant to tie the definition of discrete/continuous to particular Julia types, because it (i) seems like it could be brittle -- what if we want to expand support, and (ii) the choice may not be forced and obvious to users e.g. some users might assume that 'Real' is the type that we should use for this (which it can't be because a Bool is a Real in Julia), (iii) it makes the implementation a little more complex. I guess I also like drawing the user's attention to the distinction between the two types of random choices, since they play distinct roles in the definition of the involution.

I could also imagine wanting to support continuous random choices that are user-defined composite types (like quaternions?) and not floats or arrays of floats. (Mathematically, we haven't fleshed that out yet, but it seems that some provisioning for extensibility would be good).

But we should discuss this more.  I added it to this issue: https://github.com/probcomp/Gen/issues/230

marcoct:
@alex-lew Re automaticaliy detecting discrete / continuous, I don't want to couple this discrete / continuous distinction too tightly with specific Julia types. I would be more in favor of assembling a more independent way for choice maps to indicate whether a value is discrete or continuous. This is a general theme in Gen's architecture -- e.g. introducing our own static IR instead of using Julia's IR is another instance of this type of decision.

marcoct:
> A convenience feature that might be useful to have eventually: checking whether an address exists in the proposal / model trace. You should always be able to figure out the same information by doing some computation on the choices that _must_ exist, so this isn't urgent.

@alex-lew That's very interesting. I like it. I'll add it to an issue.

src/inference/involution_dsl.jl
marcoct: I am aggregating all issues generated by this PR here: https://github.com/probcomp/Gen/issues/230

src/inference/involution_dsl.jl
marcoct: Done

docs/src/ref/mcmc.md
marcoct: > @marcoct, to help me review further, could you please explain a bit what the "first pass" and the "Jacobian pass" are, and how they differ?

@bzinberg The first pass actually applies the involution (and now produces the new trace as well), but it doesn't compute the Jacobian correction. The second pass computes the reduced Jacobian matrix. We have to run the second pass separately because (i) we don't know what addresses actually need to be included in the reduced Jacobian until after we have run the involution, and (ii) because the AD system being used (ForwardDiff) wants to compute Jacobians for functions that map arrays to arrays.

docs/tex/mcmc.tex
marcoct: Done

src/inference/involution_dsl.jl
marcoct: Done

docs/src/ref/mcmc.md
marcoct: I added the intuition about `u'`that you mentioned.

examples/gp_structure/involution_mh.jl
marcoct: @alex-lew Yes, I had that as a future feature in the todo notes (it was actually in the old version of this feature), and now it is documented here: https://github.com/probcomp/Gen/issues/230

docs/src/ref/mcmc.md
marcoct: Done

examples/gp_structure/involution_mh.jl
marcoct: @alex-lew this is what `examples/gp_structure/involution_mh_tree.jl` does.

src/inference/involution_dsl.jl
marcoct: @bzinberg The proposal retval is convenient because sometimes the proposal and the involution are walking the same data structures, and there can be reduced code duplication, (and at least, reduced unecessary code execution), if the involution can rely on some precomputation done during the proposal. Two examples of this (where the precomputation has to do with the discrete structure of the trace) are here: https://github.com/probcomp/Gen/blob/20200322-marcoct-rjmcmcjacobians/examples/gp_structure/involution_mh.jl#L137 and here: https://github.com/probcomp/Gen/blob/20200322-marcoct-rjmcmcjacobians/examples/gp_structure/involution_mh_tree.jl#L103

But I agree that it is potentially confusing. Also, I think that automating either the detection of an invalid proposal retval or its use by the involution, or automating its contribution to the Jacobian, might be possible with Zygote, but is pretty involved.

I might be okay with not allowing dependence on proposal retval, and paying the cost of some reversible jump algorithims being more verbose and/or inefficient.

@alex-lew What do you think?

docs/tex/references.bib
marcoct: I added this to the issue here: https://github.com/probcomp/Gen/issues/230

marcoct:
@alex-lew @bzinberg Thanks for the great and detailed code reviews! I really appreciate it. I believe the only unresolved item is what to do about the proposal return value.

marcoct:
I thought more about it and it's important that we can read from the proposal return value in order to get implementations that have the right asymptotic scaling in some cases: In general we want to be able to rely on using precomputed values that are stored in the model trace (e.g. in the return value or other "auxiliary state"). If we don't have access to these values, then we will have to recompute them from scratch in the involution. Currently the proposal retval is the vehicle that allows this information to be accessible (because the proposal can read from the trace arbitrarily).

Another option would be to allow reads from the model auxiliary state directly. But that has the same challenges as reading from the proposal auxiliary state, without the added benefit of reusing computation done during the proposal.

marcoct:
I think that in this case (and Gen overall) it is better to err on the side of leaving "escape hatches" that let the user do things that the system is unable to track or account for, instead of making the system very restrictive. I'll add a clear warning that you should only rely on the proposal_retval in the involution if you know what you are doing (but with a more precise characterization of what that means).

docs/tex/references.bib
bzinberg: Thanks Marco for recording these future improvements in the issue tracker :slightly_smiling_face:

src/inference/involution_dsl.jl
```diff
@@ -0,0 +1,411 @@
+import ForwardDiff
+import MacroTools
+import LinearAlgebra
+
+struct InvolutionDSLProgram
+    fn!::Function
+end
+
+# See this math writeup for an understanding of how this code works:
+# docs/tex/mcmc.pdf
```
bzinberg, line 10: :+1:

bzinberg:
Looks great!  Thanks @marcoct for your detailed responses to the code review, and thanks @alex-lew for your thorough review as well.

> I think that in this case (and Gen overall) it is better to err on the side of leaving "escape hatches" that let the user do things that the system is unable to track or account for, instead of making the system very restrictive. I'll add a clear warning that you should only rely on the proposal_retval in the involution if you know what you are doing (but with a more precise characterization of what that means).

I think that's right.  Given your reasoning about why we need the ability to reuse computation from the proposal, I'm good with adding a "don't use this unless you know what you're doing" warning as suggested in https://github.com/probcomp/Gen/pull/216#issuecomment-608485573.  And then I think we're ready to merge!

bzinberg (approved):

docs/src/ref/mcmc.md
```diff
@@ -414,6 +414,12 @@ end
 Note that when constructing involutions that call other `@involution` functions, the function being called (`bar` in this case) need not be mathematically speaking, an involution itself, for `foo` to be mathematically be an involution.
 Also, the top-level function must take three arguments (`model_args`, `proposal_args`, and `proposal_retval`), but any other `@involution` function may have an argument signature of the user's choosing.
 
+Some additional tips for defining valid involutions:
+
+- If you find yourself copying the same source address to multiple locations, it probably means your involution is not valid (the Jacobian matrix will have rows that are identical, and so the Jacobian determinant will be nonzero).
```
alex-lew: Does this apply only to continuous choices? (The justification seems to.)

alex-lew (approved):
This all looks great! I'm ready to approve, with a couple minor notes:

* Do we want to implement the doc changes ("don't use proposal retval unless...") before merging?

* Very minor, but I wonder if `@involution` is the right syntax, for two reasons: (a) involution functions that aren't "top-level" are not involutions in any sense, (b) even the top-level function is now an involution from implicit arguments to return value, not the actual things that are listed as arguments,  and (b) "involution" is a somewhat technical term. Maybe there is a way to emphasize that these "involutions" are essentially (the deterministic part) of proposal implementations? But I'm not sure what word would be better than `@involution`, so maybe that's something we can revisit later.

docs/src/ref/mcmc.md
bzinberg: True there, but even in the discrete case, the function cannot possible be an involution if we copy a discrete latent var to two addresses.  (Not sure whether we want to strictly enforce or let the user proceed at their own risk)

alex-lew: @bzinberg Right, unless the second address is guaranteed to be the same as the first address, given the rest of the trace (which is possible in the case of a discrete choice).

marcoct: Yeah, you can have a valid involution that copies the discrete addresses that way. I'll specify that this tip only applies to continuous choices.

marcoct:
>   Do we want to implement the doc changes ("don't use proposal retval unless...") before merging?

Done

>  Very minor, but I wonder if `@involution` is the right syntax, for two reasons: (a) involution functions that aren't "top-level" are not involutions in any sense, (b) even the top-level function is now an involution from implicit arguments to return value, not the actual things that are listed as arguments,  and (b) "involution" is a somewhat technical term. Maybe there is a way to emphasize that these "involutions" are essentially (the deterministic part) of proposal implementations? But I'm not sure what word would be better than `@involution`, so maybe that's something we can revisit later.

@alex-lew  Yeah, I agree about the issue. I'm not sure what the right word is. Since this feature, or something similar to it (without update), can be used in settings where it's not an involution (and where the previous model is different than the new model), perhaps the right name will emerge in that more general setting. I want to say `@bijection` could be better name for some more generalized version (it captures the essential mathematical feature that enables computing pushforward density). My only reservation about that is that there is additional structure we are assuming here -- the fact that the input and output are tuples of choice maps (t, u) instead of just flat choice maps. If we generalize the DSL to read from a single choice map and write to a single choice map (that distinguishes between two address namespaces e.g. `:model` and `:proposal` --- this is how I implemented this feature originally actually) then I think bijection could be an even more appropriate name. But, the role of update in the involution version complicates the story a bit.

Since we don't have a solution right now, I'll merge it as is and we can revisit as we understand the braoder design space better.

PR #217, opened 2020-03-27
Slight fix in noise-tracking for Map and Unfold

alex-lew:
In both combinators, the code currently reads:

```
state.noise += (project(subtrace, EmptySelection()) - project(subtrace, EmptySelection()))
```

I believe we should be subtracting the noise for the _previous_ subtrace:

```
state.noise += (project(subtrace, EmptySelection()) - project(prev_subtrace, EmptySelection()))
```


marcoct:
@alex-lew I believe so to :)

PR #221, opened 2020-03-28
Fix eval bug with precompilation of @dist distributions in other packages

### LINKED ISSUE #220: Error precompiling when using `@dist` distribution in a module that uses Gen
I think I may have introduced this bug due to using eval in the `@dist` macro?? :( Will try to fix and do a PR accordingly. 

```
WARNING: eval into closed module Gen:
Expr(:->, Expr(:tuple, :f, Symbol("##422")), Expr(:block, #= Symbol("C:\Users\...\.julia\packages\Gen\UZP6j\src\modeling_library\dist\dist_dsl.jl"):134 =#, Expr(:call, :f, Expr(:quote, 1), Symbol("##422"))))
```

For reference, this was the function that caused the error
```julia
"Boolean corruption noise."
@dist bitflip(bit, prob) = bernoulli((1-bit)*prob + bit*(1-prob))
```

Strangely this compiles in the REPL, and causes errors despite not having types... Which is extra weird because I also defined a `labeled_cat` using `@dist`, and that doesn't cause issues with pre-compilation.

ztangent:
Fixes #220 by passing `__module__` in the `@dist` macro to invocations of `dist_call`, thereby ensuring the evaluations of the following line occur in the calling module, instead of `eval`-ing in `Gen`:

```julia
arg_passer = __module__.eval(:((f, $(new_f_arg_names...)) -> f($(actual_arg_exprs...))))
```

### DIFF
--- a/src/modeling_library/dist/dist_dsl.jl
+++ b/src/modeling_library/dist/dist_dsl.jl
@@ -109,11 +109,11 @@ end
 # ACTUAL DSL
 
 # A function call within a @dist body
-function dist_call(d::Distribution{T}, args) where T
+function dist_call(d::Distribution{T}, args, __module__) where T
     DistWithArgs{T}(d, args)
 end
 
-function dist_call(f, args)
+function dist_call(f, args, __module__)
     if any(x -> (typeof(x) <: DistWithArgs), args)
         f(args...)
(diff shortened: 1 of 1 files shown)


alex-lew:
Thanks for catching this, @ztangent! Looks good to me.

Issue #185, opened 2020-02-03
Handle macro calls in static DSL


georgematheos:
I made a PR that implements a simple version of this: #184, but realize that it leads to some inconsistent macroexpansion behavior between the static and dynamic DSL.  I include thoughts on how to implement this properly in the PR.

ztangent:
Finally addressed by #280! :D

PR #223, opened 2020-03-31
Fix the call to `@trace` in tilde desugaring

### LINKED ISSUE #222: Tilde syntax `x ~ foo` looks for `@trace` in the caller's namespace
Example:
```julia
Gen.@gen function f()
  x ~ Gen.normal(0, 1)
end
# ERROR: LoadError: UndefVarError: @trace not defined
```
I think this is a bug in the macro escaping.

bzinberg:
Fixes https://github.com/probcomp/Gen/issues/222.

Quoting https://github.com/probcomp/Gen/issues/222#issuecomment-606780545:

> I think what should be happening is that `trace` (as a function on ASTs) should be called at macro expansion time, from within Gen's internals. Instead, the desugarer was punting the `@trace`ing to the caller, which actually worked if the caller happened to have `@trace` in their namespace, but was not the intended logic.

## Tested

The minimal example in https://github.com/probcomp/Gen/issues/222#issue-591297488 now works.

### DIFF
--- a/src/dsl/dsl.jl
+++ b/src/dsl/dsl.jl
@@ -72,13 +72,17 @@ end
 
 function desugar_tildes(expr)
     MacroTools.postwalk(expr) do e
+        # Expand the `@trace` macro as defined in this module (even if the caller
+        # doesn't have an analogous macro in their module), and leave the
+        # remaining macros to be expanded in the caller's scope.
         if MacroTools.@capture(e, {*} ~ rhs_)
-            :(trace($rhs))
+            macroexpand(@__MODULE__, :(@trace($rhs)), recursive=false)
         elseif MacroTools.@capture(e, {addr_} ~ rhs_)
-            :(trace($rhs, $(addr)))
+            macroexpand(@__MODULE__, :(@trace($rhs, $(addr))), recursive=false)
         elseif MacroTools.@capture(e, lhs_ ~ rhs_)
             addr_expr = address_from_expression(lhs)
-            :($lhs = trace($rhs, $(addr_expr)))
+            macroexpand(@__MODULE__, :($lhs = @trace($rhs, $(addr_expr))),
+                        recursive=false)
         else
             e
         end
--- a/src/dsl/dsl.jl
+++ b/src/dsl/dsl.jl
@@ -73,12 +73,12 @@ end
 function desugar_tildes(expr)
     MacroTools.postwalk(expr) do e
         if MacroTools.@capture(e, {*} ~ rhs_)
-            :(@trace($rhs))
+            :(trace($rhs))
         elseif MacroTools.@capture(e, {addr_} ~ rhs_)
-            :(@trace($rhs, $(addr)))
+            :(trace($rhs, $(addr)))
         elseif MacroTools.@capture(e, lhs_ ~ rhs_)
             addr_expr = address_from_expression(lhs)
-            :($lhs = @trace($rhs, $(addr_expr)))
+            :($lhs = trace($rhs, $(addr_expr)))
         else
             e
         end
--- a/src/dsl/dsl.jl
+++ b/src/dsl/dsl.jl
@@ -108,9 +108,36 @@ function resolve_gen_macros(expr, __module__)
     end
 end
 
+function extract_quoted_exprs(expr)
+    quoted_exprs = []
+    expr = MacroTools.prewalk(expr) do e
+        if MacroTools.@capture(e, :(quoted_))
+            push!(quoted_exprs, e)
+            Expr(:placeholder, length(quoted_exprs))
+        else
+            e
+        end
+    end
+    return expr, quoted_exprs
+end
+
+function insert_quoted_exprs(expr, quoted_exprs)
+    expr = MacroTools.prewalk(expr) do e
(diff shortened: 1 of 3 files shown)


ztangent:
@bzinberg did you test if this works in the SML? I suspect it won't because the static parser explicitly tries to match `@trace` expressions. (The `@trace` macro is only defined [for the DML](https://github.com/probcomp/Gen/blob/964d8a96205de6bb13e69f13eec7eaff7906c952/src/dynamic/dynamic.jl#L63), not the SML.)

(Note that my suggested fix in #222 might not work for the SML either..., because it probably won't parse `Gen.@trace` as `@trace`.)

Separately, is `trace` a function? Can you call something defined as a macro as a function instead...?

bzinberg:
> Separately, is `trace` a function? Can you call something defined as a macro as a function instead...?

Aha!  My understanding was that every Julia macro `@foo` also comes with a function `foo` which operates on ASTs.  Looks like I was wrong (and Gen also has a function called `trace`, which hid this fact from me)!

ztangent:
Finally got around to fixing this (#222) and fixed #235 along the way!

The strategy I ended up using, which works across the DML and SML, was to desugar the tilde syntax to `GlobalRef` versions of the `@trace` macrocall, ensuring that macro-expansion for the DML always accurately looks up the definition of `@trace` in the `Gen` module. 

In addition, the `@gen` macro now pre-resolves all properly-scoped references to `@trace` and `@param` to `GlobalRef` versions. The SML parser then pattern matches against these globally referenced `@trace` and `@param` expressions (whether pre-resolved or desugared from tilde expressions), instead of the unqualified references to `@trace` and `@param`. This means that the SML parser now obeys proper scoping rules for uses of `@trace` and `@param`, which it previously did not. As such, the following examples now compile as expected:
```julia
import Gen
Gen.@gen (static) function f1()
  x = Gen.@trace(Gen.normal(0, 1), :x)
  return x
end

Gen.@gen function f2()
  x  ~ Gen.normal(0, 1)
  return x
end
```
while the following example, which previously would have compiled, now (correctly) throws an UndefVar error for its unqualified use of `@trace`: 
```julia
import Gen
Gen.@gen function f()
  x = @trace(Gen.normal(0, 1), :x)
  return x
end
```

Finally, I added some code that uses `MacroTools.prewalk` to extract and protect any quoted expressions before pre-resolving macros and tilde desugaring, and then re-inserting those quoted expressions after those pre-processing steps. This fixes the issue in #235 where tilde expressions nested within quotes were desugared.

ztangent:
I wanted to add that these changes should also pave the way for supporting macro calls in the SML (as suggested in #185, #184), while still maintaining the correct scoping rules for the surface syntax! 

In particular, now that the SML parser now only sees globally referenced versions of `@trace` and `@param`, we can use MacroTools to match against them, remap them to some other specialized expression (to protect them from macro-expansion), and then perform macro-expansion of all the other code before mapping them back. I was almost going to implement this, but figured I should hold off for a future PR.

src/dsl/dsl.jl
```diff
@@ -71,21 +72,77 @@ function address_from_expression(lhs)
...
+            idx = p.args[1]
+            quoted_exprs[idx]
+        else
+            e
+        end
+    end
+    return expr
+end
+
+function preprocess_body(expr, __module__)
```
bzinberg, line 137: Could you add your explanation in https://github.com/probcomp/Gen.jl/pull/223#issuecomment-646420005 as a docstring here, or wherever you think it best fits?

bzinberg (commented):
Awesome!!

alex-lew:
This is awesome! Today I learned about `GlobalRef` :-) Thanks so much for taking this on, Xuan!

src/dsl/dsl.jl
```diff
@@ -71,21 +72,77 @@ function address_from_expression(lhs)
 end
 
 function desugar_tildes(expr)
+    trace_ref = GlobalRef(@__MODULE__, Symbol("@trace"))
+    line_num = LineNumberNode(1, :none)
```
alex-lew, line 76: @ztangent Any idea if this changes anything re: error messages? i.e., do crashes still report the right line number for `@trace` calls?

ztangent: Pretty sure the previous thing that was being done (where we were constructing `@trace(addr)` expressions as a quoted expression) also wasn't including line numbers!

ztangent: Yup, confirmed:
```julia
julia> dump(:(@trace(call, addr)))
Expr
  head: Symbol macrocall
  args: Array{Any}((4,))
    1: Symbol @trace
    2: LineNumberNode
      line: Int64 1
      file: Symbol none
    3: Symbol call
    4: Symbol addr
```

So at the very least it's not a regression.

src/dsl/dsl.jl
ztangent: Just added explanatory comments to each line. Decided against docstrings for now -- we probably want to add docstrings consistently to all the parsing code at some later point.

ztangent:
@bzinberg @marcoct okay if I go ahead and merge?

marcoct:
Sounds good to me!

PR #225, opened 2020-04-01
Initial setup for Dirichlet distribution in Gen 

Kenta426:
I need Dirichlet distribution for my prior and quickly wrote a starter code without logpdf_grad. One way to implement logpdf_grad is to use reparameterization trick (see "Implicit Reparameterization Gradients" Figurnov et al, 2019). This method is used by other PPL, too. Probably need someone's assistance on the implementation.


xukai92:
You can also try to wrap over [Bijectors.jl](https://github.com/TuringLang/Bijectors.jl). It also has the code for gradient in https://github.com/TuringLang/Bijectors.jl/blob/master/src/bijectors/simplex.jl

Kenta426:
It seems like the build failed during installing SpecialFunctions.jl 
Could someone take a look and give me suggestions how to fix? Thank you

bzinberg:
Hi @Kenta426, thanks for pointing this out.  I think https://github.com/probcomp/Gen/pull/246 fixes this issue in master, after testing it out on this branch in https://github.com/probcomp/Gen/pull/225/commits/6333bb72764ccf3f486fdfedbca2da642012bf22.  You should be good to go now -- looks like the unit test you added is producing an error though.

ztangent:
@Kenta426 this is probably not the cause of the CI failures, but you may also want to make sure that your commits match the new directory structure for built-in distributions introduced in #243! Basically all built-in distributions should now be located in `src/modeling_library/distributions`

Kenta426:
@bzinberg Thank you for the update!
@ztangent  Yes, I didn't move the new file to `src/modeling_library/distributions` folder. Thank you for pointing that out

zhixuedu:
@Kenta426 I suppose you are working on Gaussian Mixture Models (with Dirichlet process) in Gen? Would it be possible you can share an simple example of it using Gen? I would really appreciate that!

PR #233, opened 2020-04-07
Clarify that `particle_filter_step!` falls back to the particle's previous trace for choices that exist there

bzinberg:
# Summary

Documentation-only change, requesting confirmation that the proposed doc is accurate

### DIFF
--- a/src/inference/particle_filter.jl
+++ b/src/inference/particle_filter.jl
@@ -117,8 +117,9 @@ model's internal proposal is used for proposing new latent state.  That is, for
 each particle,
 
 * The proposal function `proposal` is evaluated with arguments `Tuple(t_old,
-  proposal_args...)`, and produces a trace (call it `proposal_trace`); and
-* The old trace (call it `t_old`) is replaced by a new trace (call it `t_new`).
+  proposal_args...)` (where `t_old` is the old model trace), and produces its
+  own trace (call it `proposal_trace`); and
+* The old model trace is replaced by a new model trace (call it `t_new`).
 
 The choicemap of `t_new` satisfies the following conditions:
 
--- a/src/inference/particle_filter.jl
+++ b/src/inference/particle_filter.jl
@@ -111,7 +111,17 @@ end
     particle_filter_step!(state::ParticleFilterState, new_args::Tuple, argdiffs,
         observations::ChoiceMap, proposal::GenerativeFunction, proposal_args::Tuple)
 
-Perform a particle filter update, where the model arguments are adjusted, new observations are added, and some combination of a custom proposal and the model's internal proposal is used for proposing new latent state (whatever is not proposed from the custom proposal will be proposed using the model's internal proposal).
+Perform a particle filter update, where the model arguments are adjusted, new
+observations are added, and some combination of a custom proposal and the
+model's internal proposal is used for proposing new latent state.  That is, for
+each particle, the value of a random choice at address `a` in the new trace for
+that particle is chosen by the following rule:
+
+* If the proposal's choicemap has a random choice at address `a`, use that
+  value.
+* Otherwise, if the choicemap of the particle's previous trace has a random
+  choice at address `a`, use that value.
+* Otherwise, sample the value from the model's internal proposal.
 """
 function particle_filter_step!(state::ParticleFilterState{U}, new_args::Tuple, argdiffs::Tuple,
         observations::ChoiceMap, proposal::GenerativeFunction, proposal_args::Tuple) where {U}
--- a/src/inference/particle_filter.jl
+++ b/src/inference/particle_filter.jl
@@ -116,11 +116,10 @@ observations are added, and some combination of a custom proposal and the
 model's internal proposal is used for proposing new latent state.  That is, for
 each particle, the value of a random choice at address `a` in the new trace for
 that particle is chosen by the following rule:
-
+* The particle's previous trace must not have a random choice at address `a`:
+  overwriting old state is not allowed.
 * If the proposal's choicemap has a random choice at address `a`, use that
   value.
-* Otherwise, if the choicemap of the particle's previous trace has a random
-  choice at address `a`, use that value.
 * Otherwise, sample the value from the model's internal proposal.
 """
 function particle_filter_step!(state::ParticleFilterState{U}, new_args::Tuple, argdiffs::Tuple,
--- a/src/inference/particle_filter.jl
+++ b/src/inference/particle_filter.jl
@@ -114,13 +114,26 @@ end
 Perform a particle filter update, where the model arguments are adjusted, new
 observations are added, and some combination of a custom proposal and the
 model's internal proposal is used for proposing new latent state.  That is, for
-each particle, the value of a random choice at address `a` in the new trace for
-that particle is chosen by the following rule:
-* The particle's previous trace must not have a random choice at address `a`:
-  overwriting old state is not allowed.
-* If the proposal's choicemap has a random choice at address `a`, use that
-  value.
-* Otherwise, sample the value from the model's internal proposal.
+each particle,
+
+* The proposal function `proposal` is evaluated with arguments `Tuple(t_old,
+  proposal_args...)`, and produces a trace (call it `proposal_trace`); and
+* The old trace (call it `t_old`) is replaced by a new trace (call it `t_new`).
+
+The choicemap of `t_new` satisfies the following conditions:
+
+1. `get_choices(t_old)` is a subset of `get_choices(t_new)`;
+2. `observations` is a subset of `get_choices(t_new)`;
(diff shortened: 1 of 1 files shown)


alex-lew:
I think Gen's `particle_filter_step!` assumes that *all* choices in the original trace *will* be copied over -- i.e., not resampled by the proposal, and not deleted -- but I could be wrong; can't dive into the code at the moment but maybe @marcoct can confirm? 

EDIT: @bzinberg -- actually just checked, and this is the case. See line 133, which asserts that the discard is empty; this means no choices were deleted or overwritten from the original model.

We could create an issue for creating a more general version of SMC, which allowed overwriting/deleting choices from previous step, provided that the user provides a "backward proposal" as well.

bzinberg:
Ah, thanks for pointing that out @alex-lew -- I looked briefly before but didn't see that part.  I updated the doc to reflect this assumption.

BTW, when referring to specific lines, please use a [permalink](https://help.github.com/en/github/managing-your-work-on-github/creating-a-permanent-link-to-a-code-snippet) instead of a line number, as line numbers go stale and GitHub has extra UI for permalinks.

src/inference/particle_filter.jl
```diff
@@ -111,7 +111,16 @@ end
     particle_filter_step!(state::ParticleFilterState, new_args::Tuple, argdiffs,
         observations::ChoiceMap, proposal::GenerativeFunction, proposal_args::Tuple)
 
-Perform a particle filter update, where the model arguments are adjusted, new observations are added, and some combination of a custom proposal and the model's internal proposal is used for proposing new latent state (whatever is not proposed from the custom proposal will be proposed using the model's internal proposal).
+Perform a particle filter update, where the model arguments are adjusted, new
+observations are added, and some combination of a custom proposal and the
+model's internal proposal is used for proposing new latent state.  That is, for
+each particle, the value of a random choice at address `a` in the new trace for
+that particle is chosen by the following rule:
+* The particle's previous trace must not have a random choice at address `a`:
```
alex-lew: @bzinberg Not sure about this phrasing: the new model trace includes both the old random choices *and* new random choices.

src/inference/particle_filter.jl
```diff
@@ -111,7 +111,16 @@ end
...
+  overwriting old state is not allowed.
+* If the proposal's choicemap has a random choice at address `a`, use that
+  value.
+* Otherwise, sample the value from the model's internal proposal.
```
alex-lew: @bzinberg Actually, looking at the code, I don't think this happens? In the current code, the internal proposal is not used at all (except, I guess, for untraced randomness) -- the user's proposal is responsible for sampling all new choices. (This is because `update` is used.)

(It's not clear how to relax this restriction without losing incremental computation, in Gen's current GFI. I think a previous version of Gen had more variants of `update`, including ones that sampled "missing" values from the internal proposal.)

alex-lew: If `generate` were used [here](https://github.com/probcomp/Gen/blob/efc5a00b787cabf1e26e6849727ae3f10dc44868/src/inference/particle_filter.jl#L131) then the internal proposal would be used to generate missing addresses. But this wouldn't give us an incrementally computed weight.

bzinberg:
Come to think of it, I might be missing a piece of the picture.  When would a user want to use `particle_filter_step!` with a proposal that is anything other than, essentially, the conditional distribution `P(x_{t+1} | x_t)`?  Are there situations where the user specifies an HMM in some other way, or situations where the user specifies the HMM that way and gets inference benefits from using a different step proposal?  (Rejuvenation proposals I get, but step proposals I'm not sure.)  Would the main case be when the user gives a broader-than-prior proposal in order to increase particle diversity?

src/inference/particle_filter.jl
marcoct: @alex-lew @bzinberg update does use the internal proposal by default to fill in any choices that are not determined by the previous trace or the constraints choice map.

src/inference/particle_filter.jl
marcoct: > Come to think of it, I might be missing a piece of the picture. When would a user want to use `particle_filter_step!` with a proposal that is anything other than, essentially, the conditional distribution `P(x_{t+1} | x_t)`? Are there situations where the user specifies an HMM in some other way, or situations where the user specifies the HMM that way and gets inference benefits from using a different step proposal? (Rejuvenation proposals I get, but step proposals I'm not sure.) Would the main case be when the user gives a broader-than-prior proposal in order to increase particle diversity?

@bzinberg The optimal proposal would be P(x_{t + 1} | x_t, y_t) -- note the dependence on the newest observation y_t. But usually sampling (and evaluating the density of) the conditional distribution P(x_{t+1} | x_t, y_t) isn't possible. But it is possible to use step proposals that are better approximations to P(x_{t+1} | x_t, y_t) than P(x_{t + 1} | x_t) is.

marcoct:
> Come to think of it, I might be missing a piece of the picture. When would a user want to use `particle_filter_step!` with a proposal that is anything other than, essentially, the conditional distribution `P(x_{t+1} | x_t)`? Are there situations where the user specifies an HMM in some other way, or situations where the user specifies the HMM that way and gets inference benefits from using a different step proposal? (Rejuvenation proposals I get, but step proposals I'm not sure.) Would the main case be when the user gives a broader-than-prior proposal in order to increase particle diversity?

@bzinberg The optimal proposal would be P(x_{t + 1} | x_t, y_t) -- note the dependence on the newest observation y_t. But usually sampling (and evaluating the density of) the conditional distribution P(x_{t+1} | x_t, y_t) isn't possible. But it is possible to use step proposals that are better approximations to P(x_{t+1} | x_t, y_t) than P(x_{t + 1} | x_t) is.

bzinberg:
Ah, got it.  Thanks!

src/inference/particle_filter.jl
bzinberg: @marcoct, in that case, is this proposed doc accurate?

src/inference/particle_filter.jl
marcoct: @bzinberg  I think a more precise characterization would be: "The random choices in the new trace satisfies the following conditions: (i) All choics in the previous trace are included in the new trace, (ii) all new observed choices are included in the new trace, and (iii) all choices in the proposal choice map are included in the new trace. It is an error if no such trace exists in the support of the model (with the new arguments). If such a trace exists, then the random choices not determined by the previous trace, proposal choice map, or observations, are sampled using the internal proposal.

bzinberg: Is there also some requirement to the effect of "the proposal cannot contain choices at addresses that exist in the old trace"?

marcoct: Yes. Technically, I think that's a consequence of (i) and (iii) because a choice cannot be taken from both the previous trace and the proposal choice map.

src/inference/particle_filter.jl
marcoct: @bzinberg In my reading, the proposed doc implies that there are no addresses copied from the previous trace to the new trace, which would not be accurate.

src/inference/particle_filter.jl
bzinberg: It could (be construed as so) if there was a requirement that it have the same value in both places.

src/inference/particle_filter.jl
alex-lew: @marcoct Ah, thanks... I think I confused it with custom-proposal MH, which errors if you don't provide every choice -- but that's because of `assess` fails to compute the backward probability, not because `update` fails to sample values for missing choices.

alex-lew:
(@bzinberg And that's just the optimal *local* proposal -- the truly optimal proposal, if particle_filter_step! allowed changing previous random choices, would not look at the previous state at all and sample all latents from the posterior.)

src/inference/particle_filter.jl
bzinberg: @marcoct, @alex-lew, updated the docstring based on your feedback -- please take a look when you get the chance

src/inference/particle_filter.jl
```diff
@@ -111,7 +111,29 @@ end
...
+3. `get_choices(proposal_trace)` is a subset of `get_choices(t_new)`.
+
+Here, when we say one choicemap `a` is a "subset" of another choicemap `b`, we
+mean that all keys that occur in `a` also occur in `b`, and the values at those
+addresses are equal.
+
+It is an error if no trace `t_new` satisfying the above conditions exists in
+the support of the model (with the new arguments). If such a trace exists, then
+the random choices not determined by the above requirements are sampled using
+the internal proposal.
```
bzinberg: @marcoct, @alex-lew, updated the docstring based on your feedback -- please take a look when you get the chance

marcoct: Looks good to me!

alex-lew: Me too!

Issue #209, opened 2020-03-11
`load_generated_functions` always puts generated functions in `Main` module


georgematheos:
Sometimes we want to work with a Gen model in a module other than Main.  However, currently `Gen.load_generated_functions` is hardcoded to load the functions in the `Main` module.

This means if I do something along the lines of
```julia
module M1
using Gen
@gen (static) function ()....
...
end
Gen.load_generated_functions()
...
Gen.update(trace, ...)
end
```
we sometimes get an `UndefVarError: NoChange not defined`.  This originates from pretty deep in the macroexpansion for static trace update, and is a mess to understand (debugging these macros is painful!). It looks to me like the basic issue is that when we call `load_generated_functions`, we try to macroexpand some functions in the `Main` module, but these functions try to call `Gen.NoChange()`.  However, Gen is only imported into `M1`, not into `Main`.

***
Proposed solution:
Change `load_generated_functions` to `@load_generated_functions`.  Then instead of macroexpanding everything into `Main`, we can expand into `__module__` (ie. whatever module `@load_generated_functions` was called from).  This should be an easy change but will require updating the documentation.

ztangent:
I'd appreciate this changing / extending this behavior as well, especially now that I have a fair amount of library code written as static generative functions!

Another possible solution would be to preserve the existing behavior, but the add another method for `load_generated_functions` which takes a `module` argument:

```julia
function load_generated_functions(__module__::Module)
    for function_defn in generated_functions
        __module__.eval(function_defn)
    end
end
```

And you'd use it as either:
```julia
module M1
using Gen
load_generated_functions(M1)
```
or
```julia
module M1
using Gen
load_generated_functions(@__MODULE__)
```
if you want to avoid retyping the module name.

@marcoct any thoughts on the above solution? And also, what do you think the default behavior should be? It seems more intuitive to me that by default we should eval in the calling module, as opposed to eval-ing in Main. I don't think changing the default behavior would actually be a breaking change in existing code either, since most of the time  `load_generated_functions` is called in a script, so it would still eval in Main.

ztangent:
This is addressed by #234 :)

PR #234, opened 2020-04-07
Deprecate load_generated_functions() in favor of @load_generated_functions.

ztangent:
Addresses #209, but preserves backwards compatibility for `load_generated_functions()`.

The main motivation for switching to a macro call is so that users don't have to call `load_generated_functions()` at the top of their script when they're using module-defined static generative functions. Instead, they can do:

```julia
module MyModule
using Gen
export foo
@gen (static) function foo(mu)
    y = @trace(normal(mu, 0), :y)
    return y
end
Gen.@load_generated_functions
end
```

And then any script that uses MyModule works without needing to call `load_generated_functions()` at the top:
```
using MyModule
@assert foo(5) == 5
```

As noted in #209 , we can't simply call `load_generated_functions()` within `MyModule` -- this is because the `Core.eval` call in `load_generated_functions` evaluates the generated functions into `Main`, but when `MyModule` is part of a Julia package, eval-ing into `Main` like that doesn't work. 

`@load_generated_functions` fixes this by evaluating within `__module__` instead. For backwards compatibility and some additional flexibility, `load_generated_functions` still works as before, and also accepts an optional module argument (so, for example, `load_generated_functions(@__MODULE__)` within `MyModule.jl` would achieve the same effect as `@load_generated_functions`). But for most intents and purposes, people should probably use `@load_generated_functions` both within scripts and modules. The documentation has been updated accordingly.

### DIFF
--- a/docs/src/ref/modeling.md
+++ b/docs/src/ref/modeling.md
@@ -451,22 +451,34 @@ can instead be implemented as:
 ```
 
 ### Loading generated functions
-Before a function with a static annotation can be used, the [`load_generated_functions`](@ref) method must be called:
+Before a function with a static annotation can be used, the [`@load_generated_functions`](@ref) macro must be called:
 ```@docs
-load_generated_functions
+@load_generated_functions
 ```
 Typically, one call to this function, at the top level of a script, separates the definition of generative functions from the execution of inference code, e.g.:
 ```julia
-using Gen: load_generated_functions
+using Gen: @load_generated_functions
 
 # define generative functions and inference code
 ..
 
 # allow static generative functions defined above to be used
-load_generated_functions()
+@load_generated_functions
 
 # run inference code
 ..
+```
+
+When static generative functions are defined in a Julia module, [`@load_generated_functions`](@ref) should be called after all static functions are defined:
+
+```julia
+module MyModule
+using Gen
+# Include code that defines static generative functions
+include("my_static_gen_functions.jl")
+# Load generated functions defined in this module
+Gen.@load_generated_functions
+end
 ```
 
 ### Performance tips
--- a/docs/src/ref/modeling.md
+++ b/docs/src/ref/modeling.md
@@ -463,7 +463,7 @@ using Gen: @load_generated_functions
 ..
 
 # allow static generative functions defined above to be used
-@load_generated_functions
+@load_generated_functions()
 
 # run inference code
 ..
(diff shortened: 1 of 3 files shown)


docs/src/ref/modeling.md
```diff
@@ -451,24 +451,36 @@ can instead be implemented as:
...
+```julia
+module MyModule
+using Gen
+# Include code that defines static generative functions
+include("my_static_gen_functions.jl")
+# Load generated functions defined in this module
+Gen.@load_generated_functions
+end
+```
+
```
alex-lew, line 483: Maybe add a line saying "consumers of the module need not call `@load_generated_functions` unless they also define their own static generative functions"?

marcoct: Do you need `Gen.` here? `@load_generated_functions` is exported by the Gen module right? If it's not needed, I think it would be better to not include it.

docs/src/ref/modeling.md
```diff
@@ -451,24 +451,36 @@ can instead be implemented as:
...
 
 # allow static generative functions defined above to be used
-load_generated_functions()
+@load_generated_functions
 
 # run inference code
 ..
 ```
 
+When static generative functions are defined in a Julia module, [`@load_generated_functions`](@ref) should be called after all static functions are defined:
```
marcoct, line 472: This is now no different than when you use it not within a module right? Is it necessary to add this?

docs/src/ref/modeling.md
ztangent: Sure, okay! I thought it might be slightly better style to encourage it so that readers of the code know exactly what generated functions are being loaded, but I'm fine either way.

Also yes that makes sense @alex-lew -- will add that line.

docs/src/ref/modeling.md
ztangent: Hmm, I think it's good for clarity and as an example design pattern when you're writing library code which contains static generative functions. Importantly, as @alex-lew noted, once you call `@load_generated_functions` in a module, you no longer need to call `@load_generated_functions` in a script that uses the module. I don't think that's well communicated by the first usage example alone.

marcoct: Okay. In that case I think it could help to add a second code block showing the module being used. e.g.:

```julia
module MyModule
using Gen

@gen (static) function example_gen_fn()
...
end

# Load generated functions defined in this module
Gen.@load_generated_functions
end
```
Then, when using this module, you don't need to call `@load_generated_functions` yourself to use the static generative functions defined in the module, e.g.:
```julia
using Gen
using MyModule: example_gen_fn
trace = simulate(example_gen_fn, ())
```

marcoct:
@ztangent I noticed you're using `@load_generated_functions` without parentheses? I think it would make it clearer (to readers, and in some cases to Julia as well) how code that uses this should be parsed by adding the parentheses `@load_generated_functions()`.

docs/src/ref/modeling.md
ztangent: Sounds good!

ztangent:
Made the changes suggested! @marcoct @alex-lew

marcoct (approved):
Looks good to me!

PR #236, opened 2020-04-11
Add convenience forms for update and regenerate with optional args and argdiffs

marcoct:
Addresses https://github.com/probcomp/Gen/issues/229

### DIFF
--- a/src/gen_fn_interface.jl
+++ b/src/gen_fn_interface.jl
@@ -282,7 +282,7 @@ end
         trace, constraints::ChoiceMap, args::Tuple;
         argdiffs::Tuple=map((_) -> UnknownChange(), args))
 
-Convenience form of `update` with keyword arguments for argdiffs and constraints.
+Convenience form of `update` with keyword argument for argdiffs.
 """
 function update(trace, constraints::ChoiceMap, args::Tuple;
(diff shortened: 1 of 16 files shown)


src/gen_fn_interface.jl
```diff
@@ -272,10 +272,25 @@ that if the original `trace` was generated using non-default argument values,
...
+    update(trace;
+        args::Tuple=get_args(trace),
+        argdiffs::Tuple=map((_) -> NoChange(), args),
+        constraints::ChoiceMap=EmptyChoiceMap())
+
+Form of `update` with keyword arguments providing common defaults.
+"""
+function update(trace;
+        args::Tuple=get_args(trace),
+        argdiffs::Tuple=map((_) -> NoChange(), args),
```
alex-lew: Should the default be `NoChange` or `UnknownChange`?

marcoct: Now there are two convenience forms:

If you provide args, then argdiffs is an optional keyword argument that defaults to UnknownChange. (This is so you don't need to construct the unknown change argdiffs).

If you do not provide args then argdiffs is automatically set to NoChanges. (This is a common case e.g. for top-level generative functions on which we are doing MCMC or optimization).

PR #238, opened 2020-04-12
Improve `hmc` docstring

ali-ramadhan:
This PR just expands the `hmc` docstring to explain the arguments, especially `L` and `eps`.


marcoct (approved):
Thanks @ali-ramadhan this is great :)

PR #241, opened 2020-04-14
Remember observations in PF state + minor PF interface extensions.

ztangent:
In some custom SMC code I'm currently writing, it's helpful to have access to the history of observations provided to the particle filter as a choicemap. In particular, I want to do a form of "importance resampling rejuvenation" when the effective sample size gets too small, where I throw out a bunch of old traces in the particle filter, and sample totally new traces using a proposal distribution. This requires conditioning on the observations provided so far, which are not currently stored separately from all the choices associated with each trace in the particle filter. 

I imagine there are other custom SMC extensions that would find this useful as well -- for example, data-driven MCMC proposals within a rejuvenation move.

This PR adds support for that by storing the initial observations during `initialize_particle_filter`, and then subsequently updating the state with new observations every time `particle_filter_step!` is called. 

I've also made a few minor extensions to the particle filter interface, which I think are helpful for users writing custom SMC code.

### DIFF
--- a/docs/src/ref/pf.md
+++ b/docs/src/ref/pf.md
@@ -5,6 +5,9 @@ particle_filter_step!
 maybe_resample!
 log_ml_estimate
 get_traces
+get_observations
 get_log_weights
+get_log_norm_weights
+get_effective_sample_size
 sample_unweighted_traces
 ```
--- a/src/inference/particle_filter.jl
+++ b/src/inference/particle_filter.jl
@@ -21,6 +21,7 @@ mutable struct ParticleFilterState{U}
     log_weights::Vector{Float64}
     log_ml_est::Float64
     parents::Vector{Int}
+    observations::ChoiceMap
 end
 
 """
@@ -29,8 +30,16 @@ end
 Return the vector of traces in the current state, one for each particle.
 """
 function get_traces(state::ParticleFilterState)
-    state.traces
+    return state.traces
+end
 
+"""
+    observations = get_observations(state::ParticleFilterState)
+
+Return the choicemap of observations provided to the particle filter thus far.
+"""
+function get_observations(state::ParticleFilterState)
+    return state.observations
 end
 
 """
@@ -41,7 +50,26 @@ Return the vector of log weights for the current state, one for each particle.
 The weights are not normalized, and are in log-space.
 """
 function get_log_weights(state::ParticleFilterState)
-    state.log_weights
+    return state.log_weights
+end
+
+"""
+    log_norm_weights = get_log_norm_weights(state::ParticleFilterState)
+
+Return the vector of log normalized weights for the current state.
+"""
+function get_log_norm_weights(state::ParticleFilterState)
+    _, weights = normalize_weights(state.log_weights)
+    return weights
+end
+
+"""
+    ess = get_effective_sample_size(state::ParticleFilterState)
+
+Return the effective sample size for the current state.
+"""
+function get_effective_sample_size(state::ParticleFilterState)
+    return effective_sample_size(get_log_norm_weights(state))
 end
 
 """
@@ -87,7 +115,7 @@ function initialize_particle_filter(model::GenerativeFunction{T,U}, model_args::
         log_weights[i] = model_weight - prop_weight
     end
     ParticleFilterState{U}(traces, Vector{U}(undef, num_particles),
-        log_weights, 0., collect(1:num_particles))
+        log_weights, 0., collect(1:num_particles), observations)
 end
 
 """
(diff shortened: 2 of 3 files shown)


alex-lew:
@ztangent Would it be feasible to just pass the relevant choice maps as an argument to the proposal? (i.e., in the `proposal_args` parameter of `particle_filter_step!`)

> I imagine there are other custom SMC extensions that would find this useful as well -- for example, data-driven MCMC proposals within a rejuvenation move.

Ah, you can already do this :-) Rejuvenation moves are not implemented as calls to `particle_filter_step!` in Gen -- you just run MCMC on the current traces, using whatever proposal arguments you want (including observations).

ztangent:
@alex-lew Re the first question, I'm not doing the resampling as part of `particle_filter_step!`, so I can't do that. `particle_filter_step!` doesn't allow me to discard past choices I've already made (there's an explicit check against it), but what I'm trying to do is resample all past choices (and throw away all my past work) in order to (hopefully) better explain the data.

And yup, I'm aware about the second point! But it feels preferable to have an interface where the observations are already contained in the state of the particle filter, so that users don't have to additionally maintain the list of observations in some external code.

Just to clarify, I know that in either case I can implement things by storing the list of observations externally, rather than in the ParticleFilterState. But I'd like not to have to do that, since it means passing more arguments around. Conceptually, it also makes sense to me that the particle filter remembers the history of observations. It means I can write my function as `pf_importance_resample!(state, proposal, proposal_args, ...)`, rather than as `pf_importance_resample!(state, observations, proposal, proposal_args, ...)`. In the simplest case where I just want to use the prior as the proposal, I can just do `pf_importance_resample!(state)`.

The only downside I can think of right now is that this increases memory usage over time, which is bad for very long sequences. But traces have that issue right now too, in that new random choices will keep getting added as time goes on. Perhaps those could be collectively addressed in the future by adding functions that purge the history of past observations beyond some horizon.

ztangent:
@marcoct here's the kind of resampling I would like to do! Currently it requires passing the choicemap of all past observations as an argument, but it feels intuitive to me that I shouldn't have to.

```julia
function pf_importance_resample!(pf_state::Gen.ParticleFilterState,
                                 observations::ChoiceMap,
                                 proposal::GenerativeFunction,
                                 proposal_args::Tuple, propose_frac::Real=0.5)
    # Compute number of traces to propose vs resample
    n_proposed = Int(round(propose_frac * length(pf_state.traces)))
    n_resampled = length(pf_state.traces) - n_proposed
    # Resample old traces
    probs = softmax(get_log_weights(pf_state))
    for i in 1:n_resampled
        pf_state.parents[i] = categorical(probs)
        pf_state.new_traces[i] = pf_state.traces[pf_state.parents[i]]
    end
    # Propose new traces
    model = get_gen_fn(pf_state.traces[1])
    model_args = get_args(pf_state.traces[1])
    prop_weights = zeros(n_proposed)
    for i in 1:n_proposed
        old_trace = pf_state.traces[i]
        prop_choices, prop_weight, _ = propose(proposal, proposal_args)
        constraints = merge(observations, prop_choices)
        new_trace, model_weight = generate(model, model_args, constraints)
        pf_state.traces[n_resampled + i] = new_trace
        prop_weights[i] = model_weight - prop_weight
    end
    # Resample proposed traces
    probs = softmax(prop_weights)
    for i in 1:n_proposed
        idx = n_resampled + categorical(probs)
        pf_state.new_traces[n_resampled + i] = pf_state.traces[idx]
        pf_state.parents[n_resampled + i] = n_resampled + i
    end
    # Set all log weights to zero
    pf_state.log_weights .= 0
    # Swap references
    tmp = pf_state.traces
    pf_state.traces = pf_state.new_traces
    pf_state.new_traces = tmp
    return nothing
end
```

 @alex-lew I discussed this briefly yesterday, and agreed that this is a particular case of the more general issue that traces don't make a distinction between choices that were constrained by user (via generate or update), v.s. choices that were simply sampled as part of the proposal. Ultimately, a more unified fix would be to update the trace interface and the GFI to make that distinction. But in the meantime, this feels to me like a possibly reasonable thing to do for particle filters in particular.

I do also want to note though that after some experiments, it's no longer clear to me that I will need this sort of importance resampling step to make inference work well -- so I'd be fine if this PR doesn't ultimately get merged (though I would still like to make the minor interface extensions as a separate PR).

ztangent:
Closing this because this functionality is no longer urgent for me and there are probably more principled ways to have traces distinguish user-supplied choices from generated ones that we could discuss in the future. I may still do a separate PR for the particle filter interface extensions.

PR #126, opened 2019-09-23
Skeletal implementation of a binomial distribution

ghost:
`logpdf_grad` will need to be implemented.

PR #243, opened 2020-04-19
Add binomial and negative binomial distributions.

ztangent:
Both fully implemented with test cases. Supersedes #126.

Note that to avoid the namespace clash with `Base.binomial` (which computes binomial co-efficients), I decided to use the name `binom` instead for the Binomial distribution. The Negative Binomial is called `neg_binom`, for consistency.

I also took the liberty of reorganizing the files for the built-in distributions, because it felt like they were starting to clutter the `modeling_library` directory. They are now in the subdirectory `modeling_library/distributions`.


src/modeling_library/distributions/binom.jl
```diff
@@ -0,0 +1,28 @@
+struct Binomial <: Distribution{Int} end
+
+"""
+	binom(n::Integer, p::Real)
+Sample an `Int` from the Binomial distribution with parameters `n` and `p`.
```
bzinberg:
```suggestion
Sample an `Int` from the Binomial distribution with parameters `n` (number of
trials) and `p` (probability of success in each trial).
```

src/modeling_library/distributions/binom.jl
```diff
@@ -0,0 +1,28 @@
...
+	coefficient = loggamma(n + 1) - loggamma(n + 1 - x) - loggamma(x + 1)
+	coefficient + x * log(p) + (n - x) * log(1 - p)
+end
+
+function logpdf_grad(::Binomial, x::Integer, n::Integer, p::Real)
+	(nothing, nothing, (x / p - (n - x) / (1 - p)))
+end
+
+function random(::Binomial, n::Integer, p::Real)
+	rand(Distributions.Binomial(n, p))
```
bzinberg, line 19: I think the `rand` call shouldn't be here?

(I think I'm missing something, because I don't see how the unit test passes)

src/modeling_library/distributions/neg_binom.jl
```diff
@@ -0,0 +1,32 @@
...
+"""
+    neg_binom(r::Real, p::Real)
+Sample an `Int` from a Negative Binomial distribution. Returns the number of
+failures before the `r`th success in a sequence of independent Bernoulli trials.
+`r` is the number of successes and `p` is the probability of success per trial.
+"""
+const neg_binom = NegativeBinomial()
+
+function Gen.logpdf(::NegativeBinomial, x::Int, r::Real, p::Real)
+    Gen.Distributions.logpdf(Gen.Distributions.NegativeBinomial(r, p), x)
```
bzinberg:
```suggestion
    Distributions.logpdf(Distributions.NegativeBinomial(r, p), x)
```

src/modeling_library/distributions/neg_binom.jl
```diff
@@ -0,0 +1,32 @@
...
+end
+
+function Gen.logpdf_grad(::NegativeBinomial, x::Int, r::Real, p::Real)
+    r_grad = x == 0 ? log(p) : sum(1/(x+r-i) for i in 1:x) + log(p)
+    p_grad = x >= 0 ? r/p - (1/(1-p) * x) : 0.0
+    (nothing, r_grad, p_grad)
+end
+
+function Gen.random(::NegativeBinomial, r::Real, p::Real)
+    rand(Gen.Distributions.NegativeBinomial(r, p))
```
bzinberg:
```suggestion
    return rand(Distributions.NegativeBinomial(r, p))
```

bzinberg: Ditto about the `rand` call

src/modeling_library/distributions/neg_binom.jl
```diff
@@ -0,0 +1,32 @@
+struct NegativeBinomial <: Distribution{Int} end
+
+"""
+    neg_binom(r::Real, p::Real)
+Sample an `Int` from a Negative Binomial distribution. Returns the number of
+failures before the `r`th success in a sequence of independent Bernoulli trials.
+`r` is the number of successes and `p` is the probability of success per trial.
```
bzinberg: What if `r` is not an integer?  Presumably this paragraph defines the distribution when `r` is a nonnegative integer and then we define the general case by analytic continuation (in the complex analysis sense) of the PDF, or something like that?

bzinberg (commented):
Thanks @ztangent!  Really appreciating your attention to code quality.

BTW, general comment about chunking PRs, in case it wasn't already on your radar: if you send a PR that does two things, then there are two opportunities for it to get delayed in review: if either we need to talk at length about the first thing (code moving/renaming), or we need to talk at length about the second thing (implementation of binomial distribution).  In general I suggest decoupling independently-reviewable changes for this reason -- though in the present case it's fine :slightly_smiling_face:  (This would be a bigger deal if we had a higher volume of PRs.)

src/modeling_library/distributions/binom.jl
ztangent: It's how you sample from a `Distributions.jl` distribution! Almost all the Gen distributions are implemented that way, see for e.g.:

https://github.com/probcomp/Gen/blob/8ca4dc24844f93502b0f0a766f35673cd82a1ede/src/modeling_library/geometric.jl

src/modeling_library/distributions/neg_binom.jl
ztangent: See answer above!

src/modeling_library/distributions/neg_binom.jl
ztangent: Yes, `r` can be any positive real! Will update the docs accordingly.

src/modeling_library/distributions/neg_binom.jl
```diff
@@ -0,0 +1,32 @@
...
+
+function Gen.random(::NegativeBinomial, r::Real, p::Real)
+    rand(Gen.Distributions.NegativeBinomial(r, p))
+end
+
+is_discrete(::NegativeBinomial) = true
+
+(::NegativeBinomial)(r, p) = random(NegativeBinomial(), r, p)
+
+Gen.has_output_grad(::NegativeBinomial) = false
```
marcoct:
```suggestion
has_output_grad(::NegativeBinomial) = false
```

src/modeling_library/distributions/neg_binom.jl
```diff
@@ -0,0 +1,32 @@
...
+Gen.has_argument_grads(::NegativeBinomial) = (true, true)
```
marcoct:
```suggestion
has_argument_grads(::NegativeBinomial) = (true, true)
```

src/modeling_library/distributions/binom.jl
```diff
@@ -0,0 +1,29 @@
...
+	binom(n::Integer, p::Real)
+Sample an `Int` from the Binomial distribution with parameters `n` (number of
+trials) and `p` (probability of success in each trial).
+"""
+const binom = Binomial()
+
+function logpdf(::Binomial, x::Integer, n::Integer, p::Real)
+	if x < 0 return -Inf end
+	coefficient = loggamma(n + 1) - loggamma(n + 1 - x) - loggamma(x + 1)
+	coefficient + x * log(p) + (n - x) * log(1 - p)
```
marcoct:
```suggestion
	return coefficient + x * log(p) + (n - x) * log(1 - p)
```

marcoct: This isn't documented anywhere, and is done inconsistently now, but I think it would be good to add explicit return statements to function definitions defined with `function .. end`, as suggested in this style guide: https://github.com/jrevels/YASGuide#programming-guidelines (and I have noticed you have generally being doing this elsewhere, and I agree it is good style).

src/modeling_library/distributions/neg_binom.jl
```diff
@@ -0,0 +1,32 @@
...
+const neg_binom = NegativeBinomial()
+
+function Gen.logpdf(::NegativeBinomial, x::Int, r::Real, p::Real)
+    Gen.Distributions.logpdf(Gen.Distributions.NegativeBinomial(r, p), x)
+end
+
+function Gen.logpdf_grad(::NegativeBinomial, x::Int, r::Real, p::Real)
+    r_grad = x == 0 ? log(p) : sum(1/(x+r-i) for i in 1:x) + log(p)
+    p_grad = x >= 0 ? r/p - (1/(1-p) * x) : 0.0
+    (nothing, r_grad, p_grad)
```
marcoct:
```suggestion
    return (nothing, r_grad, p_grad)
```

marcoct (commented):
This is great. One comment about removing unnecessary `Gen.`, and a nudge to add explicit `return`s to `function .. end` definitions, which I think we should try to encourage in new code (and do a pass over old code at some point)

src/modeling_library/distributions/binom.jl
bzinberg: Oh, looks like Distributions.jl is doing something funky (but reasonable) -- since AFAIK the spec of `Base.rand(::Type{T})` is to return a random instance of `T`, and is not geared toward allowing more than one possible distribution on the same type of value, they are bending that rule a little (or crazily subclassing `Integer`).  :+1:

src/modeling_library/distributions/binom.jl
```diff
@@ -0,0 +1,29 @@
...
+	binom(n::Integer, p::Real)
+Sample an `Int` from the Binomial distribution with parameters `n` (number of
+trials) and `p` (probability of success in each trial).
+"""
+const binom = Binomial()
+
+function logpdf(::Binomial, x::Integer, n::Integer, p::Real)
+	if x < 0 return -Inf end
+	coefficient = loggamma(n + 1) - loggamma(n + 1 - x) - loggamma(x + 1)
+	return coefficient + x * log(p) + (n - x) * log(1 - p)
```
bzinberg: Can this instead call the implementation in Distributions.jl?

bzinberg:
LGTM once the remaining comments are addressed!

src/modeling_library/distributions/binom.jl
ztangent: @bzinberg is this just for consistency w the `neg_binom` implementation? Because it doesn't seem like we have a standard way to do this now -- some of the built-in distributions call out to Distributions.jl for logpdf, and some don't.

bzinberg: It's not for consistency, it's because re-implementing things creates more surface for bugs (including future changes to one of the copies that forget to touch the other one).  So I think we should use the existing implementations in Distribitions.jl (given that we already depend on it) where possible, unless there is particular reason not to.

marcoct: I agree that we should always use existing code in Distributions.jl whenever possible.

The implementations of `logpdf` that do not use Distributions.jl are a relic from an earlier time when gradients of logpdf were implemented by applying ReverseDiff.jl to `logpdf` (so it had to be written in a special way so that ReverseDiff.jl could work with it, and the implementations in Distribution.jl were not written that way). For a while `logpdf_grad` has been implemented separately, so I have been implementing `logpdf` for new distributions by calling Distributions.jl. I made an issue to track the cleanup: https://github.com/probcomp/Gen/issues/244

ztangent: Cool that makes sense -- just made the change! :)

ztangent:
@bzinberg @marcoct will merge this if there aren't any other issues!

marcoct (approved):
Looks good to me.

PR #254, opened 2020-05-06
Fix #253 (@dist dsl issue with `-`)

### LINKED ISSUE #253: `@dist` is confused by minus, I think
No problem:
```
julia> Gen.@dist f(a, b) = 1 + Gen.uniform(a, b);

julia> f(1, 2)
2.004503805783018
```

But:
```
julia> Gen.@dist f(a, b) = 1 - Gen.uniform(a, b);

julia> f(1, 2)
ERROR: MethodError: no method matching random(::Gen.UniformContinuous, ::Int64)
```

alex-lew:
This one-character change fixes the bug @bzinberg found in the `@dist` DSL, where `x::Real - random_things` wasn't working. We needed to change the `n_args` argument of `TransformedDistribution` from `1` to `0`:  the transformation of subtracting a random variable from a constant does not change the number of arguments that the distribution has.


bzinberg:
Thanks!  Could add the example I gave in https://github.com/probcomp/Gen/issues/253#issue-613454655 as a regression test?

bzinberg (approved):
LGTM!

PR #265, opened 2020-05-20
1 line bugfix for `get_submap` on static gen function choicemap which should return emptychoicemap

georgematheos:
My understanding is that calling `get_submap(choicemap, :x)` should return an `EmptyChoiceMap` if `:x` does not store a value or submap in `choicemap`.  However, the choicemaps for the trace of a static DSL generative function throws an error if you try to `get_submap` for an address which is not present on the trace.

This PR adds a one-liner to make sure these sorts of `get_submap` calls return `EmptyChoiceMap`.  I also added a few quick tests to the static DSL test file for the choicemaps on these generative functions.


alex-lew:
This looks good to me!

Issue #272, opened 2020-06-20
Improper module scoping in SML code generation


ztangent:
Noticed this bug while implementing the fixes in #223 -- code generation for the static modeling language incorrectly assumes sometimes that static Gen functions will only be written in a namespace that is `using Gen`, and sometimes that the `Gen` module is available (through `import Gen`). For example the following code:
```julia
import Gen
Gen.@gen (static) function foo()
    x = Gen.@trace(Gen.normal(0, 1), :x)
    return x
end
```
gives the error:
```julia
julia> foo()
ERROR: UndefVarError: propose not defined
```
because during code generation `propose` is not defined with the correct module scope. This should be easily fixed, so I'll open a PR to address it.

ztangent:
Not adding test cases as part of #273  because they require `import Gen` instead of `using Gen`, but here's a minimal example with nested imports that only works after PR #273.

```julia
module A import Gen end
A.Gen.@gen (static) function f()
    x = A.Gen.@trace(A.Gen.normal(0, 0), :x)
    return x
end

A.Gen.@gen (static) function g()
    y = A.Gen.@trace(f(), :y) + 10
    return y
end

@assert f() == 0
@assert g() == 10

tr = A.Gen.simulate(g, ())
@assert tr[] == 10
@assert tr[:y] == 0
```

PR #273, opened 2020-06-21
Fix static codegen scope

### LINKED ISSUE #235: Tilde syntax should not expand inside quoted code
If a Gen model includes a tilde expression in a quoted expression, then it is incorrectly replaced with `@trace`:

```julia
@gen function f()
   expr = :(x ~ normal(0, 1))
   return expr
end

f()
# => :(x = @trace(normal(0, 1), :x))

# Output should be: :(x ~ normal(0, 1))
```

I didn't realize `MacroTools.postwalk` would send us the insides of quoted expressions as if they weren't quoted! (Or maybe the problem is `MacroTools.@capture` being overly broad, and matching quoted expressions?)

This suggests that other places we use `MacroTools.postwalk` (e.g., in the `@dist` DSL) may have the same bug.

ztangent:
Addresses #272. Improper scoping was showing up in two areas:
- the call to `propose` in `generate_generative_function`.
- various generated methods for `StaticIRTrace`, e.g., `get_index`

This PR addresses them by interpolating the appropriately scoped references.

It also fixes a small bug that was unfortunately introduced by my fix in #223 that was also causing issues with SML parsing. Apparently my initial attempt at protecting quoted expressions from pre-processing (in order to fix #235) was too aggressive, and ended up protecting QuoteNodes with Symbol values as well -- but QuoteNodes with Symbol values are how Julia represents something like `:(Gen.@trace)`, which meant that these expressions were getting protected instead of appropriately resolved by `resolve_gen_macros`.

### DIFF
--- a/src/dsl/dsl.jl
+++ b/src/dsl/dsl.jl
@@ -112,7 +112,7 @@ end
 function extract_quoted_exprs(expr)
     quoted_exprs = []
     expr = MacroTools.prewalk(expr) do e
-        if MacroTools.@capture(e, :(quoted_))
+        if MacroTools.@capture(e, :(quoted_)) && !isa(quoted, Symbol)
             push!(quoted_exprs, e)
             Expr(:placeholder, length(quoted_exprs))
         else
--- a/src/static_ir/backprop.jl
+++ b/src/static_ir/backprop.jl
@@ -66,7 +66,7 @@ function back_pass!(back_marked, node::RandomChoiceNode)
         push!(back_marked, input_node)
     end
     # the value of every random choice is in back_marked, since it affects its logpdf
-    push!(back_marked, node) 
+    push!(back_marked, node)
 end
 
 function back_pass!(back_marked, node::GenerativeFunctionCallNode)
@@ -124,7 +124,7 @@ function fwd_codegen!(stmts, fwd_marked, back_marked, node::JuliaNode)
     else
 
         # regular forward execution.
-    
+
         # we need the value for initializing gradient to zero (to get the type
         # and e.g. shape), and for reference by other nodes during
         # back_codegen! we could be more selective about which JuliaNodes need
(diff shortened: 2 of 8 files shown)


marcoct (commented):
Good find! A similar type of interpolation is used elsewhere in the static IR code, and there is an existing list of these quoted expressions, to make the code more concise than the `Expr(:(.), Gen, QuoteNode(..))` syntax: https://github.com/probcomp/Gen.jl/blob/master/src/static_ir/static_ir.jl#L77-L100. The difference is that you are using symbols here, whereas there the values are quoted. I'm not sure if there is actually a difference in behavior; I assume no.

Either way, for consistency it might be nice to reuse those variables or some other more concise way of writing these QuoteNode values. But, I'm not sure I love the `qn_...` naming I used in that list. Maybe something like `gen_quote_node(:static_get_subtrace)` or `gen_quote_node(static_get_subtrace)` would make the code here (and in the existing locations I mentioned above) more readable. Thoughts?

ztangent:
Oh cool, didn't realize that list of `qn_...` constants were there -- good to know! I actually there might be a slight difference in behavior between, e.g., `QuoteNode(static_get_index)` and `Expr(:(.), Gen, QuoteNode(:static_get_index))`. I think the first one is fine to use in a function body, because when unquoted it just refers to the actual function already defined in the `Gen` module, but when you're trying to define a new method for `Gen.static_get_index`, I suspect that the Julia parser requires the expression after the `function` keyword  to be a (potentially nested) Symbol, i.e., `Expr(:(.), Gen, QuoteNode(:static_get_index))` or `GlobalRef(Gen, :static_get_index)`. But I don't know for sure -- will need to test since this stuff isn't well documented.

ztangent:
@marcoct I implemented the `GlobalRef` approach to make things more consistent / concise, and got rid of all the `qn_...` values -- lmk if you this achieves what you were hoping!

marcoct:
Looks good to me :) Hooray for paying off technical debt!

PR #278, opened 2020-07-03
support macros in gen fns

### LINKED ISSUE #185: Handle macro calls in static DSL

georgematheos:
Resolves #185 .

@ztangent I know you have been working on the DSL expansion recently so if you have time to give feedback, that would be great.  I did not quite understand what you meant by "being careful with quoted values", so it is possible this implementation does not respect something it needs to!

### DIFF
--- a/src/Gen.jl
+++ b/src/Gen.jl
@@ -23,9 +23,7 @@ Permit use of generative functions written in the static modeling language up
 to this point. Functions are loaded into the calling module.
 """
 macro load_generated_functions()
-    for function_defn in generated_functions
-        Core.eval(__module__, function_defn)
-    end
+    :(load_generated_functions($(__module__)))
 end
 
 export load_generated_functions, @load_generated_functions
--- a/src/dsl/dsl.jl
+++ b/src/dsl/dsl.jl
@@ -5,7 +5,9 @@ const DSL_ARG_GRAD_ANNOTATION = :grad
 const DSL_RET_GRAD_ANNOTATION = :grad
 const DSL_TRACK_DIFFS_ANNOTATION = :diffs
 const DSL_NO_JULIA_CACHE_ANNOTATION = :nojuliacache
-const DSL_MACROS = Set([Symbol("@trace"), Symbol("@param")])
+
+# the set of macros we should not macroexpand before parsing
+const DSL_MACROS = Set([Symbol("@trace"), Symbol("@param"), Symbol("@gen")])
 
 struct Argument
     name::Symbol
@@ -72,7 +74,7 @@ function address_from_expression(lhs)
 end
 
 function desugar_tildes(expr)
-    trace_ref = GlobalRef(@__MODULE__, Symbol("@trace"))
+    trace_ref = Symbol("@trace")
     line_num = LineNumberNode(1, :none)
     MacroTools.postwalk(expr) do e
         # Replace with globally referenced macrocalls
@@ -89,22 +91,33 @@ function desugar_tildes(expr)
     end
 end
 
-function resolve_gen_macros(expr, __module__)
+# Resolves Gen macros to globally referenced macro calls;
+# fully macroexpands all macros Gen does not recognize
+function resolve_or_expand_macros(expr, __module__)
     MacroTools.postwalk(expr) do e
-        # Resolve Gen macros to globally referenced macrocalls
-        if (MacroTools.@capture(e, @namespace_.m_(args__)) &&
-            m in DSL_MACROS && __module__.eval(namespace) == @__MODULE__)
-            macro_ref = GlobalRef(@__MODULE__, m)
-            line_num = e.args[2]
-            Expr(:macrocall, macro_ref, line_num, args...)
-        elseif (MacroTools.@capture(e, @m_(args__)) &&
-                m in DSL_MACROS && isdefined(__module__, m) &&
-                getfield(__module__, m) == getfield(@__MODULE__, m))
-            macro_ref = GlobalRef(@__MODULE__, m)
-            line_num = e.args[2]
-            Expr(:macrocall, macro_ref, line_num, args...)
-        else
-            e
+        if MacroTools.@capture(e, @namespace_.m_(args__) | @m_(args__)) # for any macro expression...
+            # if this is already a GlobalRef, the macro's name is stored in `m.name`
+            macroname = m isa GlobalRef ? m.name : m
+
+            if macroname in DSL_MACROS # if this is a Gen macro name (ie. @trace or @param)
+                # The macro is either from the local module `__module__` or explicitly specifies a `namespace`
+                if namespace === nothing
+                    mod = __module__
+                else
+                    mod =__module__.eval(namespace)
+                end
+
+                # if the macro is defined in the given module, and refers to the Gen macro
+                if isdefined(mod, macroname) && getfield(mod, macroname) == getfield(@__MODULE__, macroname)
+                    macro_ref = GlobalRef(@__MODULE__, macroname)
+                    line_num = e.args[2]
+                    return Expr(:macrocall, macro_ref, line_num, args...)
+                end
+            end
+            # if we get here, then this is not a Gen macro, so macroexpand it
+            return macroexpand(__module__, e)
+        else # not a macro
+            return e
         end
     end
 end
@@ -139,8 +152,9 @@ function preprocess_body(expr, __module__)
     expr, quoted_exprs = extract_quoted_exprs(expr)
     # Desugar tilde calls to globally referenced @trace calls
     expr = desugar_tildes(expr)
-    # Also resolve Gen macros to GlobalRefs for consistent downstream parsing 
-    expr = resolve_gen_macros(expr, __module__)
+    # Resolve Gen macros to GlobalRefs for consistent downstream parsing,
+    # and expand macros which are not from Gen
+    expr = resolve_or_expand_macros(expr, __module__)
     # Reinsert quoted expressions after pre-processing
     expr = insert_quoted_exprs(expr, quoted_exprs)
     return expr
--- a/test/dynamic_dsl.jl
+++ b/test/dynamic_dsl.jl
@@ -529,4 +529,18 @@ end
 
 end
 
+@testset "macros in dynamic function" begin
+    @gen function foo()
+        x ~ exponential(1)
+        y ~ @insert_normal_call x
+    end
+    simulate(foo, ())
+
+    @gen function bar()
+        x ~ exponential(1)
+        y = Gen.@trace(@insert_normal_call(x), :y)
+    end
+    simulate(bar, ())
+end
+
 end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -64,6 +64,11 @@ end
 
 const dx = 1e-6
 
+# macro used in a couple tests
+macro insert_normal_call(x)
+    :(normal($(esc(x)), 0.0001))
+end
+
 include("autodiff.jl")
 include("diff.jl")
 include("selection.jl")
--- a/test/static_dsl.jl
+++ b/test/static_dsl.jl
@@ -591,4 +591,20 @@ ch = get_choices(tr)
 @test length(get_submaps_shallow(ch)) == 1
 end
 
+@testset "macros in static DSL" begin
+    @gen (static) function foo()
+        x ~ exponential(1)
+        y ~ @insert_normal_call x
+    end
+    @gen (static) function bar()
+        x ~ exponential(1)
+        y = Gen.@trace(@insert_normal_call(x), :y)
+    end
+    @load_generated_functions()
+
+    simulate(foo, ())
+    simulate(bar, ())
+end
+
+
 end # @testset "static DSL"
--- a/test/tilde_sugar.jl
+++ b/test/tilde_sugar.jl
@@ -4,7 +4,7 @@ import MacroTools
 @testset "tilde syntax" begin
 
 normalize(ex) =
-    MacroTools.prewalk(MacroTools.rmlines, Gen.resolve_gen_macros(ex, Main))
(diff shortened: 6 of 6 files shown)


src/dsl/dsl.jl
```diff
@@ -72,7 +74,7 @@ function address_from_expression(lhs)
 end
 
 function desugar_tildes(expr)
-    trace_ref = GlobalRef(@__MODULE__, Symbol("@trace"))
+    trace_ref = Symbol("@trace")
```
alex-lew, line 77: @georgematheos I believe @ztangent made this change recently to fix a problem where

1. you need to recognize `Gen.@trace`, not just `@trace`, and if someone did `const X = Gen`, you also need to recognize `X.@trace`

2. you don't want to recognize `@trace` when it's not in scope (i.e. if someone does `import Gen`, then `@trace` should not work -- they should need to do `Gen.@trace`).

alex-lew: I realize this is *inserting* `@trace` into the code, but what if the user has `@trace` referring to something else?

src/dsl/dsl.jl
```diff
@@ -89,22 +91,33 @@ function desugar_tildes(expr)
...
+
+                # if the macro is defined in the given module, and refers to the Gen macro
+                if isdefined(mod, macroname) && getfield(mod, macroname) == getfield(@__MODULE__, macroname)
+                    macro_ref = GlobalRef(@__MODULE__, macroname)
+                    line_num = e.args[2]
+                    return Expr(:macrocall, macro_ref, line_num, args...)
+                end
+            end
+            # if we get here, then this is not a Gen macro, so macroexpand it
+            return macroexpand(__module__, e)
```
alex-lew, line 118: Can this macroexpansion yield additional `@trace` calls that you need to process as above (in which case maybe an explicit recursive call is necessary)? Or will it try to expand them too?

src/Gen.jl
```diff
@@ -23,9 +23,7 @@ Permit use of generative functions written in the static modeling language up
 to this point. Functions are loaded into the calling module.
 """
 macro load_generated_functions()
-    for function_defn in generated_functions
-        Core.eval(__module__, function_defn)
-    end
+    :(load_generated_functions($(__module__)))
```
ztangent, line 26: Is there a reason this change was made? This implementation depends on `load_generated_functions` still existing as a function, but we were planning to deprecate it in #234. It also _might_ lead to errors if we do `import Gen`, because `load_generated_functions` isn't referenced.

src/dsl/dsl.jl
ztangent: Yes, we need to make sure that this is a `GlobalRef` not just `@trace`.

src/dsl/dsl.jl
```diff
@@ -5,7 +5,9 @@ const DSL_ARG_GRAD_ANNOTATION = :grad
 const DSL_RET_GRAD_ANNOTATION = :grad
 const DSL_TRACK_DIFFS_ANNOTATION = :diffs
 const DSL_NO_JULIA_CACHE_ANNOTATION = :nojuliacache
-const DSL_MACROS = Set([Symbol("@trace"), Symbol("@param")])
+
+# the set of macros we should not macroexpand before parsing
+const DSL_MACROS = Set([Symbol("@trace"), Symbol("@param"), Symbol("@gen")])
```
ztangent, line 10: I think we can omit `@gen` from this list, because the manual macro-expansion is only going to be applied to the function body, not the whole function.

src/dsl/dsl.jl
ztangent: @georgematheos I think I may not have been sufficiently clear in my suggestions yesterday about how best to implement this feature! I think we should avoid doing manual macro-expansion for the DML, because Julia can handle that on its own and thereby preserve all standard Julia semantics. It's not immediately clear to me that doing macro-expansion as part of a `postwalk` will actually preserve those semantics, and so I'd avoid that if possibld. I think it's only for the SML that we want to manually macro-expand the entire function body (after protecting all the Gen-specific macros by rewriting them). And so the macro-expansion code should go under `dsl/static.jl` rather than `dsl/dsl.jl`

ztangent:
Re-iterating a comment in my review above: I think I may not have been sufficiently clear in my suggestions yesterday about how best to implement this feature! I think we should avoid doing manual macro-expansion for the DML, because Julia can handle that on its own and thereby preserve all standard Julia semantics. It's not immediately clear to me that doing macro-expansion as part of a `postwalk` will actually preserve those semantics, and so I'd avoid that if possible.

I think it's only for the SML that we want to manually macro-expand the entire function body (after protecting all the Gen-specific macros by rewriting them). And so that macro-expansion code should go under dsl/static.jl rather than dsl/dsl.jl

It seems like there are two different issues here that seem good to me to address separately:

1. The first is that currently even the DML isn't handling macros within `@trace` calls. I'm not sure why this is the case. @georgematheos if you could open a separate issue on this, and provide a MWE with the error log, it'd be helpful to diagnose what's causing this -- I know you sent something previously that threw an error because of the LineNumberNode, but it's still unclear to me why that's causing Julia to error, since all `:macrocall` expressions are supposed to have LineNumberNodes within them.

2. The second is that the SML currently does not support macros, as noted in #185. This is because the SML parser does not macro-expand any of the code before building the DAG, and can be addressed separately from the first issue by taking the SML-specific implementation approach that I suggested above.

ztangent:
Regarding, protecting quoted expressions, I was referring to how `preprocess_body` extracts any expressions in quotes and replaces them with placeholders, so that any code re-writing that happens using `postwalk` doesn't affect those expressions:

https://github.com/probcomp/Gen.jl/blob/06168315768be163091e80feace27076a63ae575/src/dsl/dsl.jl#L137-L147

To support macros in the SML, we'll want to adopt a similar approach, so that Gen macros within quotes (e.g. `:(@trace(...))`) aren't rewritten, or added as nodes with the DAG representation, but instead stay as quoted expressions. To do that, we'll probably want to add a similar function in `static.jl` called `preprocess_static_body`.

src/dsl/dsl.jl
georgematheos: Okay, I'll switch this back.  I had thought it was slightly redundant since we wrap everything in `GlobalRef` in `resolve_or_expand_macros`/`resolve_gen_macros`, but I see what you mean, we need to be careful with namespaces.

src/Gen.jl
georgematheos: I was getting some error with `@load_generated_functions` that this fixed.  I think it was somehow caused by the fact that the old version of the macro did the loading at macro runtime rather than function runtime, but I don't really understand what the issue was.  We wouldn't need to keep `load_generated_functions` to keep this moved to runtime; we just need to make sure we return the original loop as a quoted expression.  Might be worth understanding better what the issue I was getting is.

georgematheos:
@ztangent okay, I will move this explicit macro expansion to the static DSL code, and update the `@trace` macro so it can handle the line number nodes.

Can you give me an example of when a value in quotes will appear in DSL code?
UPDATE -- I think I understand now the "quoted" case now (I saw Alex's original issue on this); no need to give an example :)

georgematheos:
Actually, I'm not 100% convinced we should only do explicit macro expansion for the static DSL if we are doing any transformation on the Gen macros before this.  The issue is, what if I have code like this:

```julia
macro add_traced_call(name::Symbol)
  :($name = @trace(normal(0, 1), $name))
end

@gen (static) function foo()
  @add_traced_call x
  return x
end
```

The issue with this is if we try to deal with `@trace` calls before we do the rest of the macroexpansion, we can't handle this case.  Maybe this isn't a super important case, but I can imagine times it would be useful.  If we wanted to support this, it seems to me that the way I have implemented it now is reasonable--except that we should change from a `postwalk` to a `prewalk`.

Handling situations like this is part of the reason that it seems reasonable to me to never try to capture macros in the parsing code, and instead just let everything macroexpand, then capture calls to `traceat`, etc.  It saves us from having to think too carefully about the different ways a `@trace` could appear, and instead just let Julia do it's full macroexpansion procedure.  For instance, even if we did what I just suggested, what if a naive user wrote a macro

```julia
macro insert_tilde()
  :(x ~ normal(0, 1))
end
```
The case I just described wouldn't handle that.  It seems to me that in a sense the most natural thing (in terms of providing the most natural behavior for the user) is to do as much macroexpansion as possible before anything else.  I think users will find it most clear if they can be confident anything their macros expand into get parsed as a regular part of the generative function.  To implement this, currently it seems reasonable to do a `prewalk` expanding all macros except special Gen macros, before any other parsing occurs, or reasonable to just call fully recursive `macroexpand` and then capture `traceat` calls later instead of `@trace`.  But I'm leaning against only changing the static DSL, since it means we can't handle some of these macro use cases, even in the dynamic DSL.  Any thoughts on this @alex-lew @ztangent , and if you agree with me, which of those two implementations you prefer?

ztangent:
> Handling situations like this is part of the reason that it seems reasonable to me to never try to capture macros in the parsing code, and instead just let everything macroexpand, then capture calls to traceat, etc. It saves us from having to think too carefully about the different ways a @trace could appear, and instead just let Julia do it's full macroexpansion procedure.

Interesting, I see. Yes, I had initially hoped that it would be trivial enough to just let Julia macro-expand everything and then run the SML parser on the code post-expansion. The issue with this is that the macro-expansions for `@trace` and `@param` are currently part of the implementation of the DML, which (i) might change in the future (ii) end up getting rid of some information that's relevant to the SML, in particular the type annotation for `@param` calls. If not for this, I'd be more than willing to take the approach of matching `traceat`. 

I think ultimately the best way to address this would be to get rid of `@trace` and `@param` as language constructs altogether, and instead implement the same functionality using other surface-level syntax (e.g. the tilde syntax), as @alex-lew suggested in the comments for #222. This way we would avoid having to treat certain macros as "special" in the SML, and we could readily adopt full macro-expansion of the function body, before detecting the surface-level syntax like tildes and whatever syntax replaces `@param`. But this would be a breaking change that we should discuss with @marcoct.

I'll have to think more about what's the best approach in the interim. One approach is to explicitly document that the SML can support macros, but will fail if you try to introduce macros that macro-expand to what are essentially "keywords" in the SML, i.e., `@trace` and `@param`. We could also document that macros like `insert_tilde` will not work as expected, because tilde de-sugaring occurs prior to any macro-expansion, and emphasize that that's just how the modeling languages work (for now). 

Another approach, which I'd considered while implementing #223, is to modify the implementation of `@param` such that it macro-expands to an expression that still contains the type annotation. We can then match on both what `@param` expands to and what `@trace` expands to (`traceat`), within the SML parser. This still feels messy to me, because it makes the SML dependent on an aspects of the implementation of the DML, but it might be the only way to get all the features you want in the interim. I've thought a bit about how to make it so that the SML and DML have separate implementations of `@trace` and `@param`, but haven't been able to think of a good way to make that work.

georgematheos:
I think adding type annotations information to the expansion of `@param` sounds reasonable.  I also think that it would probably be easy enough to manually macroexpand everything except `@trace` and `@param` before we do the rest of the parsing.  (I don't think `macroexpand` does anything fancy; I think it recursively expands all macros in a "prewalk" fashion.  We could code this manually, with a special rule to not expand Gen's special macros.)  I think after your comment @ztangent I'm leaning toward this avenue, since it minimizes changes to the SML code, but I think should handle the broad range of possible macro uses.  Does this sound reasonable to you?

Re removing all macro syntax--I'm of two minds about this.  On one hand, I think when we can come up with better syntax than macros (eg. the tildes), that's great.  On the other hand in Julia, macro syntax is very natural for expressing "unusual" operations, so I think it also feels sort of natural to use this syntax when we don't have in mind some better special-case syntax.  (Eg. currently `@param` seems pretty reasonable to me--though I'm not attached to this if we come up with something better.)  But in general I'm hesitant to restrict ourselves to never use macro syntax as a hard rule to make it easier to parse the DSLs.  I think probably parsing should be doable even with Gen using macros for syntax.

ztangent:
@georgematheos I gave this some additional thought this morning, and came up with another implementation strategy that should involve minimal special casing / leaves macro-expansion to Julia as much as possible! It'll effectively be a way of making it "so that the SML and DML have separate implementations of @trace and @param", as I suggested earlier.

The idea is to rewrite the `@trace` and `@param` macros to return a custom Gen-specific `Expr`, that a post-processing step in the `@gen` macro can then handle separately for the SML and DML:

```julia
macro trace(expr::Expr, addr)
    if expr.head != :call
        error("syntax error in @trace at $(expr)")
    end
    return Expr(:gentrace, expr, addr)
end

macro param(expr_or_symbol)
    local name::Symbol
    if isa(expr_or_symbol, Symbol)
        name, type = expr_or_symbol, Any
    elseif expr_or_symbol.head == :(::)
        name, type = expr_or_symbol.args
    else
        error("Syntax in error in @param at $(expr_or_symbol)")
    end
    return Expr(:genparam, name, type)
end
```

This way, full macro-expansion of the input function body will result in an AST including these special `Expr`s in them. Following that, the SML and DML parsers can each implement their own way of handling `:gentrace` and `:genparam` expressions. The SML parser will match these expressions when building the DAG, and the DML parser will simply rewrite these expressions into actually callable expressions, like `traceat`.

Happy to take a shot at implementing this in a new branch this afternoon -- I think it's a sufficiently different approach that starting afresh might be better!

georgematheos:
@ztangent I think that's a great solution!  If you have time to implement that, that would be great, but of course no worries if you're busy.  (If so, let me know and I will try to make time to implement it.)  And I agree starting fresh is better--I don't see any advantage to following off this PR's code.

georgematheos:
Closing since Xuan's PR above should resolve this issue!

PR #286, opened 2020-07-15
Bug fix in involution MCMC docs

georgematheos:
I may be misunderstanding this part--but I'm pretty sure in this warning, the issue is that the Jacobian could be zero, not that it could be nonzero!  @marcoct

### DIFF
--- a/docs/src/ref/mcmc.md
+++ b/docs/src/ref/mcmc.md
@@ -421,7 +421,7 @@ Also, the top-level function must take three arguments (`model_args`, `proposal_
 
 Some additional tips for defining valid involutions:
 
-- If you find yourself copying the same continuous source address to multiple locations, it probably means your involution is not valid (the Jacobian matrix will have rows that are identical, and so the Jacobian determinant will be nonzero).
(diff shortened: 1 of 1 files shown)


bzinberg:
nonzero -> zero LGTM.  I.m.o. the "and have zeros in every slot except one" makes it a little harder to understand, since the determinant would be zero even without that extra condition.

georgematheos:
Isn't it possible that the determinant can be nonzero without that additional condition?  Eg. we could have the Jacobian
```
1 1 0
1 1 0
0 0 1
```
which does not have nonzero determinant even though it has 2 identical rows.

bzinberg:
The determinant of any matrix that has two identical rows is zero.

georgematheos:
Right, not sure what I was talking about (or where that "example" came from).  Will change the docs to just say that the determinant would be zero.

bzinberg (approved):

marcoct:
Thanks George! I moved around a lot of the content on my working branch for the trace transform DSL changes, so I just added this change to my PR: https://github.com/probcomp/Gen.jl/pull/290

marcoct:
Closing this, since it was incorporated into https://github.com/probcomp/Gen.jl/pull/290

PR #287, opened 2020-07-15
tests for macros in DSLs

georgematheos:
These tests are NOT ALL PASSING.  It looks like there is a bug in parsing macros with expand into `x ~ call(...)` syntax, though I am not sure yet what is causing this.  (It is possible that there is a bug in my test cases.)


georgematheos:
Have updated this so the tests are passing; the issue causing the original problems is discussed in #280

PR #288, opened 2020-07-16
Add Cauchy distribution

sharanry:
This PR adds the Cauchy distribution. I have used the Normal distribution's implementation as a reference. Please let me know if I missed something. :)

PS. I had implemented it for a stochastic volatility model I was working on. I thought I will also submit it as a PR.


georgematheos (approved):
This looks good to me!

(As a note, if we wanted, we could probably optimize the implementation of `logpdf_grad` to do a bit less computation [eg. the `x0` derivative could just be set to the negative of the `x` derivative; we currently calculate `(x-x0)^2` in several places; etc.], but these are probably extremely minor optimizations.)

sharanry:
> This looks good to me!
> 
> (As a note, if we wanted, we could probably optimize the implementation of `logpdf_grad` to do a bit less computation [eg. the `x0` derivative could just be set to the negative of the `x` derivative; we currently calculate `(x-x0)^2` in several places; etc.], but these are probably extremely minor optimizations.)

Thanks for the review. Good point! I will make the changes.

marcoct:
Looks great @sharanry!

Issue #230, opened 2020-04-02
Improvements to involution DSL


marcoct:
Some discussion in the PR for the involution DSL (https://github.com/probcomp/Gen/pull/216) and TODO notes from the original PR, spur suggestion of several possible improvements and extensions. I am putting them all in one issue now for conciseness -- if someone wants to address one of these, they can create a specific issue for it and link to/from this issue.

- Automating detection of discrete / continuous, and simplification of the syntax, resulting in 4 read/write macros, instead of the current 8. The copy macros already do not distinguish between the two.

- Adding the ability to check whether an address exists or not. (But without reading its value)

- Currently, the proposal_retval is not allowed to depend on the  values of any continuous random choices in the model or proposal trace, because otherwise continuous choice data leaks into the involution and avoids accounting by the Jacobian correction. Perhaps we could allow the dependency, while still ensuring the JAcobian is computed correctly, by using a call to `choice_gradients` on the proposal generative function?

- We could add more checks that we don't write to, or copy to, the same address twice (these can be only enabled in 'check' mode). Also, because it creates noninvertible Jacobian, it is invalid to copy from the same address twice and we could check for that to

- Users should be able to read and write multivariate random choices (i.e. array-valued) in the involution DSL. The implementation would probably need to flatten them alongside the scalar dimensions for input/output to f_array.

- Add the ability to pass two optional address namespaces to `@callinv`(one for the proposal, one for the model). Any addresses used by reads, writes or copies will occur under that address namespace.

- In the LaTex doc, explain relationship and differences from http://proceedings.mlr.press/v89/roberts19a.html

bzinberg:
> Automating detection of discrete / continuous

I have strong reservations about this.  For example, a categorical whose values are specific floats seems like something that we do not want to forbid.  It could occur in e.g. a hand-implementation of a piecewise-uniform distribution.  I think this is similar to @marcoct's concern in https://github.com/probcomp/Gen/pull/216#issuecomment-608007854.

alex-lew:
Yeah, I think ultimately the right way to handle that is to make more of this kind of information accessible in the trace -- though I think a feature like that would require careful design so that we don't compromise performance or place too high a burden on custom GFs.

marcoct:
[...]

PR #290, opened 2020-07-20
Trace Translators and new Trace Transform DSL

marcoct:
This PR makes several changes related to Gen's support for involution MCMC and SMC:

- What was previously called the "involution DSL" has now been generalized to become a "Trace Transform DSL", for transforming between arbitrary spaces of traces or pairs of traces. Whereas the involution DSL was previously restricted to the setting when the previous model and the new model were the same generative function, this DSL also can be used to define transformations when the previous model and the new model are different generative functions.

- There are many syntax improvements to the Trace Transform DSL, which is now documented [here](https://github.com/probcomp/Gen.jl/blob/20200416-marcoct-translatordsl/docs/src/ref/trace_translators.md#trace-transform-dsl).

- The Trace Transform DSL can now read and write to multivariate (vector-valued) continuous random choices, and compute Jacobians appropriately.

- There is a new abstraction called "Trace Translators", that unifies patterns of computation that occur within involutive MCMC with patterns of computation occuring within SMC algorithms. This PR includes [documentation for several types of Trace Translators](https://github.com/probcomp/Gen.jl/blob/20200416-marcoct-translatordsl/docs/src/ref/trace_translators.md), including General Trace Translators (the most general type, which can be used in constructing SMC samplers that use arbitrary sequences of models), Symmetric Trace Translators (used in involutive MCMC), Deterministic Trace Translators (primarily for pedagogical purposes), and Simple Extending Trace Translators (corresponding to a variant of SMC where the model is extended with new random choices at each step, that are sampled from a proposal distribution).

Some of these item were part of this issue: https://github.com/probcomp/Gen.jl/issues/230


src/inference/trace_translators.jl
```diff
@@ -0,0 +1,874 @@
...
+    # apply trace transform
+    (new_model_trace, backward_proposal_trace, log_abs_determinant) = symmetric_trace_translator_run_transform(
+        translator.f, prev_model_trace, forward_proposal_trace, translator.q, translator.q_args)
+
+    # compute log weight
+    prev_model_score = get_score(prev_model_trace)
+    new_model_score = get_score(new_model_trace)
+    forward_proposal_score = get_score(forward_proposal_trace)
+    backward_proposal_score = get_score(backward_proposal_trace)
+    log_weight = new_model_score - prev_model_score + backward_proposal_score + forward_proposal_score + log_abs_determinant
```
georgematheos: At risk of asking a dumb question--if I'm doing my math correctly, wouldn't we want the `forward_proposal_score` term to be negative?  Ie. `- forward_proposal_score` instead of `+ forward_proposal_score`?

georgematheos: Since we want the acceptance ratio to be `(P[new_tr] * P[new_tr -> old_tr]) / (P[old_tr] * P[old_tr -> new_tr])`, where the `P[old_tr -> new_tr]` appears in the denominator.

src/inference/mh.jl
```diff
@@ -69,45 +70,26 @@ Perform a generalized (reversible jump) Metropolis-Hastings update based on an i
...
+When continuous random choices are used, the `weight` returned by the involution must include an additive correction term that is the determinant of the the Jacobian of the bijection on the continuous random choices that is obtained by currying the involution on the discrete random choices (this correction term is automatically computed if the involution is constructed using the [Bijection DSL](@ref)).
 NOTE: The Jacobian matrix of the bijection on the continuous random choices must be full-rank (i.e. nonzero determinant).
 The `check` keyword argument to the involution can be used to enable or disable any dynamic correctness checks that the involution performs; for successful executions, `check` does not alter the return value.
 """
 function metropolis_hastings(
         trace, proposal::GenerativeFunction,
-        proposal_args::Tuple, involution;
+        proposal_args::Tuple, involution::TraceTransformDSLProgram;
```
georgematheos: Does this mean we've gotten rid of any support for involution MH which does not use the trace transform DSL?

This isn't a huge deal (it's pretty easy to implement the old version of involution MH manually if needed), but there may be cases where for performance reasons users would rather manually write an involution (and manually specify the Jacobian calculation) rather than using the DSL.  As a result, I wanted to comment that it might be worth maintaining support for involutions specified as Julia functions.

georgematheos:
One thing I just ran into while using this DSL was that I was getting a cryptic error about how the "Jacobian is 0x1 and is not square".  I eventually realized this is because I was looking up a continuous return value of a subtrace in my model in the involution I had written.  It was something along the lines of:
```julia
# `:vals => i` points to a subtrace which samples `:val ~ ...` then returns this value
@read(model_tr[:vals => 1], :cont)
```

Would it be easy to throw a custom error if a user tries to read a continuous value from a subtrace return like this, so it's clear to the user how to fix this?

src/inference/trace_translators.jl
marcoct: Yes, that's a bug. Thanks!

src/inference/mh.jl
marcoct: Yes. I realized this would remove support for code based on manual implementation of involutions. Maybe the best way of adding that support is allowing users to construct a SymmetricTraceTranslator using a custom involution not produced using the Trace Transform DSL. The user would be responsible for providing a function that implements something like what is now called `symmetric_trace_translator_run_transform`.

georgematheos: @marcoct I think that proposal makes sense.  Another possibility would be to have 1 version of `mh` which accepts a `SymmetricTraceTranslator`, and 1 version which accepts a tuple `(proposal, proposal_args, involution::Function)`.  (This is what I have been using in a fork of this for my open universe modeling.)  This second method could be used with manual involutions.  To retain the current syntax in this PR, we could have another method accepting a `(proposal, proposal_args, involution::TraceTranslatorDSLProgram)` which gets converted to the `SymmetricTraceTranslator` method internally.

ztangent: I also think it makes sense to have a version where manual involutions can be specified! It'd help with both potential performance improvements, and also backward compatibility with older involution MCMC code.

src/inference/trace_translators.jl
```diff
@@ -0,0 +1,874 @@
...
+
+    (output_trace, log_weight) = translator(input_trace)
+"""
+@with_kw struct SimpleExtendingTraceTranslator 
+    p_new_args::Tuple = ()
+    argdiffs::Tuple = ()
+    new_obs::ChoiceMap = EmptyChoiceMap()
+    q_fwd::GenerativeFunction
+    q_fwd_args::Tuple  = ()
+end
```
ztangent, line 781: I wonder if it might be possible to also support a "less simple" `ExtendingTraceTranslator", which allows users to provide a `TraceTransformDSLProgam" to map choices generated by `q_fwd` to choices within `p_new`? This might arise in cases where the proposal doesn't sample choices that also exist within the extended model. For example, in a time series model, where the models trace addresses are nested within a namespace like `1` within an `Unfold` combinator, it might be nice to write a proposal which doesn't also have to nest its choices within that namespace, and instead specify a trace transform to do the appropriate mapping.

ztangent (commented):
Love this new abstraction, as well as the new syntax! I can totally see how this helps provide a unifying formalism for various operations used in both SMC and MCMC, like we discussed previously.

One limitation of the Trace Transform DSL though, as compared to the Involution DSL, is it that it seems like `TraceTransformDSLProgram`s no longer have access to the model and proposal arguments, whereas `InvolutionDSLPrograms` did? If I'm right about that, then it seems harder to write trace transforms parameterized by various model arguments, which might be useful sometimes (e.g. in the example I gave for the use of `ExtendingTraceTranslators` in SMC), because the trace address you want to read / write from could depend on model arguments, and not just previously sampled values. Would be curious to hear more of your thoughts about this @marcoct!

marcoct:
> One limitation of the Trace Transform DSL though, as compared to the Involution DSL, is it that it seems like `TraceTransformDSLProgram`s no longer have access to the model and proposal arguments, whereas `InvolutionDSLPrograms` did? If I'm right about that, then it seems harder to write trace transforms parameterized by various model arguments, which might be useful sometimes (e.g. in the example I gave for the use of `ExtendingTraceTranslators` in SMC), because the trace address you want to read / write from could depend on model arguments, and not just previously sampled values.

@ztangent That is still supported, but with a different syntax that better matches the idea that `model_in` and `aux_in` are traces, e.g. `get_args(model_in)` and `get_args(aux_in)`.

ztangent:
> That is still supported, but with a different syntax that better matches the idea that model_in and aux_in are traces, e.g. get_args(model_in) and get_args(aux_in).

Nice, that's right! Perhaps all that's needed then is an example within the docs / tutorials of that -- or maybe there already is, and I just missed it.

src/inference/trace_translators.jl
ztangent: This will also be really helpful if `q_fwd` is itself a `Map` or `Unfold` combinator, in which case a trace transform would be required to remap all the indices from `1, ..., t` to `T+1, ..., T+t`.

georgematheos:
As another suggestion for an improvement, I think the documentation for the `@copy` and `@tcall` could both be improved.  In particular, for `@copy`, we should make it clear that
1. It is possible to copy a whole submap of addresses
2. If we copy a whole submap of addresses, it will overwrite any values currently at those addresses

I'm not 100% sure 2 is true--it is true in my fork where I redid the `set_submap!` function in my choicemap rewrite, but not sure about the main branch.  If 2 is true, we could also consider if this is the behavior we want.  I suspect it isn't.  Eg. It is sometimes nice to have an involution looking like
```julia
@transform .....
  @write(new_tr[:top => :value => 1], some_value, :disc)
  # say the proposal proposes some addresses in `:top => :space1` and `:top => :space2`, but not in `:top => :value`
  @copy(proposal_tr[:top], new_tr[:top])
end
```
Ie. we may want to write to some addresses in namespace `:top`, then copy in all the proposed addresses in namespaces `:top` without overwriting the values we wrote explicitly.

To prevent this type of overwriting should be pretty easy, we just need to change a call to `set_submap!` to `merge!`.

src/inference/mh.jl
marcoct: I added back for support for involutions defined as Julia functions. It should be backwards-compatible now.

src/inference/trace_translators.jl
marcoct: Yes, that sounds useful. I think it's separable from this PR, so let's make it an issue.

marcoct: @ztangent Here is an issue to track this: https://github.com/probcomp/Gen.jl/issues/313

PR #293, opened 2020-07-26
Fixes precompilation warnings due to overriding of functions in `ReverseDiff.jl`

### LINKED ISSUE #292: Incremental compilation warning when precompiling Gen
These warnings have been occurring during precompilation of `Gen` (or packages that use `Gen`), suggesting that incremental compilation may be broken:

```
WARNING: Method definition special_reverse_exec!(ReverseDiff.SpecialInstruction{typeof(Base.fill), I, O, C} where C where O where I) in module ReverseDiff at [...]\.julia\packages\ReverseDiff\tCOFA\src\macros.jl:213 overwritten in module Gen at [...]\.julia\packages\Gen\5JiNL\src\backprop.jl:35.

WARNING: Method definition special_forward_exec!(ReverseDiff.SpecialInstruction{typeof(Base.fill), I, O, C} where C where O where I) in module ReverseDiff at [...]\.julia\packages\ReverseDiff\tCOFA\src\macros.jl:229 overwritten in module Gen at [...]\.julia\packages\Gen\5JiNL\src\backprop.jl:44.
```

According to @marcoct this looks like it's because "someone added support for `fill` to ReverseDiff, and that Gen had this built-in for a while." The fix should be just to remove those definitions of `fill` from Gen.

ztangent:
Fixes #292 by removing the offending definitions of `Base.fill`, `ReverseDiff.special_forward_exec!`, and `ReverseDiff.special_reverse_exec!`, as suggested by @marcoct. These methods were introduced into `ReverseDiff.jl` v1.4, which lead to warnings because the Gen definitions were overriding the ReverseDiff ones. This PR also bumps the version compatibility of `ReverseDiff` to 1.4, since we now depend on their implementation of `fill`.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -17,9 +17,9 @@ uuid = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
 
 [[CommonSubexpressions]]
 deps = ["Test"]
-git-tree-sha1 = "efdaf19ab11c7889334ca247ff4c9f7c322817b0"
+git-tree-sha1 = "34aa50efad19a788db0cb2eb44d149942f64816a"
 uuid = "bbf7d656-a473-5ed7-a52c-81e309532950"
-version = "0.2.0"
+version = "0.2.1"
 
 [[CompilerSupportLibraries_jll]]
 deps = ["Libdl", "Pkg"]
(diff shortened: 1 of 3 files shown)


marcoct:
I tested one of the examples that exercises AD on `fill` from `examples/`, and it works.

PR #298, opened 2020-08-10
Unused type parameter in `hmc`

mschauer:


mschauer:
One could also get rid of the `U` but perhaps the intention was to constrain it to traces?

marcoct:
Thanks @mschauer ! Yes, that field should be `trace::Trace` and there should be no type parameters.

src/inference/hmc.jl
```diff
@@ -24,7 +24,7 @@ Neal, Radford M. (2011), "MCMC Using Hamiltonian Dynamics", Handbook of Markov C
 """
 function hmc(
         trace::U, selection::Selection; L=10, eps=0.1,
```
mschauer:
```suggestion
        trace::Trace, selection::Selection; L=10, eps=0.1,
```

src/inference/hmc.jl
```diff
@@ -24,7 +24,7 @@ Neal, Radford M. (2011), "MCMC Using Hamiltonian Dynamics", Handbook of Markov C
...
-        check=false, observations=EmptyChoiceMap()) where {T,U}
+        check=false, observations=EmptyChoiceMap()) where {U}
```
mschauer:
```suggestion
        check=false, observations=EmptyChoiceMap())
```

mschauer (commented):
Suggestions from review

Issue #311, opened 2020-09-18
Gen on `master` fails to precompile when inside Docker on a Mac


bzinberg:
When trying to run GenSceneGraphs from Docker, @jcrosenb encountered the error
```
┌ Info: Precompiling GenSceneGraphs [dcc2c3cc-8ed1-11e9-1eb3-1dd41a11fee5]
└ @ Base loading.jl:1278
ERROR: LoadError: Failed to precompile Gen [ea4f424c-a589-11e8-07c0-fd5c91b9da4a] to /root/.julia/compiled/v1.5/Gen/OEZG1_Y796z.ji.
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] compilecache(::Base.PkgId, ::String) at ./loading.jl:1305
 [3] _require(::Base.PkgId) at ./loading.jl:1030
 [4] require(::Base.PkgId) at ./loading.jl:928
 [5] require(::Module, ::Symbol) at ./loading.jl:923
 [6] include(::Function, ::Module, ::String) at ./Base.jl:380
 [7] include(::Module, ::String) at ./Base.jl:368
 [8] top-level scope at none:2
 [9] eval at ./boot.jl:331 [inlined]
 [10] eval(::Expr) at ./client.jl:467
 [11] top-level scope at ./none:3
in expression starting at /julia_projects/GenSceneGraphs/src/GenSceneGraphs.jl:12
Failed to precompile GenSceneGraphs [dcc2c3cc-8ed1-11e9-1eb3-1dd41a11fee5] to /root/.julia/compiled/v1.5/GenSceneGraphs/N6XWy_Y796z.ji.
```

This persisted even after deleting the directory `~/.julia/compiled`.

**Possible next step A:**
Build a Docker image containing Gen at a specific recent commit (e.g. current head, 07a542734dc27e7edea3bb8a1fc5aaf37cd565e1) and see whether the same precompile error occurs on @jcrosenb's machine.  If it does, we should get some specs of her machine and then see if it reproduces on a different Mac.

**Possible next step B:**
Test out https://github.com/probcomp/GenSceneGraphs.jl/pull/205 on a different Mac.  If we can find someone who has a Mac and is able to do this, that would be the easiest path to finding out whether this is really a Mac-wide issue or is specific to some aspect of @jcrosenb's setup.  (For the time being, she's attempting to get the Docker image running from within a Linux VM).

bzinberg:
[...]

PR #312, opened 2020-09-18
[DO NOT MERGE] Preliminary Ubuntu 20.04 images for both CPU and GPU as well as docs

fplk:
- Current image in master will not build at all due to `wget https://julialang.org/juliareleases.asc`. That will not resolve, since this file was moved to https://julialang.org/assets/juliareleases.asc The PRed images work.
- Currently, there are no container instructions - neither in the Dockerfile nor in the Github readme nor in the [Gen docs](https://www.gen.dev/dev/search/?q=docker), so I have added docs to build and run the image
- Images for CPU **and** GPU
- Updates Julia 0.7 (from 2018) to 1.5.1 (current)
- Updates Ubuntu 16.04 to 20.04, since 16.04 is in ESM - cmp. [release cycle](https://ubuntu.com/about/release-cycle) whereas 20.04 received full hardware & maintenance updates until 2022, LTS until 4/2025 and ESM until 04/2030
- Tweaks: Updates MAINTAINER tag which is deprecated to LABEL, removes cached files and uses `--no-install-recommends` for smaller image size, reduces number of layers caused by separate RUN lines in current image
- Forms foundation to derive other Gen project images from (I will soon push POCs for 1 external and 2 internal projects)

We might consider:
- Adding JupyterLab (which I already have in internal image)
- Consolidating this with the [gen-quickstart image](https://github.com/probcomp/gen-quickstart/blob/master/Dockerfile.ubuntu-2004) (which I have also contributed)
- First make sure that internal projects can successfully base their containers on these before merging
- Further investigate [issue 311](https://github.com/probcomp/Gen.jl/issues/311) before merging, since this currently blocks using Gen in containers on macOS
- Could add checksum verification back in like [the previous image did](https://github.com/probcomp/Gen.jl/blob/master/docker/ubuntu1604#L13)

Happy to accommodate change requests - looking forward to feedback.


docker/GenCPU.dockerfile
```diff
@@ -0,0 +1,40 @@
+FROM ubuntu:20.04
+
+LABEL maintainer="MIT Probabilistic Computing Project"
+
+# Find current Julia version on https://julialang.org/downloads/
+ARG JULIA_VERSION_SHORT="1.5"
+ARG JULIA_VERSION_FULL="${JULIA_VERSION_SHORT}.1"
+ENV JULIA_INSTALLATION_PATH=/opt/julia
```
fplk, line 8: Duplicate - see line 11.

docker/GenGPU.dockerfile
```diff
@@ -0,0 +1,40 @@
+FROM nvidia/opengl:1.0-glvnd-runtime-ubuntu20.04
+
+LABEL maintainer="MIT Probabilistic Computing Project"
+
+# Find current Julia version on https://julialang.org/downloads/
+ARG JULIA_VERSION_SHORT="1.5"
+ARG JULIA_VERSION_FULL="${JULIA_VERSION_SHORT}.1"
+ENV JULIA_INSTALLATION_PATH=/opt/julia
```
fplk, line 8: Duplicate - see line 11.

docker/Readme.md
```diff
@@ -0,0 +1,41 @@
...
+ERROR: Failed to precompile Gen [ea4f424c-a589-11e8-07c0-fd5c91b9da4a] to /root/.julia/compiled/v1.4/Gen/OEZG1_t5nDi.ji.
+Stacktrace:
+ [1] error(::String) at ./error.jl:33
+ [2] compilecache(::Base.PkgId, ::String) at ./loading.jl:1272
+ [3] _require(::Base.PkgId) at ./loading.jl:1029
+ [4] require(::Base.PkgId) at ./loading.jl:927
+ [5] require(::Module, ::Symbol) at ./loading.jl:922
+ ```
+
+ This can even happen when running inside a VM on macOS. While this bug is confirmed and reproducable, there is currently no solution for it, cmp. [issue 311](https://github.com/probcomp/Gen.jl/issues/311).
```
fplk, line 41: Typo - should be "reproducible"

docker/Readme.md
```diff
@@ -0,0 +1,41 @@
...
+
+### Docker Setup
+Please follow the [official Docker installation instructions](https://docs.docker.com/engine/install/) to setup Docker on your target platform. If you would like to use GPU acceleration, please also install the [NVIDIA Container Toolkit / nvidia-docker](https://github.com/NVIDIA/nvidia-docker).
+
+
+### Building Container
+You can build Docker containers for Gen that either target CPU-only or GPU-enabled machines.
+- Please run `docker build -f docker/GenCPU.dockerfile -t gen:cpu-ubuntu20.04 .` for the CPU image
+- Please run `docker build -f docker/GenGPU.dockerfile -t gen:gpu-ubuntu20.04 .` if you want to use GPU acceleration
+Both commands are supposed to be run from the main Gen.jl folder, not this subfolder.
```
bzinberg, line 13: Perhaps we should have the "easy version" be to pull a pre-built image from Docker Hub, and have building the container on one's own be the "advanced version"?

docker/Readme.md
```diff
@@ -0,0 +1,41 @@
+# Gen Containerization
```
bzinberg, line 1: Perhaps rather than in its own markdown file, this should be added as a subsection at the bottom of `docs/src/getting_started.md`.  I say bottom because I think Gen wants to present primarily as an ordinary Julia package rather than a monolithic stack (perhaps @marcoct can comment).  Also, we should include in the beginning of this section an explanation of why the user might ever want to use these container instructions (e.g., for portability / saves the need to install anything other than Docker).

bzinberg (commented):
It seems to me like duplicating the Dockerfiles to support both CPU and GPU is currently overkill, since none of Gen's native operations use GPU and (afaik) no one has ever asked us for help with GPU-related operations.  I guess it's ok for now, since it's not that much more work to have the second image.

docker/Readme.md
fplk: Definitely, I just lack access rights to do that for you. But I would highly recommend that as well. `docker run -it probcomp/gen:cpu-ubuntu20.04 bash` should just work, absolutely.

Issue #255, opened 2020-05-07
Dispatch combinator to handle multiple dispatch in generative functions


femtomc:
As per the `Gen` slack - one thing which may be slightly confusing to people who are initially using `Gen` is that generative functions are structs. This implies that programmers must avoid normal Julia multiple dispatch idioms. 

This is exemplified by the following code chunk which "silently" fails because types are not checked in `generate` and the generative function definition overwrites the previous. Thus, the support is completely different than what might be expected, likely unbeknownst to the programmer.
```
module MWEDispatchIssues

using Gen

@gen function foo(y::Int)
    if y < 5
        x = @trace(normal(0.0, 1.0), :x)
    else
        x = @trace(categorical([0.5, 0.5]), :x)
    end
    return x
end

@gen function foo(y::Float64)
    if y > 10
        x = @trace(normal(0.0, 10.0), :x)
        z = @trace(normal(0.0, 10.0), :z)
    else
        x = @trace(normal(0.0, 1.0), :x)
        z = @trace(normal(0.0, 1.0), :z)
    end
    return x, y
end

tr, _ = generate(foo, (5,))
display(get_choices(tr))

end
```
To solve this issues, there have been many suggestions on the Slack, which I am posting here below.

From George Matheos:
> **Alex Lew** I agree that the conceptually neatest way to do the dispatch is as a combinator! This might be the right route implementation-wise too, since it’s easy to understand, though I’m wondering if we might actually just be able to change the implementation of the static and dynamic DSLs slightly to implicitly implement this sort of combinator. I think what keorn is saying about relying on regular Julia method dispatch could probably work pretty well. I haven’t ironed out all the details, but I’m thinking: For the static DSL, the gfi methods are all custom generated methods, meaning we have a separate method for each concrete argument type. We could augment the trace type of the static DSL to include the type of the arguments, and then declare a separate generated GFI method for each of the declared static DSL methods, so that depending on the trace/argument types, julia will automatically use the code for the proper user-declared method.  
[...]

PR #314, opened 2020-09-25
Support AD-safe type checking in dynamic @gen functions

ztangent:
This PR ensures that type annotations in dynamic `@gen` function definitions get propagated to the underlying `julia_function`. This should help prevent unexpected errors or behavior due to `@gen` functions accepting the wrong type. For example, instead of the behavior documented in #255 where re-defined `@gen` functions continued to work on the old types, they will throw a method error:

```
julia> @gen foo(x::Int) = y ~ uniform_discrete(1, x);
julia> foo(5)
3
julia> @gen foo(x::Float64) = y ~ normal(x, 1);
julia> foo(5)
ERROR: MethodError: no method matching ##foo#449(::Gen.GFProposeState, ::Int64)
julia> foo(5.0)
5.632113778420875
```

If we ever decide to support multiple-dispatch as requested in #255, this PR also helps lay the groundwork for that, by ensuring the inner Julia methods do not automatically accept every argument type.

I previously tried to implement this feature when adding support for optional arguments in #195, but ran into AD errors due to the [restrictions on ReverseDiff](http://www.juliadiff.org/ReverseDiff.jl/limits/). This implementation is AD-safe , because it automatically promotes the types of `(grad)`-annotated arguments to `Real` or `AbstractArray{<:Real}`.

Note that loose type-checking [i.e., type conversion via `convert`] is already performed for static `@gen` functions during construction of the `StaticTrace`.

### DIFF
--- a/src/dsl/dsl.jl
+++ b/src/dsl/dsl.jl
@@ -61,6 +61,22 @@ function parse_arg(expr)
     arg
 end
 
+function resolve_grad_arg(arg, __module__)
+    # Ensure that differentiable arguments are supported by ReverseDiff
+    if !(DSL_ARG_GRAD_ANNOTATION in arg.annotations) return arg end
+    typ = Core.eval(__module__, arg.typ)
+    if typ <: Real
+        new_typ = :Real
+    elseif typ <: AbstractArray{<:Real} && IndexStyle(typ) == IndexLinear()
+        new_typ = :(AbstractArray{<:Real})
+    elseif Real <: typ || AbstractArray{<:Real} <: typ
+        new_typ = arg.typ
+    else
+        error("Type of $(arg.name)::$(arg.typ) does not support differentiation.")
+    end
+    return Argument(arg.name, new_typ, arg.annotations, arg.default)
+end
+
 include("dynamic.jl")
 include("static.jl")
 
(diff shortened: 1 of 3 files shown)


marcoct:
Nice! @ztangent Would you mind checking that the following examples still run in `examples/run_examples.jl` after this change? Some of these exercise AD through generative function arguments.

#include("decode/run.jl")
include("regression/static_mala_hmc.jl")
include("regression/static_map_optimize.jl")
include("regression/static_mh.jl")
include("regression/static_mh_collapsed.jl")
include("regression/quickstart.jl")
include("regression/dynamic_map_optimize_gibbs.jl")
include("regression/dynamic_mh.jl")
include("regression/dynamic_resimulation_mh.jl")
#include("coal/coal.jl")
#include("gp_structure/lightweight.jl")
#include("gp_structure/incremental.jl")
include("mle/static_mle.jl")

We probably should add some of these examples to the automatically run test suite. I didn't add the examples to the test suite because some of the examples depend on a few other packages (CSV, plotting, etc.) that I don't want to introduce as dependencies of Gen.

But for now, when there are changes I think have a high probability impacting some feature, I tend to run the examples.

ztangent:
Can do!

Re: running those examples as tests without increasing Gen's dependencies, I think we might want adopt the use of [test-specific dependencies](https://julialang.github.io/Pkg.jl/v1/creating-packages/#Test-specific-dependencies-in-Julia-1.2-and-above-1), which are supported in Julia 1.2 onward. Pretty sure this means that they won't end up being installed as dependencies via `add Gen`!

marcoct:
> Re: running those examples as tests without increasing Gen's dependencies, I think we might want adopt the use of [test-specific dependencies](https://julialang.github.io/Pkg.jl/v1/creating-packages/#Test-specific-dependencies-in-Julia-1.2-and-above-1), which are supported in Julia 1.2 onward. Pretty sure this means that they won't end up being installed as dependencies via `add Gen`!

Good idea. But I still think its good to be able to run the tests without those dependencies. An option might be to move the examples over to gen-quickstart and have a separate travis job that runs them there on the head of Gen master branch whenever it is updated.

ztangent:
@marcoct Okay I've run the examples that make use of (grad) annotated arguments -- all seem to be working!

ztangent:
@marcoct should I go ahead and merge?

marcoct:
Sounds good!

marcoct (approved):

PR #315, opened 2020-09-25
Fixed small naming issue with the selection test sets.

ztangent:
For whatever reason, the test sets for `Selection` functionality weren't named properly (the test-set names were accidentally placed after the `begin` keyword, instead of before). This is a small PR to address that.


ztangent:
@bzinberg lmk if I can just merge this since it doesn't change any functionality.

bzinberg (approved):
oh yeah this is clearly a correction

Issue #317, opened 2020-09-28
Cannot return `nothing` from static modeling language


marcoct:
Minimal example:
```julia
using Gen

@gen (static) function foo()
    return nothing
end
```

Results in error:
```
ERROR: LoadError: LoadError: Tried to return nothing, which is not a locally bound variable
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] parse_return_line!(::Array{Expr,1}, ::Dict{Symbol,Symbol}, ::Symbol) at /home/marcoct/.julia/packages/Gen/wW2ee/src/dsl/static.jl:190
 [3] parse_static_dsl_line!(::Array{Expr,1}, ::Dict{Symbol,Symbol}, ::Expr) at /home/marcoct/.julia/packages/Gen/wW2ee/src/dsl/static.jl:233
 [4] parse_static_dsl_function_body!(::Array{Expr,1}, ::Dict{Symbol,Symbol}, ::Expr) at /home/marcoct/.julia/packages/Gen/wW2ee/src/dsl/static.jl:250
 [5] make_static_gen_function(::Symbol, ::Array{Gen.Argument,1}, ::Expr, ::Symbol, ::Set{Symbol}) at /home/marcoct/.julia/packages/Gen/wW2ee/src/dsl/static.jl:275
 [6] parse_gen_function(::Expr, ::Set{Symbol}, ::Module) at /home/marcoct/.julia/packages/Gen/wW2ee/src/dsl/dsl.jl:146
 [7] @gen(::LineNumberNode, ::Module, ::Any, ::Expr) at /home/marcoct/.julia/packages/Gen/wW2ee/src/dsl/dsl.jl:156
 [8] include(::Function, ::Module, ::String) at ./Base.jl:380
 [9] include(::Module, ::String) at ./Base.jl:368
 [10] exec_options(::Base.JLOptions) at ./client.jl:296
 [11] _start() at ./client.jl:506
```

Using Julia 1.5.1, current master branch of Gen (https://github.com/probcomp/Gen.jl/tree/93f6fc2a66295495034e692ac8095dc182c54159)

ztangent:
Interesting, looks like the issue is that there's an ambiguity as to whether `nothing` is a constant vs a local variable, because at parse  time, all we can see is that it's a symbol. It's possible that similar errors might arise with `missing`, `true` or `false`, though I'm not sure about the latter two -- depends on Julia's parser.

[...]

PR #318, opened 2020-09-28
20200927 marcoct fixsml

marcoct:
Fix for https://github.com/probcomp/Gen.jl/issues/317


ztangent (approved):
Ooh smart -- I guess this adds an extra node to the DAG, but seems worth it for the conciseness / generality!

Issue #244, opened 2020-04-20
Implement `logpdf` using Distributions.jl when possible


marcoct:
`logpdf` for a few distributions is implemented from scratch in Gen, and should use Distributions.jl implementations.

keorn:
On a related note, what is the rationale behind not using Distributions.jl? That would make Gen a bit more interoperable in the ecosystem and allow for using some convenient interface methods they have.

marcoct:
Good question.

There are tradeoffs involved in making a decision to integrate tightly with an existing library by using the types of the library directly as part of Gen's interfaces (e.g. inheritance) instead of simply calling the existing library (composition). The benefit is the potential for less glue code, but the downsides is that it makes Gen less easy to evolve over time. This is related to https://en.wikipedia.org/wiki/Composition_over_inheritance. Gen started out as a risky and exploratory research project, and we wanted the ability to keep things flexible and not couple too tightly with existing libraries.

Note that we are definitely using Distribution.jl code, but just not the the types at the moment. In principle users of Gen should not need to interact with Distributions themselves, so the only methods of a Distribution that are needed are the ones that Gen already uses. @alex-lew has a one-liner that wraps a Distributions.jl Distribution in a Gen Distribution, using https://github.com/TuringLang/DistributionsAD.jl for gradients, which fits with this "composition not inheritance" approach.

But I think it is worth discussing this more. Have you found a need for particular additional interface methods provided by Distributions.jl when using Gen?

keorn:
So there are a few considerations:
1. Distributions.jl implements a whole slew of statistics methods that can be used when needed like `insupport`, `kurtosis` or `entropy`.
2. When writing probabilistic code I would often use plain Julia + Distributions.jl to prototype and then want to transform it into a probabilistic program, switching to Gen requires rewriting some code (especially for distributions where Gen does not offer a distribution object, but samples straight away). Programs with probabilistic choices are more primitive than Gen programs, so I prefer them to be the basis of my implementation.
3. There are always some tweaks / optimisations / additions to distributions that one may want to contribute upstream, but it feels wrong to contribute these to Gen rather than a more widely used Distributions.

I understand the need for Gen to evolve freely and would not want to hamper it, but I think if there is a good practice floating around then its worth considering.

(would be keen to see that one liner @alex-lew)

PR #321, opened 2020-10-03
Fix overflows in normal distribution logpdf calculation

keorn:
`logpdf` was returning positive numbers for large differences between `x` and `mu` (large `std` overflowed too).

Checked against Distributions.jl, but please make sure it is correct. Another reason for https://github.com/probcomp/Gen.jl/issues/244


ztangent:
I've checked the math and everything looks correct! Just needed to remove 2 extra lines of code so I created a new PR at #430.

PR #323, opened 2020-10-08
Specify the datatype of `choice_grads`

bzinberg:
Based on the docs, it was not obvious to me what data type the `choice_grads` return value of `Gen.choice_gradients` is supposed to have.  Looking at the unit tests, it seems it's expected to be a choicemap.


bzinberg:
I'll merge this if no one has comments on it by Friday.

PR #324, opened 2020-10-13
Add shorthand variants of `update` and `regenerate` that assume the arguments are unchanged

bzinberg:
Adds shorthand variants of `update` and `regenerate` that assume the arguments are unchanged.  I have so far been in this special case most of the time, and this would save a bit of boilerplate.


bzinberg:
should probably add a unit test to make sure calling the function in the documented way doesn't crash.

marcoct:
That's great. I have also wanted this. It would be great to also do the same for `regenerate`, the form of which would take in a trace and a selection.

bzinberg:
Added the `regenerate` shorthand and unit tests, and I think it's ready to merge.

marcoct (approved):

PR #325, opened 2020-10-15
Note in docs that definitions of `(static)` generative functions and calls to `@load_generated_functions()` are supported as top-level expressions only

bzinberg:
See [thread](https://probcomp.slack.com/archives/C5T7HNLRY/p1602713555118100?thread_ts=1602701467.114500&cid=C5T7HNLRY).


bzinberg:
I'll merge this if no one has comments by Friday.

PR #332, opened 2020-11-07
Fix test case with wrong input type.

ztangent:
Now that Gen's dynamic modeling language type-checks function arguments, one of the test cases ended up erroring because it passed in an `Int` when a `Float64` was expected. This PR fixes the error.


bzinberg (approved):
LGTM.  Curious whether changing the declared types of `x` and `y` to `Real` would have also worked?

ztangent:
Great! And yes @bzinberg, changing to `Real` would also have worked.

PR #341, opened 2020-12-09
Add a constructor for mixture distributions

marcoct:
This PR adds a macro for defining a mixture distribution family for a base distribution family.

Example usage:
```julia
mixture_of_normals = @mixturedist normal (0, 0) Float64

@gen function foo()
     x ~ mixture_of_normals([0.4, 0.5], [-1.0, 1.0], [1.0, 1.0])
end
```
The first argument to the macro is the base distribution, which should be a Gen.Distribution.
The second argument defines whether the arguments to the base distribution are scalars or `Array`s and if they are `Array`s, what their dimension is.
The third argument defines the output type, which should match the output type of the base distribution.

Currently, only distributions with all-scalar arguments are supported (e.g. mixtures of multivariate normal is not supported).


ztangent:
Yay! It seems like it'd be nice to also support mixtures of different base distributions? e.g. normals + continuous uniforms?

Also, I wonder if a shorter `@mixture` macro would be enough -- would that clash with anything?

ztangent:
It looks like Distributions.jl also has a `MixtureModel` structure: https://juliastats.org/Distributions.jl/stable/mixture/

marcoct:
I added support for heterogeneous mixtures formed by a fixed list of base distributions, so now there are two macros, one for defining fixed-length heterogeneous mixtures, and the other for defining variable-length homogeneous mixtures.

I also added support for base distributions that take non-scalar arguments, and this functionality is tested for a homogeneous mixture of multivariate normal distributions.

src/modeling_library/mixture.jl
```diff
@@ -1,33 +1,47 @@
...
+The third argument `arg_dims` must be a literal tuple of Ints, of length equal
+to the number of arguments taken by the base distribution family.  A value of 0
+at a position in the tuple an indicates that the corresponding argument to the
+base distribution is a scalar, and integer values of i for i >= 1 indicate that
+the corresponding argument is an i-dimensional array.
 
 Example:
 
     mixture_of_normals = @hom_mixture Float64 normal (0, 0)
 """
```
alex-lew: Can we update these docs to also describe the usage of the resulting mixture distribution? (It's not obvious to me that the calling pattern is `dist(weights, arg1_for_all_components, arg2_for_all_components, ...)`.)

src/modeling_library/mixture.jl
```diff
@@ -37,131 +51,145 @@ macro hom_mixture(output_type, base_dist, arg_dims)
...
 export @hom_mixture
 
+
 ##############################################################################
 # heterogeneous mixture: fixed number of potentially different distributions #
 ##############################################################################
 
-const mixed_syntax_err_msg ="Expected syntax of the form @mixed Float64 (normal, 2) (uniform, 2)"
-const mixed_weight_vector_err_msg = "length of weight vector must match the number of components"
+const MIXED_SYNTAX_ERR_MSG ="Expected syntax of the form @mixed Float64 (normal, 2) (uniform, 2)"
```
alex-lew: Should this say `@het_mixture` rather than `@mixed`?

src/modeling_library/mixture.jl
```diff
@@ -0,0 +1,283 @@
...
+
+The remainder of arguments are tuples, where the first element must have type
+Gen.Distribution, and defines the distribution family of this component in the
+mixture, and the second element is a literal `Int` that gives the the number of
+arguments taken by this distribution family.
+
+Example:
+
+    uniform_beta_mixture = @het_mixture Float64 (uniform, 2) (beta, 2)
+"""
```
alex-lew: Can we clarify here too that the resulting calling pattern is `dist(weights, args_for_dist_1..., args_for_dist_2...)`?

alex-lew:
@marcoct This is great! I left a couple comments, but overall I'm happy to merge.

I do have some questions for the longer term:

1. Why do we need this to be a macro, rather than a 'distribution combinator'? Is it that this keeps more avenues open for performance engineering in the future?

2. I wonder if the interfaces for `hom_mixture` and `het_mixture` can be made more similar -- e.g., you could imagine a unified calling pattern for the created distributions (maybe `dist(weights, component_1_args, component_2_args, ...)`). If each set of `args` is in its own tuple / vector, then you wouldn't even need the user to supply the information about number of arguments (or argument dimension, if I understand correctly how that info is currently being used).

src/modeling_library/mixture.jl
marcoct: > * Why do we need this to be a macro, rather than a 'distribution combinator'? Is it that this keeps more avenues open for performance engineering in the future?
> *

That was ostensibly the idea. But I'll see what a non-macro version looks like.

marcoct:
> I wonder if the interfaces for hom_mixture and het_mixture can be made more similar -- e.g., you could imagine a unified calling pattern for the created distributions (maybe dist(weights, component_1_args, component_2_args, ...)). If each set of args is in its own tuple / vector, then you wouldn't even need the user to supply the information about number of arguments (or argument dimension, if I understand correctly how that info is currently being used).

I decided to vectorize arguments in the homogeneous case so that ReverseDiff.jl-based AD would work out of the box (it works with arrays, but not with tuples). (Also, capturing the homogeneity in in the interface might also be useful for provisioning for performance optimizations.)

marcoct:
I implemented a non-macro version of the homogeneous mixture case, and there is a 1-3x slow down for the test cases, maybe due to increased dependency on splatting. But the code is of course substantially easier to read.

alex-lew:
> I implemented a non-macro version of the homogeneous mixture case, and there is a 1-3x slow down for the test cases, maybe due to increased dependency on splatting. But the code is of course substantially easier to read.

@marcoct Oh, that reminds me — one worry about the `arg[:, :, j]` thing is that I think it performs a copy?

marcoct:
I also implemented a non-macro version of the heterogenous case, which resulted in a 2x slowdown.

PR #343, opened 2020-12-12
(Ready for review): gradient propagation for state in params/choice gradients in Unfold

### LINKED ISSUE #79: Support backpropagation through unfold and recurse
Currently, this is not supported:

https://github.com/probcomp/Gen/blob/master/src/modeling_library/unfold/unfold.jl#L40-L41
https://github.com/probcomp/Gen/blob/master/src/modeling_library/recurse/recurse.jl#L130-L131

femtomc:
Fixes (part of) #79.

Initial commit should fix backpropagation for state in `accumulate_param_grads!` - beginning testing with algebraic computations to make sure that the `state_arg_grad` is correct.

Then I'll begin working on `choice_gradients`.

Note - I had to significantly change `accumulate_param_grads!` in `Unfold`:

```julia
    local kernel_arg_grads::Tuple = (nothing, nothing) # initialize for (len index, state)
    for key = len : -1 : 1 # walks backward over chain
        subtrace = trace.subtraces[key]
        kernel_retval_grad = (retval_grad == nothing) ? nothing : retval_grad[key]
        
        if state_has_grad
            state_retval_grad = (key == len) ? nothing : kernel_arg_grads[2]
            kernel_arg_grads = map((kernel_retval_grad, state_retval_grad)) do a
                accumulate_param_gradients!(subtrace, a)
            end # run two passes - first for retval, second for state.
            kernel_arg_grads = Tuple(map(x -> fold_sum(x), zip(kernel_arg_grads...))) # collapse grads
            @assert kernel_arg_grads[1] == nothing
            for (i, (grad, has_grad)) in enumerate(zip(kernel_arg_grads[3:end], params_has_grad))
                if has_grad
                    params_grad[i][key] = grad
                end
            end
        
        else # ignore state grad
            kernel_arg_grads = accumulate_param_gradients!(subtrace, kernel_retval_grad)
            @assert kernel_arg_grads[1] == nothing
            @assert kernel_arg_grads[2] == nothing
            for (i, (grad, has_grad)) in enumerate(zip(kernel_arg_grads[3:end], params_has_grad))
                if has_grad
                    params_grad[i][key] = grad
                end
            end
        end
    end
    (nothing, kernel_arg_grads[2], params_grad...)
```

Importantly, I expect that this "reversal" of the gradient walk is the correct thing - because you need to start at the end to correctly compute the `state_arg_grad` for the next iteration.


femtomc:
@marcoct or @bzinberg this PR is ready for review. 

I've implemented two new sets of tests for the new gradient functionality (e.g. propagation through Unfold state):

1. Compute parameter accumulation for a simple algebraic chain (the gradient is just a polynomial evaluated at a value).
2. Compute choice gradients for a Markov chain with propagation over state with `nothing` retval grads and retval grads.

Let me know when you get a chance to take a look. I had to do a bit of surgery to the gradient code in `Unfold` so I suspect that there will be some things to chat about.

femtomc:
PS I think I removed the `Manifest.toml` file from the toplevel - not sure if this should stay around anyways ? I can restore if required.

ztangent:
@femtomc I think it's a good idea to put `Manifest.toml` back for now! Last I asked @marcoct about this, he wanted to include it so that a working environment could be easily replicated.

src/modeling_library/unfold/backprop.jl
```diff
@@ -26,17 +66,21 @@ function accumulate_param_gradients!(trace::VectorTrace{UnfoldType,T,U}, retval_
...
+        @assert kernel_arg_grads[1] == nothing
+        state_has_grad || @assert kernel_arg_grads[2] == nothing
+        for (i, (grad, has_grad)) in enumerate(zip(kernel_arg_grads[3:end], params_has_grad))
             if has_grad
                 params_grad[i][key] = grad
             end
         end
     end
-    (nothing, nothing, params_grad...)
+    (nothing, kernel_arg_grads[2], params_grad...)
```
marcoct, line 85: I think there was previously a bug here involving `params_grad` (they need to be summed across 'keys' and they are not), and this bug remains. Also, the corresponding test code was and remains commented out: 

        #@test isapprox(input_grads[3], expected_xs_grad) # alpha
        #@test isapprox(input_grads[4], expected_ys_grad) # beta
        #@test isapprox(input_grads[4], expected_ys_grad) # beta

PR #344, opened 2020-12-14
(Ready for review): add diagram for Switch combinator to docs

femtomc:
1. What font is used in the LaTeX generation ? I used OmniGraffle to render the `.png`, and wanted to copy the style and font but couldn't identify the font.


ztangent:
@femtomc the layout of the diagram itself looks great to me! As for the font, it looks like the other combinator papers were taken from the [Gen PLDI 2019 paper](https://dl.acm.org/doi/10.1145/3314221.3314642), which uses the ACM article fonts. This StackExchange answer shows how to use the same fonts: https://tex.stackexchange.com/questions/540011/font-used-in-acm-template.

marcoct:
I am happy to merge this. Establishing uniformity of visual style and diagrams across Gen docs can be a separate issue that I think should also involve adding source code for the visuals that will make it easier for new diagrams to be added in the same visual style. Here is a new issue I created for this: https://github.com/probcomp/Gen.jl/issues/377

Issue #345, opened 2021-01-04
Splicing of generative functions fails for untraced calls.


ztangent:
It appears I introduced this regression in #322, oops:

```julia
MethodError: no method matching splice(::Gen.GFUntracedState, ::DynamicDSLFunction{Any}, ::Tuple{ProbAStarPlanner,Domain,State,GoalSpec})
Closest candidates are:
  splice(::Gen.GFSimulateState, ::DynamicDSLFunction, ::Tuple) at C:\Users\...\.julia\packages\Gen\2eMtm\src\dynamic\simulate.jl:52     
  splice(::Gen.GFGenerateState, ::DynamicDSLFunction, ::Tuple) at C:\Users\...\.julia\packages\Gen\2eMtm\src\dynamic\generate.jl:74     
  splice(::Gen.GFProposeState, ::DynamicDSLFunction, ::Tuple) at C:\Users\...\.julia\packages\Gen\2eMtm\src\dynamic\propose.jl:51       
  ...
```

ztangent:
Addressed by #346

PR #346, opened 2021-01-05
Fix splicing of generative functions in untraced calls

ztangent:
Addresses #345 by adding an additional method for `splice`. Also adds a test case so we don't miss this in the future.


marcoct (approved):

marcoct:
Nice catch!

Issue #350, opened 2021-01-30
sampling from beta distribution leads to +Inf  logpdf


belledon:
Currently, it is possible that `Gen.random(beta, 0.01, 0.01)` samples `0,1` leading to a `Inf` in the logpdf which in turn leads to `nan`s in associated log scores. 

this is due to a problem illustrated over in `Distributions.jl` https://github.com/JuliaStats/Distributions.jl/issues/1003

not sure if there is much to do, but the source was difficult for me to locate and perhaps it will save someone time. 

Given that `Inf` loglikelihoods should be impossible (not 100% sure), perhaps there can be a check for this a runtime when the logscore is being computed? happy to chip in where possible

jamesonquinn:
https://github.com/probcomp/Gen.jl/pull/359

belledon:
sorry just saw this, many thanks!

PR #359, opened 2021-02-12
Avoid Inf from logpdf(::Beta,...) (issue #350)

jamesonquinn:
This adds 2 special cases to logpdf(::Beta,...)

- If x==0 and alpha<1, then effectively use x=eps/e (machine epsilon divided by e)
- If x==1 and beta<1, then effectively use x=1-eps/e (machine epsilon divided by e)

### DIFF
--- a/src/modeling_library/distributions/beta.jl
+++ b/src/modeling_library/distributions/beta.jl
@@ -12,7 +12,15 @@ const beta = Beta()
 
 function logpdf(::Beta, x::Real, alpha::Real, beta::Real)
     (x < 0 || x > 1 ? -Inf :
-    (alpha - 1) * log(x) + (beta - 1) * log1p(-x) - logbeta(alpha, beta) )
+        ((x == 0 && alpha <= 1) ?      #Special case to avoid infinite PDF at 0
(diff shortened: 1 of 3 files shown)


src/modeling_library/distributions/beta.jl
```diff
@@ -12,7 +12,15 @@ const beta = Beta()
 
 function logpdf(::Beta, x::Real, alpha::Real, beta::Real)
     (x < 0 || x > 1 ? -Inf :
-    (alpha - 1) * log(x) + (beta - 1) * log1p(-x) - logbeta(alpha, beta) )
+        ((x == 0 && alpha <= 1) ?      #Special case to avoid infinite PDF at 0
+            (alpha - 1) * log(eps(typeof(x))) -
```
alex-lew, line 16: If the user uses a constraint like `choicemap(:x => 0)`, there's a chance that `x` will be of type `Int`. It looks like this will cause `eps(typeof(x))` to error, but that's probably fine (the user can always use 0.0).

alex-lew (approved):
Looks good to me -- thanks!

Issue #256, opened 2020-05-07
Sampling bug when one of the choices in `categorical` has probability 1


ztangent:
There seems to be a bug with sampling from categorical when the probabilities given place all mass on a single distribution. It looks like this caused by a buggy upstream implementation of sampling from a categorical in `Distributions.jl` :
```julia
julia> sum([categorical([1, 0]) for i in 1:100])
155
julia> sum([categorical([0.9999, 0.0001]) for i in 1:100])
100
julia> sum([rand(Gen.Distributions.Categorical([1, 0])) for i in 1:100])
155
```
Will open a corresponding issue in `Distributions.jl` if this bug is present in the latest version.

ztangent:
Yup, just confirmed that the bug is also present in the latest version of Distributions.jl: https://github.com/JuliaStats/Distributions.jl/issues/1111

ztangent:
As a hot-fix, it seems like passing in 1.0 instead of 1 leads to the correct behavior:
```
julia> sum([categorical([1.0, 0]) for i in 1:100])
100
```
Thanks @alex-lew for noticing this!

jamesonquinn:
Works for me with Distributions v0.24.10. I'll add your example as a test case to close this bug.

PR #360, opened 2021-02-12
Add test case for #256

jamesonquinn:
"Fixes" (actually, confirms the upstream fix of) issue #256.


ztangent:
Looks great, thanks!

ztangent:
Actually, I just realized that since the upstream fix only seems to have been merged in Distributions 0.24.9 onwards, I think we should only merge this if we bump the `[compat]` version number for Distributions.jl in the `Project.toml` up to 0.24. Thoughts on whether that should be fine @marcoct?

Issue #129, opened 2019-09-30
Make traces more reliably (de)serializable


marcoct:
Traces of generative functions aren't generally reliably (de)serializable because they rely on anonymous functions (which have gensym'd names), and in some cases use generated types that make heavy use of gensyms. This means that the environment the code is deserialized into needs to _exactly_ match the environment that it was serialized from in terms of use of gensyms, which is not reliably the case.

Note that the author of a generative function can always write custom (de)serialization code for their traces, and this is the recommended approach at the moment. Also, there is no general guarantee that the data types used as arguments, return values, or random choices inside a trace are themselves reliably serializable in a generic way.

One reason that traces of dynamic modeling language generative functions (https://github.com/probcomp/Gen/blob/master/src/dynamic/trace.jl) are not reliably (de)serializable is that they refer  to the generative function that produced them, which in turn refers to an anonymous Julia function, whose name is a gensym. (Traces of static modeling language generative functions are likely less reliably serializable than dynamic traces, because they involve much more generated code.)

Gen could provide a `SerializableTrace` data type, which would be something like the trace of a dynamic modeling language function, but would be more reliably (de)serializable (using whatever means the user wants e.g. Julia's `serialize` JLD2 etc.). This new data type would not include a reference to the source generative function, avoiding the particular gensym issue above for dynamic traces. Each generic trace type would implement the `Gen.to_serializable_trace(::Trace)` function, and a `Gen.from_serializable_trace(::SerializableTrace, ::GenerativeFunction)`. The serializable version of the trace does not need to store any cached values, as long as they can be reconstructed deterministically during `from_serializable_trace`. 

Note that care should be taken with respect to the interaction with untraced randomness. An implementation of `from_serializable_trace` that ran `Gen.generate` on a choice trie would not produce the entire trace. That is, storing the arguments and choice trie alone in the serialized trace is not sufficient (e.g. the score could change).

Also, this should be implemented in a compositional way, so that each generative function type (including user-defined types) has its own implementation of `to_serializable_trace` and `from_serializable_trace`, and the implementations of these for built-in types should recursively call the same functions on their subtraces.

marcoct:
Here is a version that works when there is no non-addressable (aka 'untraced') randomness (see https://www.gen.dev/dev/ref/gfi/#Mathematical-concepts-1):

```
import Serialization
import Gen
function save_trace(trace, filename)
	choices = Gen.DynamicChoiceMap(get_choices(trace))
[...]

PR #369, opened 2021-02-13
[WORK IN PROGRESS]  More Reliable Trace Serialization

georgematheos:
Here's some progress on #129 from the hackathon.  Some more debugging work is needed, as well as writing a bit of documentation.

The tests appeared to be passing when I ran them last, but I think there are still bugs to fix which they are not catching.

The 2 main issues I have in mind:
1. It looked to me like currently, serializing Static DSL traces is not working as intended, and is falling back to constructing `DefaultST`s instead of using `GenericST`s.  (If this is true, the intended mechanism for static trace serialization is untested.)
2. The current test for whether serializing and deserializing traces works is insufficient, because we attempt to deserialize in the exact same environment as where we serialized.  This means that the problems #129 describes are not triggered by this test!  The test also currently checks for trace equality by testing choicemap equality, which may be insufficient.


georgematheos:
A few more TODO reminders:
1. Consider renaming `GenericST` to `GenericSerializableTrace` (and maybe same for `DefaultST`).
2. If I'm going to include `DefaultST`, I should test it.
3.  I currently have a commented-out implementation of a `ChoiceMapST` which serializes the choicemap for a trace and unserializes using the choicemap as well.  I should think about whether I want to (A) uncomment it and add documentation so users can access it, but not use it, (B) uncomment it and change eg. `CustomDeterm...` to use it, or (C) just remove it.

marcoct:
@georgematheos Great! When you get a chance, can you also document how a user can ensure their custom generative function trace (and traces of any e.g. DML or SML functions that invoke it) is serializable? E.g. should they implement a custom serialize method for it, or is there another intended way? Or, if that's out of scope, that's fine to document too.

georgematheos:
@marcoct this is now ready for review.

During review, could you please look at:
- In the documentation, should we list the serialization-related methods as being part of the GFI?
- Let me know if you have suggestions on the docs in the extending Gen section.

A possible shortcoming of the current testing suite is that it does not simulate the process of saving a trace, quitting a Julia session, and then reloading the trace.  It just serializes and deserializes into IOBuffers in RAM -- but it is possible that there are errors relating to the loaded Julia environment changing that this would not catch.  Perhaps a better test suite would actually launch a new Julia process to deserialize a trace in.

To confirm that things work, I manually serialized a static DSL trace, then deserialized it in a fresh Julia session; it worked.

Issue #313, opened 2020-09-21
Variant of extending trace translator


marcoct:
From https://github.com/probcomp/Gen.jl/pull/290:

> I wonder if it might be possible to also support a "less simple" ExtendingTraceTranslator", which allows users to provide a TraceTransformDSLProgam" to map choices generated by q_fwd to choices within p_new? This might arise in cases where the proposal doesn't sample choices that also exist within the extended model. For example, in a time series model, where the models trace addresses are nested within a namespace like 1 within an Unfold combinator, it might be nice to write a proposal which doesn't also have to nest its choices within that namespace, and instead specify a trace transform to do the appropriate mapping.

> This will also be really helpful if q_fwd is itself a Map or Unfold combinator, in which case a trace transform would be required to remap all the indices from 1, ..., t to T+1, ..., T+t.

PR #370, opened 2021-02-13
(WIP) Add ExtendingTraceTranslator.

ztangent:
Addresses #313 by adding a more flexible variant of `SimpleExtendingTraceTranslator`: `ExtendingTraceTranslator` accepts a bijective trace transform that remaps the choices sampled by the forward proposal distribution, before merging those choices into the extended model via `update`.

Before I move forward with adding documentation and test cases, I'm wondering if perhaps this should just replace the existing `SimpleExtendingTraceTranslator`? This does everything that `SimpleExtendingTraceTranslator` does, and I don't think it adds much of a performance overhead to have the `if-else` clause in `ExtendingTraceTranslator` that checks whether a trace transform needs to be applied. For backwards compatibility, I could even make `SimpleExtendingTraceTranslator` a constructor alias for an  `ExtendingTraceTranslator` with the bijection filled in with `nothing`.

### DIFF
--- a/src/inference/trace_translators.jl
+++ b/src/inference/trace_translators.jl
@@ -764,7 +764,7 @@ end
         argdiffs::Tuple = (),
         new_obs::ChoiceMap = EmptyChoiceMap(),
         q_fwd::GenerativeFunction,
-        q_fwd_args::Tuple  = ())
+        q_fwd_args::Tuple = ())
 
 Constructor for a simple extending trace translator.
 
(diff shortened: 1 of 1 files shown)


ztangent:
This work is due to be superseded by the implementations of `ExtendingTraceTranslator` and `UpdatingTraceTranslator` in:
https://github.com/probcomp/GenParticleFilters.jl/blob/3394832987e43d4214bddf49fcdccbf4098c7c18/src/translate.jl

PR #371, opened 2021-02-13
Add `has_submap`  method for ChoiceMaps (#124)

### LINKED ISSUE #124: `has_submap(choicemap, addr)::Bool`
Seems like a useful thing to have in the API.  Currently I'm using the implementation
```jl
function has_submap(c::Gen.ChoiceMap, addr)
  addr in keys(collect(Gen.get_submaps_shallow(c)))
end
```

alex-lew:
This PR implements the `has_submap` method that Ben suggested adding in #124.

The implementation is slightly different from what he suggested. A ChoiceMap has a submap at an address if:

* It does not have a value at that address, and
* `get_submap(choices, addr)` returns a non-empty ChoiceMap.

This accords with the spec of `get_submap`, which is supposed to throw an error if it finds a value at the given address, and return an empty ChoiceMap if there is no submap at the address.

[That said, I've sometimes seen "orphan" empty choice maps show up when printing, which makes me think that at least some kinds of choicemap implement `get_submaps_shallow` to return both empty and non-empty submaps (the docs say to return only the non-empty ones).]

### DIFF
--- a/src/choice_map.jl
+++ b/src/choice_map.jl
@@ -31,6 +31,13 @@ no value exists. A syntactic sugar is `Base.getindex`:
 function get_value end
 
 """
+    has_submap(choices::ChoiceMap, addr)
+
+Return true if there is a non-empty sub-assignment at the given address.
+"""
+function has_submap end
+
+"""
     key_submap_iterable = get_submaps_shallow(choices::ChoiceMap)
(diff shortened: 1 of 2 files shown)


marcoct:
Looks good to me!

PR #374, opened 2021-02-13
Fix length of iterator usages

marcoct:


marcoct:
This was causing GenExamples.jl/gp_structure/incremental.jl to fail

PR #382, opened 2021-02-16
Pretty print for static IR

ztangent:
Given a static generative function like this:
```julia
@gen (static) function datum(x::Float64, (grad)(params::Params))
    is_outlier = @trace(bernoulli(params.prob_outlier), :z)
    std = is_outlier ? params.outlier_std : params.inlier_std
    y = @trace(normal(x * params.slope + params.intercept, std), :y)
    return y
end
```

This PR makes it so that the IR is automatically pretty-printed like this:
```julia
julia> Gen.get_ir(datum)
== Static IR ==
Arguments: (x, params)
  ##.#445 = #15(params)
  ##z#442 = @trace(Gen.Bernoulli()(##.#445), :z)
  std = #16(params, ##z#442)
  ##call#451 = #17(params, x)
  ##y#448 = @trace(Gen.Normal()(##call#451, std), :y)
  return ##y#448
```

Should come in handy for debugging domain-specific compilers to IR, as well as any future improvements made to the IR or static function compilation.

### DIFF
--- a/src/static_ir/static_ir.jl
+++ b/src/static_ir/static_ir.jl
@@ -27,6 +27,9 @@ Most generative function interface methods are generated from the intermediate r
 """
 abstract type StaticIRGenerativeFunction{T,U} <: GenerativeFunction{T,U} end
 
+Base.nameof(gen_fn::StaticIRGenerativeFunction) =
+    nameof(type_of(gen_fn))
+
 function get_ir end
 function get_gen_fn_type end
 function get_options end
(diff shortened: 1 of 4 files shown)


src/static_ir/static_ir.jl
```diff
@@ -52,6 +52,7 @@ function generate_generative_function(ir::StaticIR, name::Symbol, options::Stati
             params::Dict{Symbol,Any}
         end
         (gen_fn::$gen_fn_type_name)(args...) = $(GlobalRef(Gen, :propose))(gen_fn, args)[3]
+        $(GlobalRef(Gen, :get_ir))(::$gen_fn_type_name) = $(QuoteNode(ir))
```
alex-lew, line 55: This seems reasonable, but it's making me wonder whether there was any reason we didn't have this in the first place. I guess the answer is probably that `get_ir` was mostly used from `@generated` functions, which only had access to the type of the GF, not the GF itself.

alex-lew (approved):

Issue #396, opened 2021-03-27
Error when running DeterministicTraceTranslator


ztangent:
When running `DeterministicTraceTranslator`, `run_first_pass` results in an error because it passes in `aux_trace` as `nothing` instead of an actual `Trace`, which prevents `FirstPassState` from being properly constructed:

```julia
ERROR: MethodError: no method matching get_choices(::Nothing)
Closest candidates are:
  get_choices(::var"##StaticIRTrace_line_model#460") at none:0
  get_choices(::var"##StaticIRTrace_line_step#441") at none:0
  get_choices(::Gen.CallAtTrace) at C:\Users\...\.julia\packages\Gen\k43lp\src\modeling_library\call_at\call_at.jl:52
  ...
Stacktrace:
 [1] Gen.FirstPassState(::Gen.DynamicDSLTrace{DynamicDSLFunction{Any}}, ::Nothing) at C:\Users\...\.julia\packages\Gen\k43lp\src\inference\trace_translators.jl:260
 [2] run_first_pass(::TraceTransformDSLProgram, ::Gen.DynamicDSLTrace{DynamicDSLFunction{Any}}, ::Nothing) 
at C:\Users\...\.julia\packages\Gen\k43lp\src\inference\trace_translators.jl:266
```

It looks like commit 77e5dab introduced a regression when it changed the internal structure of `FirstPassState` to be more stringent and catch certain kinds of errors.

ztangent:
I'm now running into a separate bug, caused by a parsing error in the `@transform` macro: at line 129, `src_expr` should be `dest_expr` instead:

https://github.com/probcomp/Gen.jl/blob/ee903a4d206413c5379f90de1eb3aee52e15bb74/src/inference/trace_translators.jl#L128-L133

`@transform` also currently does not throw an error if `err` becomes true, which should be fixed.

ztangent:
And one more bug: The following code assumes that `bij_state.aux_trace` is always a trace, but when using a `DeterministicTraceTranslator`, it is just `nothing`, which causes an error at line 144:

https://github.com/probcomp/Gen.jl/blob/ee903a4d206413c5379f90de1eb3aee52e15bb74/src/inference/trace_translators.jl#L139-L151

ztangent:
Addressed by #397.

Issue #395, opened 2021-03-27
Trace translator type improvements.


ztangent:
Two improvements I wanted to suggest that I believe will useful in downstream applications or library code:

1. Making trace translators `mutable`, so that we can change the `new_observations` or `p_new_args` that will be supplied without constructing a new object each type (useful in e.g. particle filtering).
2. Making trace translators inherit from an abstract `TraceTranslator` type, so that downstream code can dispatch on this type.

ztangent:
Addressed by #397.

Issue #385, opened 2021-02-19
Typos and inconsistent naming in trace translators.


SamWitty:
When running the example code in https://www.gen.dev/dev/ref/trace_translators/#General-Trace-Translators-1, julia throws the following error:

`ERROR: UndefVarError: proposal not defined
Stacktrace:
 [1] (::GeneralTraceTranslator)(::Gen.DynamicDSLTrace{DynamicDSLFunction{Any}}; check::Bool, prev_observations::EmptyChoiceMap) at /Users/switty/.julia/packages/Gen/unzLD/src/inference/trace_translators.jl:727
 [2] (::GeneralTraceTranslator)(::Gen.DynamicDSLTrace{DynamicDSLFunction{Any}}) at /Users/switty/.julia/packages/Gen/unzLD/src/inference/trace_translators.jl:727
 [3] top-level scope at none:1`

I think the issue will be fixed by replacing `proposal` with `translator.q_forward` at https://github.com/probcomp/Gen.jl/blob/a96a77991e0e43f208272e9241c8f2434ffdedbf/src/inference/trace_translators.jl#L727, but I don't know enough about the codebase to feel comfortable making that change.

For completeness, here is the code that generates the error.

```
@gen function p1()
    x ~ uniform(0, 1)
    y ~ uniform(0, 1)
end

@gen function p2()
    i ~ uniform_discrete(1, 10) # interval [(i-1)/10, i/10]
    j ~ uniform_discrete(1, 10) # interval [(j-1)/10, j/10]
end

@gen function q1()
end

@gen function q2(p2_trace)
    i = p2_trace[:i]
    j = p2_trace[:j]
    dx ~ uniform(0.0, 0.1)
    dy ~ uniform(0.0, 0.1)
end

@transform f (p1_trace, q1_trace) to (p2_trace, q2_trace) begin
    x = @read(p1_trace[:x], :continuous)
    y = @read(p1_trace[:y], :continuous)
    i = ceil(x * 10)
    j = ceil(y * 10)
    @write(p2_trace[:i], i, :discrete)
    @write(p2_trace[:j], j, :discrete)
    @write(q2_trace[:dx], x / 10, :continuous)
    @write(q2_trace[:dy], y / 10, :continuous)
end

@transform f_inv (p2_trace, q2_trace) to (p1_trace, q1_trace) begin
    i = @read(p2_trace[:i], :discrete)
    j = @read(p2_trace[:j], :discrete)
[...]

PR #397, opened 2021-03-27
Trace translator fixes, test cases, and minor enhancements.

ztangent:
This PR fixes the multiple bugs noted in #385 , #396 that prevented `DeterministicTraceTranslator`s and `GeneralTraceTranslator`s from running correctly. It also adds test cases based on the examples in the trace translator documentation, and fixes a few bugs that were present in those examples. Finally, it implements the minor enhancements suggested in #395, creating an abstract `TraceTranslator` type, and making the existing structs mutable for easier re-use (e.g. changing the observations supplied by a trace translator).

Working test case for `DeterministicTraceTranslator`s:

https://github.com/probcomp/Gen.jl/blob/14b24035fe1cdb49727aa068bc5bcb97a7da65c8/test/inference/trace_translators.jl#L3-L42

Working test case for `GeneralTraceTranslator`s:

https://github.com/probcomp/Gen.jl/blob/14b24035fe1cdb49727aa068bc5bcb97a7da65c8/test/inference/trace_translators.jl#L116-L169

I also added documentation for `SimpleExtendingTraceTranslator`, and refactored some code for readability / reusability.


docs/src/ref/trace_translators.md
```diff
@@ -78,7 +78,7 @@ Note that the transform DSL code does not specify what the two generative functi
 This information will be required for computing probabilities and probability densities of traces.
 We provide this information by constructing a **Trace Translator** that wraps the transform along with this transformation:
 ```julia
-translator = DeterministicTraceTranslator(p2, (), f)
+translator = DeterministicTraceTranslator(p2, (), choicemap(), f)
```
marcoct, line 81: DeterministicTraceTranslator accepts keyword arguments, and the bug was that I didn't include them in the original docs here. I want to keep this simple and avoid the choicemap() at this point in the docs.
```
translator = DeterministicTraceTranslator(p_new=p2, p_new_args=(), transform=f)
```
(then, we can rename `f` to `transform` and `p_args` to `p_new_args` as well below)

Also, in the docs below in the Observations section, we should use:

```
translator = DeterministicTraceTranslator(p_new=p2, p_new_args=(), transform=f, observations=observations)
```

I can try to make these changes in this PR if you don't mind.

ztangent: Sounds good! I think we should just have consistent fieldnames across all the TraceTranslator types -- `SymmetricTraceTranslator` currently uses `involution`, and all of the others use `f`, I believe. Agree that it should be `p_new_args`, since that's what the others all use.

src/inference/trace_translators.jl
```diff
@@ -813,26 +864,32 @@ end
...
     q::GenerativeFunction
     q_args::Tuple = ()
     involution::T # an involution
 end
 
-function symmetric_trace_translator_run_transform(
-        involution::TraceTransformDSLProgram,
-        prev_model_trace::Trace, forward_proposal_trace::Trace,
-        q::GenerativeFunction, q_args::Tuple)
+function inverse(translator::SymmetricTraceTranslator, prev_model_trace=nothing)
```
marcoct, line 886: Format nitpick (if formatting changes are in scope :) ); that's a good candidate for:

```
inverse(translator::SymmetricTraceTranslator, prev_model_trace=nothing) = translator
```

(see first part of https://github.com/invenia/BlueStyle#method-definitions)

test/inference/trace_translators.jl
```diff
@@ -0,0 +1,171 @@
+@testset "trace translators" begin
```
marcoct, line 1: Nice!

marcoct (approved):
This is great, especially the tests and bug fixes, but also some of the code cleanup.

I am happy to merge, but there are two name changes for keyword arguments (and corresponding name changes in the docs) that would be great to include in this PR.

docs/src/ref/trace_translators.md
ztangent: Oh, one thing to keep in mind though is that some code already uses `involution` or `f` as a keyword argument, and if we change the keyword argument now, that'll be a breaking change. I wonder if it's a good idea to just define `involution` and `f` as aliases of transform...? e.g. by adding constructor definitions like:

```julia
SymmetricTraceTranslator(;involution::TraceTransformDSL=error("No default for `involution`."), kwargs...) =
    SymmetricTraceTranslator(;transform=involution, kwargs...)
```

We can even throw in a deprecation warning.

ztangent:
After some discussion, outstanding issues have been copied to #399, and this PR will be merged!

PR #413, opened 2021-05-13
Replace FixedStepSizeGradientDescent with FixedStepGradientDescent in Gen docs.

JGameCreation:
… example code


ztangent:
Thanks for this catch! The Travis build is erroring due to [an upstream bug](https://github.com/JuliaDiff/ReverseDiff.jl/issues/175), not this patch (which only changes the docs), so I'm going to merge this :)

PR #416, opened 2021-05-14
Fix typos in the Involutive MCMC examples

Werpers:
* `merge_means` was referred to as `merge_mean` in the involution definition.
* `split_merge_kernel` returned both the new trace and wether the move was accepted from `mh`. 

This is my first interaction with Gen so please make sure the suggested changes make sense! :)


ztangent (approved):
Thanks for catching these!

PR #422, opened 2021-06-16
Allow SML functions to be generated by macros

marcoct:
Previously there was a bug [here](https://github.com/probcomp/Gen.jl/blob/3669eb9db361972d0191a0b3e62eb8a06f4b158e/src/dsl/static.jl#L109-L111) that involved excess use of `esc` when constructing the anonymous Julia functions used to represent Julia computations in SML code.

This bug caused a `LoadError` when a macro returned a quoted expression that included a `@gen (static) function .. end` block that includes Julia expressions in the body, because the code that is returned is malformed due to having qualified symbols used as formal parameter names in the anonymous function definition:
```julia
macro make_bar(constant )
    return quote
        @gen (static) function bar(a, b)
            c = $constant
            d = c + a + b
            return d
        end
    end
end

@make_bar 1.0
```
The error was:
```
ERROR: LoadError: syntax: "Main.a" is not a valid function argument name around /home/marcoct/dev/gen_issue_test/minimal_example.jl:44
```
The code returned from the macro was:
```julia
begin
    #= /home/marcoct/dev/gen_issue_test/minimal_example.jl:44 =#
    begin
        var"#4#bindings" = Gen.Dict{Gen.Symbol, Gen.StaticIRNode}()
        var"#5#builder" = Gen.StaticIRBuilder()
        Gen.set_accepts_output_grad!(var"#5#builder", $(QuoteNode(false)))
        Main.:(var"##a#263") = Gen.add_argument_node!(var"#5#builder", name = :a, typ = :Any, compute_grad = $(QuoteNode(false)))
        Main.:(var"##b#264") = Gen.add_argument_node!(var"#5#builder", name = :b, typ = :Any, compute_grad = $(QuoteNode(false)))
        Main.:(var"##c#265") = Gen.add_julia_node!(var"#5#builder", function ()
                    $(QuoteNode(1.0))
                end, inputs = [], name = :c, typ = $(QuoteNode(:($(QuoteNode(Any))))))
        Main.:(var"##d#266") = Gen.add_julia_node!(var"#5#builder", function (Main.a, Main.b, Main.c)
                    Main.c + Main.a + Main.b
                end, inputs = [Main.:(var"##a#263"), Main.:(var"##b#264"), Main.:(var"##c#265")], name = :d, typ = $(QuoteNode(:($(QuoteNode(Any))))))
        Gen.set_return_node!(var"#5#builder", Main.:(var"##d#266"))
        var"#6#ir" = Gen.build_ir(var"#5#builder")
        begin
            $(Expr(:meta, :doc))
            Main.bar = Main.eval(Gen.generate_generative_function(var"#6#ir", :bar, $(QuoteNode(Gen.StaticIRGenerativeFunctionOptions(false, true)))))
        end
    end
end
```
The offending code expression was:
```
function (Main.a, Main.b, Main.c)
    Main.c + Main.a + Main.b
end
```
This PR removes these escapes to fix this issue. I think this error did not occur for SML functions that are not generated by macros because in that case escaping the argument names causes Julia macro-expansion to not produce a qualified name for these variables (since the `Main` module is the default module anyways, so a qualified name would be unnecessary).

However, that caused a regression, because the variables inside the body of the anomymous Julia function are then resolved relative to the `Gen` module if they are not bound in the local scope of the SML function:
```julia
module MyModule
using Gen: @gen, @load_generated_functions

external_var = 1
@gen (static) function uses_external()
    return external_var + 1
end

@load_generated_functions()

end
```
The generated code includes the anonymous Julia function
```
function () Gen.external_var + 1 end
```

This is fixed [here](https://github.com/probcomp/Gen.jl/blob/005d24199fe381ec21e47ba6dacff996c404e79f/src/dsl/static.jl#L122) by ensuring that all symbols that are not defined resolved within the local scope of the SML function are resolved relative to the module that includes the SML definition and not `Gen`.

Also, since this (and likely other places in the implementation) does not respect Julia's scoping rules for Julia function definitions that appear in the SML function body, this is clarified in the documentation [here](https://github.com/probcomp/Gen.jl/commit/93955065d18cb78dae56ea4887daa53fe48c986d).

### DIFF
--- a/src/dsl/static.jl
+++ b/src/dsl/static.jl
@@ -101,7 +101,7 @@ function parse_trace_expr!(stmts, bindings, fn, args, addr, __module__)
     return name
 end
 
-function set_external_vars_module(expr, bindings, __module__)
+function set_module_for_global_constants(expr, bindings, __module__)
     expr = MacroTools.postwalk(expr) do e
         if MacroTools.@capture(e, var_Symbol) && !haskey(bindings, var) && var != :end
             :($__module__.$var)
@@ -119,7 +119,7 @@ function parse_julia_expr!(stmts, bindings, name::Symbol, expr::Expr,
     inputs = collect(resolved)
     input_vars = map((x) -> x[1], inputs)
     input_nodes = map((x) -> esc(x[2]), inputs)
-    expr = set_external_vars_module(expr, bindings, __module__)
+    expr = set_module_for_global_constants(expr, bindings, __module__)
     fn = Expr(:function, Expr(:tuple, input_vars...), expr)
     node = gensym(name)
     push!(stmts, :($(esc(node)) = add_julia_node!(
--- a/src/dsl/static.jl
+++ b/src/dsl/static.jl
@@ -58,7 +58,7 @@ gen_node_name(arg::Symbol) = gensym(arg)
 gen_node_name(arg::QuoteNode) = gensym(string(arg.value))
 
 "Parse @trace expression and add corresponding node to IR."
-function parse_trace_expr!(stmts, bindings, fn, args, addr)
+function parse_trace_expr!(stmts, bindings, fn, args, addr, __module__)
     expr_s = "@trace($fn($(join(args, ", "))), $addr)"
     name = gen_node_name(addr) # Each @trace node is named after its address
     node = gen_node_name(addr) # Generate a variable name for the StaticIRNode
@@ -91,7 +91,7 @@ function parse_trace_expr!(stmts, bindings, fn, args, addr)
         # Create Julia node for each argument to gen_fn_or_dist
         arg_name = gen_node_name(arg_expr)
         push!(inputs, parse_julia_expr!(stmts, bindings,
-            arg_name, arg_expr, QuoteNode(Any)))
+            arg_name, arg_expr, QuoteNode(Any), __module__))
     end
     # Add addr node (a GenerativeFunctionCallNode or RandomChoiceNode)
     push!(stmts, :($(esc(node)) = add_addr_node!(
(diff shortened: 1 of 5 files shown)


marcoct:
Programmatic generation of variants of an SML generative function is generally useful. I need this for a current project where I am benchmarking multiple generative functions that are all the same except for one component (the likelihood model).

src/dsl/static.jl
```diff
@@ -101,14 +101,26 @@ function parse_trace_expr!(stmts, bindings, fn, args, addr)
     return name
 end
 
+function set_module_for_global_constants(expr, bindings, __module__)
+    expr = MacroTools.postwalk(expr) do e
+        if MacroTools.@capture(e, var_Symbol) && !haskey(bindings, var) && var != :end
+            :($__module__.$var)
```
georgematheos, line 107: If I'm understanding, the reason to have `:($__module__.$var)` instead of `esc(var)` is to make sure that `var` is bound in the proper module--is this right?

georgematheos (approved):
@marcoct agreed -- I have wanted this feature several times (including in something I am working on today).  Thank you for implementing it!

I've read through the code, and it looks good to me.  I left one question in the code for a line I didn't fully understand, but I would guess it is fine.

georgematheos:
I have also tested this code locally and it seems to be working for me.  

In particular, I used it for this macro to construct a static dynamical model from a model to produce an initial latent state, a step model to produce subsequent latent states, and an observation model.
```julia
macro DynamicModel(
    initial_latent_model, # no arguments
    latent_step_model, # args = (prev_latents...,)
    obs_model, # args = (current_latents...,)
    n_latents
)
    prev_latent_names = [Symbol("prev_latent$i") for i=1:n_latents]
    latent_names = [Symbol("latent$i") for i=1:n_latents]
    return quote
        @gen (static) function initial_step()
            latents ~ $(initial_latent_model)()
            # unpack the tuple of latents into local variables, as needed to call the obs model
            ($(latent_names...),) = latents
            obs ~ $(obs_model)($(latent_names...))
            return latents
        end

        @gen (static) function take_step(t, prev_latents)
            ($(prev_latent_names...),) = prev_latents
            latents ~ $(latent_step_model)($(prev_latent_names...))
            ($(latent_names...),) = latents
            obs ~ $(obs_model)($(latent_names...))
            return latents
        end

        @gen (static) function dynamic_model(T)
            init ~ initial_step()
            steps ~ Unfold(take_step)(T, init)
        end
    end
end
```

src/dsl/static.jl
ztangent: I think `esc` might also work, but the semantics of `esc` have often confused me, and I don't mind that we're doing this more specifically!

ztangent (approved):
Looks great to me! Really great to enable code gen of SML functions.

PR #425, opened 2021-07-09
Regenerating Switch-Traces should expect kernel arguments

mirkoklukas:
The function `regenerate` for switch traces reads out the kernel index via `args[1]`. However `regenerate` typically gets called with `get_args(tr)`, which for switch-traces returns the *kernel* arguments **not** the *switch* arguments.

### DIFF
--- a/src/modeling_library/switch/regenerate.jl
+++ b/src/modeling_library/switch/regenerate.jl
@@ -53,8 +53,8 @@ function regenerate(trace::SwitchTrace{T},
                     argdiffs::Tuple,
                     selection::Selection) where T
     gen_fn = trace.gen_fn
-    index, index_argdiff = args[1], argdiffs[1]
+    index, index_argdiff = trace.index, NoChange()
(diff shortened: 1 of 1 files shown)


src/modeling_library/switch/regenerate.jl
```diff
@@ -53,8 +53,8 @@ function regenerate(trace::SwitchTrace{T},
                     argdiffs::Tuple,
                     selection::Selection) where T
     gen_fn = trace.gen_fn
-    index, index_argdiff = args[1], argdiffs[1]
+    index, index_argdiff = trace.index, NoChange()
```
femtomc, line 56: The user is allowed to change the branch -- so this is semantically incorrect.

femtomc: The correct fix is to expand `get_args` with `(trace.index, args...)` and then fix any bugs which occur from that.

ztangent:
Fixed in #426!

PR #426, opened 2021-07-09
Fixed inconsistency in Switch trace get_args

### LINKED ISSUE #427: MH updates for SwitchTraces fail
The function `regenerate` for switch traces reads out the kernel index via `args[1]`. However regenerate typically gets called with `get_args(tr)`, which for switch-traces returns the _kernel_ arguments **not** the _switch_ arguments. Error example below.

I think the confusion lies here: The `Switch` is called with args of the form `args = (index, kernel_args...)` but its trace's args are `tr.args = kernel_args`, which are the args of `tr.branch`.

Minimal example:
```julia
@gen function f1(x)
    @trace(bernoulli(0.5), :a)
end

@gen function f2(x)
    @trace(bernoulli(0.5), :a)
end

F = Switch(f1,f2)
tr  = simulate(F, (1, 0.0))
tr, = mh(tr, select(:a))
```

Even simpler:
```julia
@gen function f1()
    @trace(bernoulli(0.5), :a)
end

@gen function f2()
    @trace(bernoulli(0.5), :a)
end

F = Switch(f1,f2)
tr  = simulate(F, (1,))
tr, = mh(tr, select(:a))
```

The MH calls in the examples could also be replaced by:
```julia
args = get_args(tr)
argdiffs = map((_) -> NoChange(), args)
selection = select(:a)
tr, = regenerate(tr, args, argdiffs, selection)
```

femtomc:


femtomc:
Additionally, I moved `cond.jl` (which only held the implementation of the Switch trace) into `trace.jl` under the `switch` directory.

femtomc:
@mirkoklukas if possible, try model on this branch.

PR #433, opened 2021-07-14
Fix `logpdf_grad` for BroadcastedNormal.

### LINKED ISSUE #432: `logpdf_grad(::BroadcastedNormal, ...)` sums over the wrong dims
> the `logpdf_grad` implementation of `broadcasted_norm` seems to only handle the cases where _all_ arguments are scalars (note that it returns 3 scalars as gradients)

Good catch, thanks @MrVPlusOne!  Indeed, in https://github.com/probcomp/Gen.jl/commit/005f97c7176ba19cd6f6b88d00cfa2cb32f25c92#, I blindly inserted a `sum` into both `logpdf` and `logpdf_grad`... whoops.  And indeed, what `logpdf_grad` should instead do is sum over the dims that are "missing" from each arg, compared to the fully broadcasted shape, as you do above.

@MrVPlusOne, if you're up for sending a PR with this fix, I'd be happy to review it!

_Originally posted by @bzinberg in https://github.com/probcomp/Gen.jl/issues/420#issuecomment-879447064_

MrVPlusOne:
As requested by #432, this PR fixes the gradient of BroadcastedNormal for the non-scalar cases.

### DIFF
--- a/src/modeling_library/distributions/normal.jl
+++ b/src/modeling_library/distributions/normal.jl
@@ -86,26 +86,26 @@ function logpdf_grad(::BroadcastedNormal,
     deriv_x = - z ./ std
     deriv_mu = -deriv_x
     deriv_std = -1. ./ std .+ abs2.(z) ./ std
-    (_unbroadcast_for_arg(x, deriv_x), 
-     _unbroadcast_for_arg(mu, deriv_mu), 
-     _unbroadcast_for_arg(std, deriv_std))
+    (_unbroadcast_like(x, deriv_x), 
+     _unbroadcast_like(mu, deriv_mu), 
+     _unbroadcast_like(std, deriv_std))
 end
 
-_unbroadcast_for_arg(::Real, grad) = sum(grad)
-_unbroadcast_for_arg(::Array{Float64, 0}, grad::Real) = fill(grad)
-function _unbroadcast_for_arg(arg::AbstractArray{<:Real, N},
-                              grad::AbstractArray{T}
-                             )::AbstractArray{T, N} where {N,T}
-    size(arg) == size(grad) ? grad : unbroadcast_grad(size(arg), grad)
+_unbroadcast_like(::Real, full_arr) = sum(full_arr)
+_unbroadcast_like(::AbstractArray{<:Real, 0}, full_arr::Real) = fill(full_arr)
+function _unbroadcast_like(arg::AbstractArray{<:Real, N},
+                           full_arr::AbstractArray{T}
+                          )::AbstractArray{T, N} where {N,T}
+    size(arg) == size(full_arr) ? full_arr : _unbroadcast_to_shape(size(arg), full_arr)
 end
 
-function unbroadcast_grad(old_shape::NTuple{l_old, Int},
-                          grad::AbstractArray{T, l_new}
-                         ) where {T, l_old, l_new}
-    @assert l_new >= l_old  
-    new_shape = size(grad)
-    dims=filter(i -> i > l_old || old_shape[i] == 1 && new_shape[i] > 1, 1:l_new)
-    dropdims(sum(grad; dims=dims); dims=tuple((l_old+1:l_new)...))::AbstractArray{T, l_old}
+function _unbroadcast_to_shape(target_shape::NTuple{target_ndims, Int},
+                               full_arr::AbstractArray{T, full_ndims}
+                         ) where {T, target_ndims, full_ndims}
+    @assert full_ndims >= target_ndims  
+    full_shape = size(full_arr)
+    dims=filter(i -> i > target_ndims || target_shape[i] == 1 && full_shape[i] > 1, 1:full_ndims)
+    dropdims(sum(full_arr; dims=dims); dims=tuple((target_ndims+1:full_ndims)...))::AbstractArray{T, target_ndims}
 end
 
 random(::Normal, mu::Real, std::Real) = mu + std * randn()
--- a/src/modeling_library/distributions/normal.jl
+++ b/src/modeling_library/distributions/normal.jl
@@ -65,8 +65,6 @@ function logpdf(::BroadcastedNormal,
     assert_has_shape(x, broadcast_shapes_or_crash(mu, std);
                      msg="Shape of `x` does not agree with the sample space")
     z = (x .- mu) ./ std
-    var = std .* std
-    diff = x .- mu
     sum(- (abs2.(z) .+ log(2π)) / 2 .- log.(std))
 end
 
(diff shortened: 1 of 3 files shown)


ztangent:
Thanks for making this PR! I'm not sure why the Travis build errored, but separate of that, it'd great if you could adjust the implementation to be overflow-safe in line with #321!

src/modeling_library/distributions/normal.jl
```diff
@@ -85,10 +83,29 @@ function logpdf_grad(::BroadcastedNormal,
     assert_has_shape(x, broadcast_shapes_or_crash(mu, std);
                      msg="Shape of `x` does not agree with the sample space")
     z = (x .- mu) ./ std
-    deriv_x = sum(- z ./ std)
+    deriv_x = - z ./ std
     deriv_mu = -deriv_x
-    deriv_std = sum(-1. ./ std .+ abs2.(z) ./ std)
-    (deriv_x, deriv_mu, deriv_std)
+    deriv_std = -1. ./ std .+ abs2(z) ./ std
```
bzinberg: We need to keep `z`'s shape, right?
```suggestion
    deriv_std = -1. ./ std .+ abs2.(z) ./ std
```

src/modeling_library/distributions/normal.jl
```diff
@@ -85,10 +83,29 @@ function logpdf_grad(::BroadcastedNormal,
...
+    (unbroadcast_for_arg(x, deriv_x), 
+    unbroadcast_for_arg(mu, deriv_mu), 
+    unbroadcast_for_arg(std, deriv_std))
```
bzinberg:
```suggestion
    (unbroadcast_for_arg(x, deriv_x), 
     unbroadcast_for_arg(mu, deriv_mu), 
     unbroadcast_for_arg(std, deriv_std))
```

test/modeling_library/distributions.jl
```diff
@@ -126,9 +126,9 @@ end
     f = (x, mu, std) -> logpdf(broadcasted_normal, x, mu, std)
     args = (fill(0.4), fill(0.2), fill(0.3))
     actual = logpdf_grad(broadcasted_normal, args...)
-    @test isapprox(actual[1], finite_diff(f, args, 1, dx; broadcast=true))
-    @test isapprox(actual[2], finite_diff(f, args, 2, dx; broadcast=true))
-    @test isapprox(actual[3], finite_diff(f, args, 3, dx; broadcast=true))
```
bzinberg: Why this change?

bzinberg (commented):
Thanks @MrVPlusOne!  A couple of initial comments now; I'll come back later to suggest unit / regression tests.

src/modeling_library/distributions/normal.jl
MrVPlusOne: Yeah, you are absolutely right!

test/modeling_library/distributions.jl
MrVPlusOne: I'm actually a little unsure how to fix these tests. It seems that the first argument of `isapprox` is a zero-dimensional array (which I believe is the expected behavior) but the second argument is a `Float64`, which causes Julia to complain that no method of `isapprox` matches the given argument types. Any suggestions on how to fix this?

bzinberg: After investigating this a bit, I think it boils down to https://github.com/JuliaLang/julia/issues/28866, which I consider to be a bug in Julia.  Yes -- in the version of the code before this PR, both the LHS and the RHS should be 0-dimensional arrays.  For the RHS, the issue appears to be that these lines
https://github.com/probcomp/Gen.jl/blob/d79aded6909a8e44b3a813179758add8237ed46a/test/runtests.jl#L22-L23
do `::Aray{T,0} .+ ::T` and `::Aray{T,0} .- ::T`, which should give an `::Array{T,0}` but due to https://github.com/JuliaLang/julia/issues/28866 gives a scalar.  So the LHS of the `./` on the following line should have an `Array{T,0}` as the `i`th argument to `f`, thus (since `broadcast = true`) should output an `Array{T,0}`, and consequently the division operation should be `::Array{T,0} ./ ::T` which should output an `Array{T,0}`, but that is not what happens.

I think the cleanest way to fix this is to add a workaround to the (test-only) function `finite_diff` so that it handles zero-dimensional array valued arguments correctly.

bzinberg: Oh whoops, I take that back -- both sides of the `./` are scalars:
https://github.com/probcomp/Gen.jl/blob/d79aded6909a8e44b3a813179758add8237ed46a/test/runtests.jl#L24
I guess what my previous comment really shows is that, if `f` requires its argument to be a zero-dimensional array and errors if given a scalar, `finite_diff` will hit that error.

A fix to `finite_diff` for the zero-dimensional case still seems right :slightly_smiling_face:

bzinberg: (I'm currently drafting this.)

bzinberg: Aha, in addition to the above, the function `f` being supplied to `finite_diff` does not satisfy the requirements in its docstring: it takes array-valued arguments, but is not the broadcast of a function that takes scalar-valued arguments (i.e. `logpdf_grad(vector-valued distribution parameters)` is not a broadcast of a bunch of `logpdf_grad(scalar-valued distribution parameters)`'s).  So I guess we should add a variant of `finite_diff` that has the semantics we need.

MrVPlusOne: Yeah, I think to properly test this, we will need a more general `finite_diff` function or use an autodiff library implementation as a reference.

bzinberg: you clearly trust autodiff more than I do :slightly_smiling_face:

bzinberg:
@MrVPlusOne, I've drafted a fix to the unit test breakage.  Take a look and LMKWYT?

MrVPlusOne:
@bzinberg Thanks! The fix looks good to me!

bzinberg:
@MrVPlusOne, I've made a few more changes that are mostly by way of tidying.

* Rename the vars in the `unbroadcast` helper functions to generically describe what they do, rather than coupling them to how they're used in `logpdf_grad`; e.g., `unbroadcast_for_arg(arg, grad)` becomes `_unbroadcast_like(a, full_arr)`
* Add some documentation, mainly for what "unbroadcast" means since I can imagine that being ambiguous or confusing to a reader who doesn't have the same context we have
* A bit of code tidying without changing the logic
* Line length and indentation

The one functional change I made was to add to a couple of the unit tests an explicit check on the shape of the returned arrays.

I think regression testing is covered by the shape checks that were fixed in https://github.com/probcomp/Gen.jl/pull/433/commits/e712c71b56f9b86b3d6bd872cfcc1c41052f8f99.  They passed before, but that is because the unit tests were incorrect.

@MrVPlusOne, any comments you'd like to make on the above?  If looks good to you, I think this is ready to merge.

MrVPlusOne:
Thanks for the cleaning up; they all look good to me! I think one more nice thing to have would be to add a unit test for the case where the two arguments of `broadcasted_norm` have different numbers of dimensions. We currently don't have a test for that, right?

bzinberg:
Great idea @MrVPlusOne, how's [this](https://github.com/probcomp/Gen.jl/pull/433/commits/5d3f192dc03069956f7b4e23e6889e84857aa744)?

MrVPlusOne:
Thanks, looks good to me!

bzinberg:
Thanks @MrVPlusOne for using and improving Gen!

Issue #367, opened 2021-02-13
Progress: change hint propagation for user defined struct


georgematheos:
Caching some work from the hackathon today.  To enable propagating change hints when constructing objects which hold several values and unpacking values from these objects, we introduce a `Construct(type)` generative function, and a `GetField(type, fieldname)` generative function.

Usage example
```julia
@gen (static) function foo()
    x ~ normal(0, 1)
    y ~ normal(0, 1)
    return Construct(Point2f0)(x, y)
end

@gen (static) function bar(p::Point2f0)
    x ~ GetField(Point2f0, :x)(p::Point2f0)
    return long_computation(x)
end
```

We could add sugar so that if the user writes `Point2f0(x, y)`, it is automatically converted to `Construct(Point2f0)(x, y)`, and if the user writes `point.x, it is converted to `GetField(Point2f0, :x)(point)`.

Here is an untested implementation:
```julia
struct StructDiff{type, T} <: Diff
    diffs::T
    function StructDiff{type}(diffs::T) where {T <: Tuple{Vararg{<:Diff}}}
        return new{type, T}(diffs)
    end
end
function _get_diff(diff::StructDiff, fieldname::Symbol)
    return _static_get_diff(diff, Val(fieldname))
end
@generated function _static_get_diff(diff::StructDiff{type}, ::Val{fieldname}) where {type, fieldname}
    idx = findfirst(fieldnames(type) .== fieldname)
    return :(diff.diffs[$idx])
end

struct Construct{type} <: CustomUpdateGF{type, type} end
Construct(type) = Construct{type}()

function Gen.apply_with_state(::Construct{type}, args) where {type}
    obj = type(args...)
    return (obj, obj)
end

function Gen.update_with_state(::Construct{type}, obj, args, argdiffs::Tuple{Vararg{NoChange}}) where {type}
    return (
        obj,
        obj,
        StructDiff{type}(argdiffs)
    )
end

function Gen.update_with_state(::Construct{type}, obj, args, argdiffs::Tuple) where {type}
    new_obj = type(args...)
    return (
        new_obj,
        new_obj,
        StructDiff{type}(argdiffs)
    )
end

struct GetField{type, fieldname, valtype} <: CustomUpdateGF{valtype, Nothing}
    function GetField{type, fieldname}() where {type, fieldname}
        return new{type, fieldname, fieldtype(type, fieldname)}()
    end
end
GetField(type, fieldname) = GetField{type, fieldname}()

[...]

PR #434, opened 2021-07-14
Change hint propagation for user-defined structs.

ztangent:
Finishes the implementation of `Construct` and `GetField` started by @georgematheos in #367. Also adds tests, and some logic to handle cases where types have non-default constructors (although this is not exhaustive, because we can't prevent users from writing pathological constructors).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -28,7 +28,7 @@ MacroTools = "0.5"
 Parameters = "0.12"
 ReverseDiff = "1.4, 1.5"
 SpecialFunctions = "0.8, 0.9, 0.10, 1"
-julia = "1"
+julia = "1.3"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
--- a/docs/src/ref/extending.md
+++ b/docs/src/ref/extending.md
@@ -97,6 +97,55 @@ apply_with_state
 update_with_state
 ```
 
+### Incremental computation for user-defined types
+
+Gen provides the generative functions [`Construct`](@ref) and
+[`GetField`](@ref) to support incremental computation for user-defined composite
+types, i.e., structs. They are useful when structs contain fields which are
+returned by other generative functions, or sampled from distributions. For
+example, we might define a struct representing 2D points:
+```julia
+struct Point2f0
+    x::Float64
+    y::Float64
+end
+```
+We might then use [`Construct`](@ref) and [`GetField`](@ref) within the
+following model of point motion:
+```julia
+@gen (static) function point_motion()
+    x ~ normal(0, 1)
+    y ~ normal(0, 1)
+    p ~ Construct(Point2f0)(x, y)
+    new_x ~ x_motion(p)
+    new_y ~ y_motion(p)
+end
+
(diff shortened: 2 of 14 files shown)


ztangent:
@marcoct Any thoughts on how / where we should document these changes? I can't think of a natural place it fits in the docs.

marcoct:
@ztangent I think [Extending Gen](https://github.com/probcomp/Gen.jl/blob/master/docs/src/ref/extending.md) would make sense.

marcoct:
@ztangent This looks great! My only comments are:

- Adding documentation

- Is there a simple description for what subset of struct and constructor combinations this feature works on? If the description is complex (maybe it is not?) I wonder if it might make sense to make the definition more strict and reject other cases?

I'll also create another issue for adding support for AD through these structs.

ztangent:
> Is there a simple description for what subset of struct and constructor combinations this feature works on? If the description is complex (maybe it is not?) I wonder if it might make sense to make the definition more strict and reject other cases?

This feature works on any structs which can be "constructed-by-field", i.e. if the struct is defined as:
```julia
struct Foo
    field1
    field2
    field3
    ...
end
```
then in order to propagate per-field changes, `Construct` should be used with a constructor `foo = Foo(arg1, arg2, arg3, ...)` such that `foo.field1 == arg1`, `foo.field2 == arg2`, and so on. The default constructor of a user-defined type adheres to this property. 

`Construct` also tries to detect if it is being used with a compatible constructor by checking the number and type signature of the arguments it is called with. In cases where a non-compatible constructor can be detected (e.g. the number of arguments is different from the number of fields, or the argument types cannot be `convert`ed to the corresponding field types), any changes to one of the inputs always results in an `UnknownChange()`, which is safe behavior. But there are pathological constructors that meet the type signature requirements, but are not field-constructable, e.g.:

```julia
struct EvilStruct
    x::Int
    y::Int
    EvilStruct(x, y) = new(y, x)
end
```

Unfortunately, it doesn't seem like there's any way to detect cases like this without looking up the IR of the constructor and doing some fancy analysis. So type-based solution I came up with is "complete but unsound", in that it will work without complaint for any constructor, but for some constructors changes will be propagated inaccurately. I would prefer a "sound but incomplete" approach, but as of now, I think the only way to do that is to always return "UnknownChange", because there's no way to distinguish constructor like `EvilStruct(x, y) = new(y, x)` from a default constructor without inspecting IR.

It'd be great if you have thoughts on the best approach given these constraints @marcoct !

ztangent:
Added documentation [here](https://github.com/probcomp/Gen.jl/blob/0a037c74a53e08025c0be407520e171949be82ca/docs/src/ref/extending.md#incremental-computation-for-user-defined-types)! Here's how I've stated the scope of `Construct` for now. It technically works safely (though not usefully) on more constructors, but I suppose it's best to just tell users what constructors they should stick to:

> Note that that Construct should be used with constructors `T(args...)` where the ``n``th argument corresponds to the ``n``th field of the type `T`. Default constructors meet this requirement. Other constructors are not guaranteed to propagate changes correctly.

PR #446, opened 2022-01-11
Use newly rendered updated tutorials

alex-lew:
This pull request includes rendered versions of the newly updated tutorials from https://github.com/probcomp/gen-quickstart/pull/84, as links on the Tutorials page of the website. (Note that it is a PR into the `gh-pages` branch, and not the main branch.)

I have not done any real post-processing on the tutorials, so they may not be the most aesthetically pleasing. Additional clean-up -- of the markdown or the visualizations -- could help.

Another thing to note is that Julia's RNG seems to have changed between versions, so old examples that were 'fixed in place' by a particular random seed are now based on newly generated traces.


alex-lew:
@marcoct One thing to note about the static modeling language notebook. The final experiment is intended to demonstrate constant-factor speedup by switching the _inner_ generative function in a linear regression model (the kernel of a `Map` combinator) to use the static DSL. However, at least on my machine using Julia 1.7.1, the constant factors appear better for the dynamic DSL than for the static one. 

I added a note to this effect (https://github.com/probcomp/Gen.jl/blame/d86a3dbefe45b5a91e3225e2ea00cc53edf7b178/tutorials/scaling-with-combinators-new/tutorial.md#L431). But I left the old version in the repo, with folder title `scaling-with-combinators`, in case we want to switch back. (It's also probably worth replicating this result on another machine at some point.)

marcoct (approved):
Beautiful! Within the notebooks some of the links to headings within a notebook or to other notebooks seem broken, but this can be a separate PR.

Also -- if you have an hour or two, now might be a good time to split the website from the Gen repository (i.e. not render the website on `gh-pages` branch). The documentation generation will still probably be in `gh-pages`, but the gen.dev Jekyll website that includes the tutorials should be a separate repo. Each time we update the tutorials we are adding SVGs and GIFs and this will bloat the repository significantly over time. (It is already a bit bloated, but the GIFs have me wondering if this the best time to make this change).

alex-lew:
@marcoct That does sound like a good idea. I looked briefly at

https://tink.uk/using-a-custom-domain-with-multiple-github-repositories/

and it seems like I might need to tinker with our DNS settings to support multiple repositories all publishing to gen.dev. (i.e., the docs and the 'main page'). I'm not sure I have access to our DNS, and I don't know if I'll have the time this week to debug.

Since the GIF files are already in this branch, have we already bloated the repo? Wondering whether it's worth holding off until we can investigate this further, or whether we should merge now and then revisit. (In the latter case, maybe we can purge the large [~3MB] tutorial files when the time comes. It seems a bit thorny, but [there are ways](https://stackoverflow.com/questions/2100907/how-to-remove-delete-a-large-file-from-commit-history-in-the-git-repository) to remove large files from Git histories...)

marcoct:
> and it seems like I might need to tinker with our DNS settings to support multiple repositories all publishing to gen.dev. (i.e., the docs and the 'main page'). I'm not sure I have access to our DNS, and I don't know if I'll have the time this week to debug.

@alex-lew We can publish the docs to one site (e.g. probcomp.github.io/gen-docs/) from the Gen.jl repository, and we can publish the gen.dev site from another repository (e.g. gen-website), and use a static link from gen.dev to the docs site right?

marcoct:
> Since the GIF files are already in this branch, have we already bloated the repo?

@alex-lew Yes, I was thinking we would remove the files from the history.

PR #448, opened 2022-02-22
Fix Unfold propagating gradients to default parameters.

femtomc:
This should correctly accumulate gradients for default parameters in each `Unfold` chain iteration.

Gradients are collected at each iteration, and then accumulated -- parameters with `no_grad` are ignored (and return `nothing` for the gradient).


femtomc:
Re -- it looks like this failed test occurs irrespective of the changes from this PR. I'm not sure the source of this issue (in the `static_ir`).

alex-lew:
I think this looks good to me! To summarize the issue:

* When we write `bar = Unfold(foo)`, we expect `foo`'s arguments to have the form `(current_state, additional_params...)`. Then `bar`'s first argument is an integer (how many iterations to run), its second argument is an initial state, and its remaining arguments are the `additional_params` of `foo`, kept the same at each iteration.
* The GFI methods `choice_gradients` and `accumulate_param_gradients!` both return gradients of a loss function (the log density of the trace, plus a user-defined 'rest of the loss', denoted `J` in the docs) with respect to the arguments.
*  The current code behaves as though `bar` accepted a *vector* for each additional_argument, specifying the value it should take at each iteration. But in fact, `bar` accepts a single copy of each additional_argument, which it duplicates for each call.
* The gradient of the loss with respect to that single argument is just the sum of the gradients with respect to all the 'copies'.

Is that right?

femtomc:
That’s correct! I think perhaps this was just originally copied from Map, and never tested — so it slipped into the codebase and hasn’t been re-visited.

ztangent:
> The gradient of the loss with respect to that single argument is just the sum of the gradients with respect to all the 'copies'.

This is the bit that isn't obvious to me. Is it because the log density of the trace of `bar = Unfold(foo)` is the sum of the log densities for each application of `foo` across time? And does this apply to the rest of the loss `J` as well?

The thing that's confusing for me here is that the return-value of `Unfold` is vector-valued. And hence `retgrad` is vector-valued as well. Does the matrix multiplication work out such that the gradient with respect to `additional_params` ends up being scalar-valued?

ztangent:
Okay, I think I worked it out in my head! Will add some math in a bit but I think we're good to merge.

femtomc:
I think the key thing is that the fact that retgrad is vector valued doesn’t imply that param grads must or must not be a vector.

I’m thinking of retgrad as gradients out of an intermediate layer like a neural network — just because there’s a vector of gradients for one part of the network doesn’t imply that a scalar parameter gets a vector gradient. That can only occur if the loss is vector valued.

as far as I can tell, we don’t really consider that case with the gradient API — and that requires more complex AD machinery.

ztangent (approved):

alex-lew:
It looks like this bug (and fix) were discussed here: https://github.com/probcomp/Gen.jl/pull/343#discussion_r576338990

@femtomc Can we re-enable those tests? Do they exercise this code path? If not, is there another test we could add to catch regressions?

Thanks!

femtomc:
Great, I'll re-enable.

Edit: they do exercise this path, sorry.

alex-lew:
Great, looks good to me!

PR #453, opened 2022-03-02
Remove Dockerfile and Docker tests (fix #451)

### LINKED ISSUE #451: Docker build runs latest version of Julia instead of intended version
[Based on this Travis log](https://app.travis-ci.com/github/probcomp/Gen.jl/jobs/560758605), it appears that the Docker build uses the latest version of Julia (1.7 as of the time of filing this issue) instead of the intended version (1.4, for the linked log). I'm not an expert on Docker, so maybe someone else more familiar could fix this?

alex-lew:
As per discussion on #451, this removes the Dockerfile, and removes the "Docker" stage of the automated Travis tests.

(A typical Gen install no longer requires any special setup, beyond installing Julia; as such, Docker is overkill. Using GenPyTorch and other libraries that involve Python interop are more involved, but our Docker build wasn't testing that functionality anyway.)


ztangent (approved):
Looks great to me! :)

PR #456, opened 2022-03-03
Update Travis config

alex-lew:
This PR does two things:

1. Changes the versions of Julia we test, from 1.4/1.5/1.6, to 1.3/1.6/nightly. Julia 1.3 is the oldest version we support, Julia 1.6 is the LTS version, and testing on nightly should alert us to any breaking changes (thanks @femtomc for the suggestion).
2. Removes the explicit 'test' build stage, which was running an duplicated suite of tests on Julia 1.4. Thanks @ztangent for finding other projects' `.travis.yml` files that we could compare to, to debug this issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,12 +6,10 @@ julia:
   - 1.4
   - 1.5
   - 1.6
+  - nightly
 
 jobs:
   include:
-    - stage: test
-      script:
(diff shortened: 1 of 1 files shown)


ztangent:
Looks great! Will merge :)

PR #459, opened 2022-03-09
fix(trie): implement get_values_shallow and get_submaps_shallow for trie

### LINKED ISSUE #444: `Gen.Trie` does not implement `get_values_shallow`, despite being a `ChoiceMap`
The `Gen.Trie` data structure is declared as an instance of the `ChoiceMap` abstract data type (https://github.com/probcomp/Gen.jl/blob/44f5f1408caf05ff95932bb58d9f57907c5084e3/src/trie.jl#L5).

However, there is no implementation of `Gen.get_values_shallow` for it.

One place this comes up is when executing code that returns a trace in an environment like Atom, VS Code, or Pluto.jl. These IDEs allow users to click into compound data and see pretty-printed representations of each component. Because Tries are ChoiceMaps, printing `trace.trie` triggers a call to our ChoiceMap pretty printer (https://github.com/probcomp/Gen.jl/blob/master/src/choice_map.jl#L122), which in turn attempts to call `get_values_shallow` (yielding a Method Not Implemented error).

We could fix either by implementing Gen.get_values_shallow for Tries, or by removing the declaration that a Trie is a ChoiceMap (I have not investigated / thought about the consequences of that option, and am not sure off the top of my head whether a Trie being a ChoiceMap is important elsewhere in the codebase).

chentoast:
fixes #444. I'm not that familiar with the codebase yet, so please let me know if the changes should go somewhere else, or if this is not the correct fix. Thanks!

### DIFF
--- a/src/trie.jl
+++ b/src/trie.jl
@@ -2,7 +2,7 @@
 # Trie #
 ##################
 
-struct Trie{K,V} <: ChoiceMap
+struct Trie{K,V}
     leaf_nodes::Dict{K,V}
     internal_nodes::Dict{K,Trie{K,V}}
 end


femtomc:
Re: @chentau -- as per #444, do you mind removing the implementations and the corresponding annotation on `Trie` (?) I don't want to prevent you from having a PR merged (having already put work into it!) -- so I'd like to give you the option to make these changes on your PR.

If you don't have time to do this, I can do so and push some commits to your PR.

femtomc:
Thank you Tony!

PR #460, opened 2022-03-16
feat(inference): add support for custom resamplers in particle filtering

chentoast:
The current PF implementation uses multinomial resampling, which is not really ideal. Add support for custom resampling algorithms like systematic or residual sampling in `maybe_resample!`.

Side note: is there a reason why we save `state.parents` and `state.new_trace`, instead of just assigning them as local variables? I don't really see them used anywhere else.

### DIFF
--- a/src/inference/particle_filter.jl
+++ b/src/inference/particle_filter.jl
@@ -218,5 +218,40 @@ function maybe_resample!(state::ParticleFilterState{U};
     return do_resample
 end
 
+"""
+    did_resample::Bool = maybe_resample!(state::ParticleFilterState, resampler::Function;
+        ess_threshold::Float64=length(state.traces)/2, verbose=false)
+
+Do a resampling step if the effective sample size is below the given threshold.
+{resampler} is a function that takes in a vector of log normalized weights corresponding to each particle,
+and returns a vector of sampled particle indices.
+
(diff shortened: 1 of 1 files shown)


femtomc:
Hi @chentau -- we actually have a package for this: https://github.com/probcomp/GenParticleFilters.jl

which includes extended forms of resampling.

Feel free to add a PR there, if there is anything missing that you want to implement.

chentoast:
Thanks for letting me know :)

femtomc:
Re @chentau -- @ztangent may know better about `state.parents` and `state.new_traces`. It's true that you can allocate them as local variables in the implementation that you're describing.

One thing you can do is pre-allocate arrays for both when you instantiate `ParticleFilterState` -- which is often a good performance idiom. With the state carried throughout inference, the GC will avoid collecting these (e.g. if you don't explicitly write to stack allocate).

That's one reason I can see.

ztangent:
Yes, I suspect avoiding GC might be an important reason! I actually continued using `state.parents` and `state.new_traces` in `GenParticleFilters` because that was what @marcoct was doing, and he may have more insight as to why.

Also, @chentau, glad to know that others are finding a use case for custom resampling methods! I hope the ones implement in GenParticleFilters are sufficient, and if not, feel free to make a PR there :)

PR #461, opened 2022-03-31
Fix `esc` bug with kernel auxiliary arguments in kernel DSL.

femtomc:


femtomc:
One other comment: `postwalk` was actually walking into `tr ~ k(tr, args...)` twice, applying `Gen.reversal` two times.

Would like to add tests for the DSL somehow.

src/inference/kernel_dsl.jl
```diff
@@ -25,78 +26,74 @@ The first argument of the function should be a trace, and the return value of th
...
-        Gen.check_is_kernel(::typeof($(esc(f)))) = true
-    end
+    @capture(ex, function f_(args__) body_ end) || error("expected a function")
+    esc(quote
+            function $(f)($(args...))
+                $body
+            end
+            Gen.is_custom_primitive_kernel(::typeof($(f))) = true
+            Gen.check_is_kernel(::typeof($(f))) = true
+        end)
```
ztangent: If you're wrapping an `esc` around a reference to `Gen`, you may want to check if it works in cases where `Gen` is imported under a different name, or imported as a submodule of another module. I think in cases like these, I often do replace `Gen.is_custom_primitive_kernel` with `$(GlobalRef(Gen, :is_custom_primitive_kernel))`.

ztangent: I think maybe a more concise way to achieve things here is:
```julia
    quote
        esc(ex)
        Gen.is_custom_primitive_kernel(::typeof($(esc(f)))) = true
        Gen.check_is_kernel(::typeof($(esc(f)))) = true
    end
```

because `ex` is just the original function expression.

src/inference/kernel_dsl.jl
```diff
@@ -141,25 +135,22 @@ Declare that two primitive stationary kernels are reversals of one another.
...
+    @capture(ex, k_ : l_) || error("expected a pair of functions")
     quote
-        Gen.is_custom_primitive_kernel($(esc(k))) || error("first function is not a custom primitive kernel")
-        Gen.is_custom_primitive_kernel($(esc(l))) || error("second function is not a custom primitive kernel")
-        Gen.reversal(::typeof($(esc(k)))) = $(esc(l))
-        Gen.reversal(::typeof($(esc(l)))) = $(esc(k))
+        Gen.is_custom_primitive_kernel($(k)) || error("first function is not a custom primitive kernel")
+        Gen.is_custom_primitive_kernel($(l)) || error("second function is not a custom primitive kernel")
+        Gen.reversal(::typeof($(k))) = $(l)
+        Gen.reversal(::typeof($(l))) = $(k)
```
ztangent: I'm confused why the `esc`s are being removed here. I think they should stay...? What's the result of calling `@macroexpand` on the version without the `esc`s?

src/inference/kernel_dsl.jl
```diff
@@ -209,18 +202,21 @@ The resulting object is a Julia function that is annotated as a composite MCMC k
...
         $rev_kern_ex
 
         # bind the reversals for both
-        Gen.reversal(::typeof($(esc(kern)))) = $(esc(rev_kern))
-        Gen.reversal(::typeof($(esc(rev_kern)))) = $(esc(kern))
+        Gen.reversal(::typeof($(kern))) = $(rev_kern)
+        Gen.reversal(::typeof($(rev_kern))) = $(kern)
     end
+    expr = postwalk(flatten ∘ unblock ∘ rmlines, expr)
+    esc(expr)
```
ztangent: Again, I'm not sure whether `esc`aping a reference to `Gen` will do the right thing in all scenarios. I think we should just return:

```julia
    expr = quote
        # define forward kerel
        $(esc(kern_ex))

        # define reversal kernel
        $(esc(rev_kern_ex))

        # bind the reversals for both
        Gen.reversal(::typeof($(esc(kern)))) = $(esc(rev_kern))
        Gen.reversal(::typeof($(esc(rev_kern)))) = $(esc(kern))
  end
```

src/inference/kernel_dsl.jl
```diff
@@ -25,78 +26,74 @@ The first argument of the function should be a trace, and the return value of th
...
-        elseif MacroTools.@capture(x, T_ ~ k_(T_, args__))
+        elseif @capture(x, T_ ~ k_(T_, args__))
 
             # applying a kernel
             quote
-                check && Gen.check_is_kernel($(esc(k)))
-                $(esc(T)) = $(esc(k))(
-                    $(esc(T)), $(map(esc, args)...),
-                    check=check, observations=observations)[1]
+                check && Gen.check_is_kernel($(k))
```
ztangent, line 94: Replace Gen.check_is_kernel with GlobalRef.

src/inference/kernel_dsl.jl
```diff
@@ -107,19 +104,16 @@ function expand_kern_ex(ex)
...
+                      check=false, observations=EmptyChoiceMap())
             $body
-            check && check_observations(get_choices($(esc(trace))), observations)
+            check && check_observations(get_choices($(trace)), observations)
             metadata = nothing
-            ($(esc(trace)), metadata)
+            ($(trace), metadata)
         end
-        Gen.check_is_kernel(::typeof($(esc(f)))) = true
+        Gen.check_is_kernel(::typeof($(f))) = true
```
ztangent: Replace Gen.check_is_kernel with GlobalRef.

src/inference/kernel_dsl.jl
```diff
@@ -107,19 +104,16 @@ function expand_kern_ex(ex)
...
 
     ex = quote
-        function $(esc(f))(
-                $(esc(trace))::Trace, $(toplevel_args...);
-                check=false, observations=EmptyChoiceMap())
+        function $(f)($(trace)::Trace, $(toplevel_args...);
+                      check=false, observations=EmptyChoiceMap())
             $body
-            check && check_observations(get_choices($(esc(trace))), observations)
+            check && check_observations(get_choices($(trace)), observations)
```
ztangent: If we escape the reference to `check_observations`, there's a chance it will end up referring to a different function defined by the user called `check_observations` in whatever module `@kern` gets called. I'm not sure which module `check_observations` comes from, but it should probably be replaced with a GlobalRef.

src/inference/kernel_dsl.jl
```diff
@@ -168,11 +159,13 @@ function reversal_ex(ex)
...
 
-        elseif MacroTools.@capture(x, T_ ~ k_(T_, args__))
+        elseif @capture(x, T_ ~ k_(T_, args__))
 
             # applying a kernel - apply the reverse kernel
             quote
-                $T ~ (Gen.reversal($k))($T, $(args...))
+                check && Gen.check_is_kernel(Gen.reversal($k))
+                $(T) = Gen.reversal($k)($(T), $(args...),
+                            check=check, observations=observations)[1]
```
ztangent, line 168: Replace references to Gen functions with GlobalRefs.

src/inference/kernel_dsl.jl
```diff
@@ -107,19 +104,16 @@ function expand_kern_ex(ex)
     end
 
     ex = quote
-        function $(esc(f))(
-                $(esc(trace))::Trace, $(toplevel_args...);
-                check=false, observations=EmptyChoiceMap())
+        function $(f)($(trace)::Trace, $(toplevel_args...);
+                      check=false, observations=EmptyChoiceMap())
```
ztangent, line 108: We might want to replace `check` and `observations` with `gensym`-ed versions just to be safe, and make sure we use the `gensym`-ed versions throughout the rest of the expression --- otherwise, there's a (small) chance that the `check` and `observations` names we introduced will clash with the names of the arguments that the user writes when defining the kernel. Normally macro expansion would handle the gensyms for us, but if we are escaping everything, then `check` will remain as `check`.

ztangent: Actually, I've changed my mind about this --- what I said above makes sense for positional arguments, which I believe are automatically gen-symed by macro-expansion. But I don't think it holds true for keyword arguments, since they ought to have the same name after macro-expansion.

ztangent (changes requested):
I think this is mostly making the right fixes, but we have to be careful about how `esc`aping the final `Expr` affects name resolution for names and references we're injecting into the generated code. 

For example, if you escape a block of code which contains a line like `Gen.reversal(...)`, `Gen.reversal` may not always end up pointing to the right function, depending on what `Gen` refers to in the context where the macro is called. To avoid this, we should either avoid escaping lines like that, or if easier, replace `Gen.reversal` with `$(GlobalRef(Gen, :reversal))`.

src/inference/kernel_dsl.jl
ztangent: Yup, confirmed the above about how positional arguments vs keyword arguments are handled by macro-expansion. 

```julia
macro make_f(arg)
    quote
        function f(a, b, $(esc(arg)); c=1, d=2)
            a + b + c + d + $(esc(arg))
        end
    end
end

julia> @macroexpand @make_f(c)
quote
    #= none:3 =#
    function var"#27#f"(var"#28#a", var"#29#b", c; c = 1, d = 2)
        #= none:3 =#
        #= none:4 =#
        var"#28#a" + var"#29#b" + c + d + c
    end
end

julia> @make_f(c)
ERROR: syntax: function argument name not unique: "c"
```

So there was always a chance something bad would happen if the user named a positional argument the same thing as one of the keyword arguments, and if we escaped the user-provided argument names.

ztangent: This makes me think the best thing to do here is actually `esc` as few things as possible! So remove the `esc` of the final `Expr`. The only thing we need to be escaping are the user-provided names of the kernels themselves, because that's how we'll refer to them afterwards. We don't have to escape the user-provided argument names, either within the function header or within the function body --- macro-expansion should automatically do the right thing for us by gen-syming or not gen-syming them appropriately.

femtomc: Good analysis — I’ll implement these changes.

femtomc: > I think this is mostly making the right fixes, but we have to be careful about how `esc`aping the final `Expr` affects name resolution for names and references we're injecting into the generated code.
> 
> For example, if you escape a block of code which contains a line like `Gen.reversal(...)`, `Gen.reversal` may not always end up pointing to the right function, depending on what `Gen` refers to in the context where the macro is called. To avoid this, we should either avoid escaping lines like that, or if easier, replace `Gen.reversal` with `$(GlobalRef(Gen, :reversal))`.

I think in this context, more things are going to go wrong if the user writes `import Gen as G` and then tries to use the kernel DSL -- in fact, in that case, we need to annotate every single type with `G` (because `import Gen as G` does not define `Trace`, for example, in the scope of the import).

femtomc: I think annotating everything with `GlobalRef` is not necessary. I think it's overly conservative -- e.g. a lot of the functionality of Gen assumes the user writes `using Gen` _somewhere_ in their development module.

Given that we make that assumption elsewhere in the codebase, I don't see any reason to fix it here.

femtomc: Re -- I implemented `gensym` versions of the keyword arguments, and made these consistent in both the forward and reversal kernel.

femtomc:
Also, here's a reference expansion:

```julia
quote
    function myk(tr::Trace; var"##check#282" = false, var"##observations#283" = EmptyChoiceMap())
        var"##check#282" && Gen.check_is_kernel(mh)
        tr = (mh(tr, select(:a), var"##check#282" = var"##check#282", var"##observations#283" = var"##observations#283"))[1]
        var"##check#282" && Gen.check_observations(get_choices(tr), var"##observations#283")
        metadata = nothing
        (tr, metadata)
    end
    Gen.check_is_kernel(::typeof(myk)) = true
    function var"##reverse_kernel#284"(tr::Trace; var"##check#282" = false, var"##observations#283" = EmptyChoiceMap())
        var"##check#282" && Gen.check_is_kernel(Gen.reversal(mh))
        tr = ((Gen.reversal(mh))(tr, select(:a), var"##check#282" = var"##check#282", var"##observations#283" = var"##observations#283"))[1]
        var"##check#282" && Gen.check_observations(get_choices(tr), var"##observations#283")
        metadata = nothing
        (tr, metadata)
    end
    Gen.check_is_kernel(::typeof(var"##reverse_kernel#284")) = true
    Gen.reversal(::typeof(myk)) = var"##reverse_kernel#284"
    Gen.reversal(::typeof(var"##reverse_kernel#284")) = myk
end
```

femtomc:
Ah, @ztangent -- we can't `gensym` the keyword arguments, because we assume `check` and `observations` throughout the Gen inference library (for MCMC kernels) :(

femtomc:
Okay, I removed outermost `esc` -- now only escaping kernel + reversal names, and macro expansion takes care of gensyming the argument names.

src/inference/kernel_dsl.jl
ztangent: If we don't `esc` the final returned expression, then we don't have to do the `GlobalRef` annotation, so that's fine!

We do use the `GlobalRef` approach in other parts of the code though, particularly in codegen for the static modeling language -- see Issue #272 and the corresponding fix PR #273.

ztangent (approved):
Looks great to me! Thanks for doing this :)

PR #462, opened 2022-04-13
feat(inference): add nuts sampler

chentoast:
work in progress. implements a no u turn sampler but does not implement a dual averaging adaption process for setting epsilon, which can be done as follow up work. code generally needs to be cleaned up a little as well

todo:
- [x] docs
- [x] tests
- [x] calculate if a trajectory has diverged
- [x] add all sampler statistics necessary for dual averaging

### DIFF
--- a/src/inference/hmc_common.jl
+++ b/src/inference/hmc_common.jl
@@ -1,0 +1,39 @@
+function sample_momenta(n::Int)
+  Float64[random(normal, 0, 1) for _=1:n]
+end
+
+function assess_momenta(momenta)
+  logprob = 0.
+  for val in momenta
+      logprob += logpdf(normal, val, 0, 1)
+  end
+  logprob
+end
+
+function add_choicemaps(a::ChoiceMap, b::ChoiceMap)
+  out = choicemap()
+
+  for (name, val) in get_values_shallow(a)
+    out[name] = val + b[name]
+  end
+
+  for (name, submap) in get_submaps_shallow(a)
+    out.internal_nodes[name] = add_choicemaps(submap, get_submap(b, name))
+  end
+
+  return out
+end
+
+function scale_choicemap(a::ChoiceMap, scale)
+  out = choicemap()
+
+  for (name, val) in get_values_shallow(a)
+    out[name] = val * scale
+  end
+
+  for (name, submap) in get_submaps_shallow(a)
+    out.internal_nodes[name] = scale_choicemap(submap, scale)
+  end
+
+  return out
+end
--- a/src/inference/nuts.jl
+++ b/src/inference/nuts.jl
@@ -1,0 +1,179 @@
+using LinearAlgebra: dot
+
+struct Tree
+  val_left :: ChoiceMap
+  momenta_left :: ChoiceMap
+  val_right :: ChoiceMap
+  momenta_right :: ChoiceMap
+  val_sample :: ChoiceMap
+  n :: Int
+  weight :: Float64
+  stop :: Bool
+  diverging :: Bool
+end
+
+function u_turn(values_left, values_right, momenta_left, momenta_right)
+  return (dot(values_left - values_right, momenta_right) >= 0) &&
+  (dot(values_right - values_left, momenta_left) >= 0)
+end
+
+function leapfrog(values_trie, momenta_trie, eps, integrator_state)
+  selection, retval_grad, trace = integrator_state
+
+  (trace, _, _) = update(trace, values_trie)
+  (_, _, gradient_trie) = choice_gradients(trace, selection, retval_grad)
+
+  # half step on momenta
+  momenta_trie = add_choicemaps(momenta_trie, scale_choicemap(gradient_trie, eps / 2))
+
+  # full step on positions
+  values_trie = add_choicemaps(values_trie, scale_choicemap(momenta_trie, eps))
+
+  # get new gradient
+  (trace, _, _) = update(trace, values_trie)
+  (_, _, gradient_trie) = choice_gradients(trace, selection, retval_grad)
+
+  # half step on momenta
+  momenta_trie = add_choicemaps(momenta_trie, scale_choicemap(gradient_trie, eps / 2))
+  return values_trie, momenta_trie, get_score(trace)
+end
+
+function build_root(val, momenta, eps, direction, weight_init, integrator_state)
+  val, momenta, lp = leapfrog(val, momenta, direction * eps, integrator_state)
+  weight = lp + assess_momenta(to_array(momenta, Float64))
+
+  diverging = weight - weight_init > 1000
+
+  return Tree(val, momenta, val, momenta, val, 1, weight, false, diverging)
+end
+
+function merge_trees(tree_left, tree_right)
+  # multinomial sampling
+  if log(rand()) < tree_right.weight - tree_left.weight
+    sample = tree_right.val_sample
+  else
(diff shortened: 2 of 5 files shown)


femtomc:
Cool Tony! looks interesting -- I converted this to a draft PR, when you're ready to put it in the pipeline, you should be able to convert it back.

chentoast:
ping @femtomc :)

femtomc:
@chentoast Hi! I'll likely get to take a look at this sometime next week -- thank you for your work on this PR!

src/inference/hmc_common.jl
```diff
@@ -0,0 +1,52 @@
...
+  end
+  logprob
+end
+
+function add_choicemaps(a::ChoiceMap, b::ChoiceMap)
+  out = choicemap()
+
+  for (name, val) in get_values_shallow(a)
+    out[name] = val + b[name]
+  end
```
femtomc, line 18: I'm not sure how this function is intended to work if the choice maps have different shapes. Should we treat missing addresses in one choice map as "zero" for the value type of the choice map which does have the address?

It's unclear to me what the correct thing is here -- or whether you'd ever be adding choice maps with different address shapes.

femtomc: If this invariant is enforced implicitly, it would be nice to document the code to indicate that.

src/inference/nuts.jl
```diff
@@ -0,0 +1,191 @@
...
+      new_momenta_score = assess_momenta_trie(tree.momenta_left)
+    end
+
+    # accept or reject
+    alpha = new_model_score + new_momenta_score - weight_init
+    if log(rand()) < alpha
+      return (new_trace, SamplerStats(depth, tree.n, tree.diverging, true))
+    else
+      return (trace, SamplerStats(depth, tree.n, tree.diverging, false))
+    end
```
femtomc, line 175: One small comment -- here, the normal interface for kernels is slightly violated (but in a way which I'm not unhappy about). I think normally, primitive kernels have a return signature of `(Trace, accept_or_reject)`.

One interesting idea (to improve the information we can extract from our kernels) would be to create a new abstract type for the second element of the returned tuple, and allow definition of new types to be returned from kernels.

This change is a bit more drastic -- so I recommend that (for now), we remove `SamplerStats` from this PR -- and setup a new issue to handle the bigger question of how to better profile the kernel statistics.

src/inference/nuts.jl
```diff
@@ -0,0 +1,191 @@
...
+  val_left
+  momenta_left
+  val_right
+  momenta_right
+  val_sample
+  n :: Int
+  weight :: Float64
+  stop :: Bool
+  diverging :: Bool
+end
```
femtomc, line 13: For a structure like this (which I assume has recursive references to instances of itself) -- do you mind adding documentation to indicate this?

Alternatively, you might use an abstract type to directly indicate the recursion.

src/inference/hmc_common.jl
```diff
@@ -0,0 +1,52 @@
...
+  for (_, val) in get_values_shallow(momenta_trie)
+    logprob += logpdf(normal, val, 0, 1)
+  end
+
+  for (_, submap) in get_submaps_shallow(momenta_trie)
+    logprob += assess_momenta_trie(submap)
+  end
+
+  return logprob
+end
```
femtomc: Depending on usage below, it might be useful to optimize this operation by assuming you pass in a `Vector` -- a flattened version of `momenta_trie` to avoid the lookup overhead of recursively following references. That way, the logprob could be computed by evaluating the logprob of a single isotropic MvNormal.

src/inference/hmc_common.jl
```diff
@@ -0,0 +1,52 @@
...
+  for (name, val) in get_values_shallow(a)
+    out[name] = val * scale
+  end
+
+  for (name, submap) in get_submaps_shallow(a)
+    out.internal_nodes[name] = scale_choicemap(submap, scale)
+  end
+
+  return out
+end
```
femtomc, line 39: I don't really like that this allocates a new choice map each time you call it -- but I need to examine and think about the code below further to suggest a better solution.

Even if we created a view-like structure, we'd still need to eat the allocations when we decide to modify (scale) the values.

src/inference/nuts.jl
```diff
@@ -0,0 +1,191 @@
...
+
+struct SamplerStats
+  depth
+  n
+  diverging
+  accept
+end
+
```
femtomc: Similar here -- type annotations would be appreciated.

src/inference/nuts.jl
```diff
@@ -0,0 +1,191 @@
...
+  end
+
+  weight = logsumexp(tree_left.weight, tree_right.weight)
+  n = tree_left.n + tree_right.n
+  
+  if u_turn(to_array(tree_left.val_left, Float64),
+    to_array(tree_right.val_right, Float64),
+    to_array(tree_left.momenta_left, Float64),
+    to_array(tree_right.momenta_right, Float64))
+  end
```
femtomc: What is the point of this block?

src/inference/nuts.jl
chentoast: Makes sense - I'll do that in a follow-up PR.

src/inference/nuts.jl
chentoast: Whoops, debugging block that I forgot to remove

src/inference/hmc_common.jl
chentoast: I actually implemented a version of this function that operated in place, but I didn't see noticeable performance benefits (at least, on my tiny toy problem). But I don't see any downsides of making it work in place, at least for now.

src/inference/hmc_common.jl
chentoast: Makes sense

src/inference/hmc_common.jl
chentoast: Hmm, that makes sense. I can't see a way for the invariant to be violated - we are always applying these functions to values and momentum choicemaps that are defined to have the same mistake at the start of the function. Violating that invariant would involve somehow changing the selection while in the middle of a leapfrog step, which doesn't seem to be possible.

chentoast:
This pr is getting closer, but I would hold off on merging until I've had a chance to test it out on some more non-trivial models

hedonhermdev:
Hi, it seems like this PR has been abandoned. Are there plans to merge this? I would like to have the NUTS sampler included in Gen. Willing to pitch in if there's any work left before it's ready to merge.

ztangent:
I think @chentoast hasn't been working on this for a while, so @hedonhermdev you are welcome to pick up where the PR was left off! It'd be great if you could:
- [ ] Check that all the issues from the last PR review appear to be resolved
- [ ] Test that NUTS works on a reasonably-sized problem (e.g. the Bayesian linear regression model in https://www.gen.dev/docs/dev/tutorials/getting_started/), comparing it against the posterior induced by e.g. importance sampling with lots of samples (or exact inference over a grid of possible latent values), and also regular HMC.
- [ ] Add documentation to the docs page for MCMC: https://github.com/probcomp/Gen.jl/blob/master/docs/src/ref/inference/mcmc.md

SamuelBrand1:
Since NUTS is a meta-algorithm that combines (semi) automated discovery of HMC parameters with a tree-based roll out of the HMC then maybe #559 could be a useful stepping stone here?

PR #463, opened 2022-04-14
(Kernel DSL) Revert esc change toplevel, include GlobalRef to handle evaluation ambiguity in modules/submodules.

femtomc:


src/inference/kernel_dsl.jl
```diff
@@ -27,12 +27,14 @@ There should be keyword arguments check and observations.
 """
 macro pkern(ex)
     @capture(ex, function f_(args__) body_ end) || error("expected a function")
+    gr1 = GlobalRef(Gen, :is_custom_primitive_kernel)
+    gr2 = GlobalRef(Gen, :check_is_kernel)
```
ztangent: I would maybe rename `gr1` and `gr2` to `_is_custom_primitive_kernel` and `_check_is_kernel` for greater readability! Same for other places where this pattern is used.

src/inference/kernel_dsl.jl
```diff
@@ -212,11 +212,31 @@ macro kern(ex)
...
+macro kern(ex)
+    expr = toplevel(ex)
+    esc(expr)
+end
+
+macro kern(debug_flag, ex)
+    expr = toplevel(ex)
+    debug_flag == :debug && display(expr)
+    esc(expr)
+end
```
ztangent: If we're going to add this variant of the macro, we should also document it I think! Also, is there a reason why the user would need to use this on the regular, as opposed to using `macroexpand`..?

ztangent (commented):
Looks great, just suggested a few style changes for greater readability!

src/inference/kernel_dsl.jl
femtomc: @ztangent I think `macroexpand` will forcibly expand everything, so it's a question of the "coarseness" of view on `Expr` required for debugging. The displayed `Expr` here will just be the `@kern`, etc level.

I could totally get rid of this, however -- it's just an idiom I sometimes use for debugging macros.

src/inference/kernel_dsl.jl
femtomc: Great, will do.

src/inference/kernel_dsl.jl
femtomc: I just found about `macroexpand` and `@macroexpand1` -- which I think allows you to control the nesting!

src/inference/kernel_dsl.jl
```diff
@@ -13,6 +13,11 @@ end
 is_custom_primitive_kernel(::Function) = false
 check_is_kernel(::Function) = false
 
+_is_custom_primitive_kernel = GlobalRef(Gen, :is_custom_primitive_kernel)
+_check_is_kernel = GlobalRef(Gen, :check_is_kernel)
+_check_observations = GlobalRef(Gen, :check_observations)
+_reversal = GlobalRef(Gen, :reversal)
```
ztangent: If we're doing these as globals, we should probably make them `const` -- it won't affect run-time behavior, but it will probably make the macros run just a little bit faster. Fine either way though!

ztangent (commented):
Just have a minor comment about whether to make the GlobalRefs `const`, since they are global variables now. Fine either way though!

ztangent:
Oh, the other thing we should do is add a test case that makes sure the error you ran into earlier (of using a custom proposal) doesn't happen again!

femtomc:
Re -- I'll address the `const` issue and think up a test case!

ztangent (commented):
Looks great to me! For the test cases, maybe we should also test a kernel that uses a custom proposal, to ensure we don't reintroduce that bug again?

femtomc:
@ztangent I don't think the issue was using a custom proposal in `MH` iirc (?) One of the issues was argument name capture.

Or, do you mean -- capturing the name of the proposal in the kernel at module scope?

I can (and should) easily add the latter.

ztangent:
>  capturing the name of the proposal in the kernel at module scope

Yup, this is what I meant! Because it's the issue that came up when we tried not-escaping everything.

Issue #431, opened 2021-07-13
Exporting `ifelse` has spurious side-effect


vchuravy:
`ifelse` as a name is owned by Base and shouldn't be exported by Gen. 

```
julia> f(x) = ifelse(x, :a, :b)
f (generic function with 1 method)

julia> using Gen

julia> @code_typed f(true)
WARNING: both Gen and Base export "ifelse"; uses of it in module Main must be qualified
CodeInfo(
1 ─ %1 = Main.ifelse::Any
│   %2 = (%1)(x, :a, :b)::Any
└──      return %2
) => Any
```

ztangent:
Resolved by #469.

PR #469, opened 2022-05-26
Unexport `ifelse` to avoid conflicts with `Core.ifelse`.

ztangent:
Per #431, we should avoid exporting `Gen.ifelse` to avoid conflicts with `Core.ifelse`. Having searched through the code for the Gen ecosystem, it looks no tests make use of `Gen.ifelse`, and [only one example in GenExamples.jl](https://github.com/probcomp/GenExamples.jl/blob/6bc4a840fe6d823e008c1c4bf69e67ae1076e18e/regression/static_model.jl) makes use of it with an explicit import. It should thus be safe to unexport.

### DIFF
--- a/src/diff.jl
+++ b/src/diff.jl
@@ -412,5 +412,3 @@ end
 function ifelse(c::Bool, x, y)
     c ? x : y
 end
-
-export ifelse


ztangent:
At some point we may want to consider importing and extending [IfElse](https://github.com/SciML/IfElse.jl) instead, since `Core.ifelse` is not extensible.

ztangent:
@alexlew @femtomc it'd be good to get your brief take on this before I merge!

PR #472, opened 2022-06-15
Eliminate the need for `@load_generated_functions`.

ztangent:
This PR eliminates the need to call `@load_generated_functions` in order to make use of a generative function written in the static modeling language. Instead, new `@generated` GFI methods (`simulate`, `update`, etc.) are defined for each static generative function. These GFI definitions are evaluated after the `get_ir` and `get_options` definitions returned by `Gen.generate_generative_function`, preventing world-age issues.

Besides making the static modeling language easier to use, this change prevents incremental precompilation warnings and breakages which were occurring when multiple sub-packages called `@load_generated_functions`, of the form:

```
WARNING: Method definition simulate(Gen.StaticIRGenerativeFunction{T, U} where U where T, Tuple)
in module TestSubPkgA at C:\...\Gen.jl\src\static_ir\simulate.jl:86 overwritten in module TestSubPkgB
on the same line (check for duplicate calls to `include`).
```

Minimal working example: [TestPkg.zip](https://github.com/probcomp/Gen.jl/files/8910161/TestPkg.zip)

This warning occurs because `@generated function simulate(...)` was being defined multiple times:
1. Once when some dependency`TestSubPkgA` that calls `@load_generated_functions` is precompiled.
2. Again where another dependency `TestSubPkgB` that calls `@load_generated_functions` is precompiled.

The second definition overwrites the first definition for the exact same `@generated` method, leading to precompilation warnings. This PR addresses the issue by defining a *different* `@generated` method for each new generative function.

In addition to the main change:
- All tests have been updated to remove calls to `load_generated_functions` or `@load_generated_functions`.
- Deprecation warnings have been added to the docstrings and function bodies of those methods
- Documentation has been updated to reflect that calling `@load_generated_functions` is no longer necessary.
 
 For documentation and docstrings, we may want to specify a version number, e.g. Gen v0.4.6, for when `load_generated_functions` is no longer necessary. But apart from that, I'm pretty sure this is safe for the rest of the Gen ecosystem!

### DIFF
--- a/docs/src/ref/modeling.md
+++ b/docs/src/ref/modeling.md
@@ -427,7 +427,6 @@ This will produce a file `test.pdf` in the current working directory containing 
 ### Restrictions
 
 First, the definition of a `(static)` generative function is always expected to occur as a [top-level definition](https://docs.julialang.org/en/v1/manual/modules/) (aka global variable); usage in non–top-level scopes is unsupported and may result in incorrect behavior.
-Recall also that the macro [`@load_generated_functions`](@ref) is expected to be called as a top-level expression only.
 
 Next, in order to be able to construct the static graph, Gen restricts the permitted syntax that can be used in functions annotated with `static`.
 In particular, each statement in the body must be one of the following:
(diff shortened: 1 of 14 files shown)


femtomc (approved):
LGTM! A nice straightforward fix!

PR #478, opened 2022-09-09
remove unbound type parameters

nsajko:
Unbound type parameters often cause performance issues and run time dispatch.

Issue found using https://github.com/JuliaLang/julia/pull/46608


fsaad:
Thanks @nsajko, this fix was implemented in https://github.com/probcomp/Gen.jl/pull/481

PR #481, opened 2022-10-31
20221031 fsaad minor fixes

fsaad:
Minor fixes for whitespace, semi-colons, and similar.


femtomc (approved):
These minor fixes look good to me.

I think in general we might want to add a formatter checker to the CI, and add a formatter script to the main repo.

E.g. https://github.com/domluna/JuliaFormatter.jl

We can handle this separately.

Issue #483, opened 2022-10-31
Migrate tests from Travis CI to Github Actions


fsaad:
https://docs.github.com/en/actions/migrating-to-github-actions/migrating-from-travis-ci-to-github-actions

As of now we use Github Actions for CompatHelper.yml under `.github/workflows` and Travis for testing and documentation under `.travis.yml`. We should unify all the status checks under Github Actions; some example config files can be found here:
- https://github.com/JuliaPy/PyPlot.jl/blob/master/.github/workflows/
- https://github.com/SciML/DiffEqDocs.jl/tree/master/.github/workflows

Eliminating Travis will also make it easier to incrementally add more checks, such as Formatting and the like, within the Github interface.

ztangent:
Resolved by #484!

PR #484, opened 2022-10-31
Draft Github CI integrations

fsaad:
Related #483


fsaad:
The ContinuousIntegration actions appears to work correctly.

While the Documentation action works within this PR, I still have not verified that it will correctly push the documentation to gen.dev. I'm not sure that there is a way to easily test this behavior outside of merging this PR into master.

.github/workflows/ContinuousIntegration.yml
```diff
@@ -0,0 +1,30 @@
...
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  ContinuousIntegration:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        julia-version: ['1.3', 'nightly']
```
femtomc: `1.3` is quite old -- most packages in the ecosystem have dropped support for `1.3`. Let's test the latest LTS instead `1.6.7`.

fsaad: I am happy to bump it, but also not sure why the current Travis file tests against 1.3, maybe it is related to how Gen is configured in the official Julia Registry?

https://github.com/JuliaRegistries/General/blob/01717d7c358686ddbbe8ce40d159b6452b870f52/G/Gen/Compat.toml#L33-L35

ztangent: It's not about the Julia Registry, it's that the current version of Gen (v0.4) still officially supports Julia 1.3 according to our `Project.toml` file: https://github.com/probcomp/Gen.jl/blob/master/Project.toml

We could decide to drop support for Julia 1.3 and only support Julia 1.6 onwards, but I think that should be a separate change, and in the meantime I think our CI tests should still test on the versions we're claiming to support!

fsaad: Agreed, let's keep 1.3 for now and handle the upgrade to 1.6 in a future PR.

.github/workflows/Documentation.yml
```diff
@@ -0,0 +1,32 @@
+# Adapted from
+# https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#GitHub-Actions
+#
+# TODO: Do we need to install TagBot.yml for documenting tagged versions?
+# See discussion in the above URL regarding permissions for tagged release.
+
```
fsaad, line 3: It does not seem that we need TagBot.yml, so long as we maintain discipline in manually tagging each Github releas in our repo before adding it to the Julia Registry.

ztangent: Yes, I think that's what we've been doing.

.github/workflows/ContinuousIntegration.yml
ztangent: So it looks for the Travis file, we are currently testing 1.3, 1.6, 1 (latest stable), and nightly -- I think we should do the same for the GitHub CI action.

fsaad: Testing against fewer versions has the advantage that it is faster to get the status checks (where each version can take several minutes) and saves computational resources, especially when the status checks are run against each commit in a PR (which is something we should disable).

For this PR I'm happy to test against the same versions of Travis, and then we can do a separate PR for deciding which versions CI should test against and formally support (e.g., "1.6" and "nightly").

ztangent:
It looks like the CI test for Julia 1.6 is failing [on this test case](https://github.com/probcomp/Gen.jl/blob/b959223ff31a5dd254e5256ae5d83275931b2055/test/inference/particle_filter.jl#L146-L170) due to some platform specific issue leading to different RNG seeds:

```
default proposal: Test Failed at /home/runner/work/Gen.jl/Gen.jl/test/inference/particle_filter.jl:169
  Expression: isapprox(expected_log_ml, actual_log_ml_est, atol = 0.02)
   Evaluated: isapprox(-4.87645083351704, -4.897679480149515; atol = 0.02)
```

In contrast [a recent Travis job for Julia 1.6](https://app.travis-ci.com/github/probcomp/Gen.jl/jobs/587146734/config) passes that test just fine.

We should probably change the failing test to make them more error tolerant in a separate PR. But for the purposes of this PR, it looks like one difference is that previously, the (passing) Travis build was running on an x64 architecture:

```
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Xeon(R) CPU @ 2.80GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, cascadelake)
```

Whereas the failing GitHub CI action is running on a x86 architecture (note the value for `WORD_SIZE`):

```
Platform Info:
  OS: Linux (i686-pc-linux-gnu)
  CPU: Intel(R) Xeon(R) Platinum 8272CL CPU @ 2.60GHz
  WORD_SIZE: 32
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, skylake-avx512)
```

To address this, I think we should try changing the GitHub action to run on `x64` instead, and hope that this fixes the issue. It would also maintain consistency with the previous test suite.

.github/workflows/ContinuousIntegration.yml
```diff
@@ -0,0 +1,30 @@
...
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  ContinuousIntegration:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        julia-version: ['1', '1.3', '1.6', 'nightly']
+        julia-arch: [x86]
```
ztangent: Change to `x64` for now to avoid platform specific RNG issues leading to failed test cases.

fsaad: Good catch, changed now.

ztangent:
Looks like the checks passed! Merging :)

PR #486, opened 2022-11-08
Doc clarification for "regenerate" and "update"

fsaad:
Demystify the returned weight of "update" and "regenerate" as difference.

The current log weight is hard to read, breaking it into a difference of log weights make the spec clearer. (Please double check).


ztangent:
Looks right to me! How do we get the checks to run?

fsaad:
> Looks right to me! How do we get the checks to run?

I'm not sure... maybe https://github.com/probcomp/Gen.jl/pull/484 needs to be merged into `master` first?

~~EDIT: I think pushing a dummy commit will trigger the checks too.~~

EDIT (2): I think pushing the dummy commit did in fact trigger the checks... they are in progress.

PR #492, opened 2022-12-14
Refer to auxiliary randomness r more consistently.

fsaad:
1. Use p(r | t; x) to denote the conditional of p(r, t; x) rather than p(r; x, t).  The current notation does not make it clear that p(r; x, t) must coincide with the proper conditional (r | t; x), according to Bayes rule.

2. Use p(r,t;x) throughout, rather than the current mix of p(t,r;x) and p(r,t;x).


ztangent (approved):
Looks great to me!

PR #495, opened 2022-12-29
feat: add dirichlet distribution

chentoast:
I found it actually kind of nontrivial to test gradients of the dirichlet with finite differences, since the simplex condition means that it's impossible to perturb one element of the simplex vector without having to adjust all of the others. So I just manually checked for a couple of values against pytorch, but better ideas would be very welcome.


test/modeling_library/distributions.jl
```diff
@@ -233,6 +233,34 @@ end
...
+    @test logpdf(dirichlet, [.5, .5], [-1., 1.]) == -Inf
+    @test logpdf(dirichlet, [.5, .5], [-1., 1.]) == -Inf
+    @test logpdf(dirichlet, [0., 1], [1., 1.]) != -Inf
+
+    @test isapprox(logpdf(dirichlet, [.01, .99], [2., 2.]),
+                   Distributions.logpdf(Distributions.Dirichlet([2., 2.]), [.01, .99]))
+    @test isapprox(logpdf(dirichlet, [.01, .99], [1., 4.]),
+                   Distributions.logpdf(Distributions.Dirichlet([1., 4.]), [.01, .99]))
+    @test isapprox(logpdf(dirichlet, [.01, .99], [.01, .01]),
+                   Distributions.logpdf(Distributions.Dirichlet([.01, .01]), [.01, .99]))
```
ztangent, line 255: Can we add tests for Dirichlet distributions over 3 categories as well? Just to make sure it works beyond the case where it's equivalent to a Beta distribution.

test/modeling_library/distributions.jl
```diff
@@ -233,6 +233,34 @@ end
...
+
+    f = (x, alpha) -> logpdf(dirichlet, x, alpha)
+    args = ([.5, .5], [1., 2.])
+    actual = logpdf_grad(dirichlet, args...)
+
+    @test isapprox(actual[2][1], finite_diff_vec(f, args, 2, 1, dx))
+    @test isapprox(actual[2][2], finite_diff_vec(f, args, 2, 2, dx))
```
ztangent: For finite differencing for `N >= 3`, maybe we can write a version of `finite_diff_vec` that maintains the constraint of remaining on the simplex? E.g. if you increase the probability vector at position `i` by `dx`, then decrease the probability vector at positions `j != i` by `dx / N-1`.

Also, am I right to say that these test cases check whether perturbations to `alpha` are resulting in the right gradients? If so, I would either rewrite the code or add comments to make that a little clearer.

src/modeling_library/distributions/dirichlet.jl
```diff
@@ -0,0 +1,42 @@
...
+    end
+end
+
+function logpdf_grad(::Dirichlet, x::AbstractVector{T}, alpha::AbstractVector{U}) where {T <: Real, U <: Real}
+    if isapprox(sum(x), 1) && all(x .>= 0) && all(alpha .>= 0)
+        deriv_x = (alpha .- 1) ./ x
+        deriv_alpha = log.(x) .- digamma.(alpha) .+ digamma(sum(alpha))
+        (deriv_x, deriv_alpha)
+    else
+        (0., 0.)
```
ztangent: Shouldn't these return values have the same shape as the input vectors, even if they are all zero? i.e. the gradient with respect to `x` should have length of `x`, and gradient with respect to `alpha` should have length `alpha`.

Maybe also add a check to both `logpdf` and `logpdf_grad` requiring that both `x` and `alpha` have the same length.

ztangent (changes requested):
Thanks for taking a stab at this! Made some comments and suggestions.

src/modeling_library/distributions/dirichlet.jl
chentoast: Ah, good catch

test/modeling_library/distributions.jl
chentoast: So that was what I tried at first, and the problem with that approach is that you're now modifying every entry of the vector x, meaning that the finite difference you get out won't approximate the partial derivative - which is by definition, what happens when you hold every element besides `i` constant.

chentoast:
Lmk if there are any more things that I should do! Otherwise, I can squash when it's ready

test/modeling_library/distributions.jl
ztangent: Hmm, don't they end up being equal though, for sufficiently small differences? Or is that not the case?

ztangent: Another option - what about if we instead test the approximate correctness of the gradient of `logpdf(f(x))`, where `f` is a function that maps from R^n to the simplex, such as `softmax`? Since the gradient of `softmax` is known, we can multiply it by the output of `logpdf_grad` to get the full gradient via the chain rule. And then we can check that value against finite differencing of `logpdf(softmax(x))`.

ztangent (approved):
Looks good to me! I would maybe move the definition of `softmax` to above the first time it is used in `f_normalized`, and also to include a test for the gradients with `d > 2`, but since the tests are passing, also fine to merge as is.

chentoast:
@ztangent I think it should be ready! Feel free to merge whenever :)

Issue #496, opened 2023-01-10
Bug in `logpdf_grad` implementation of `@dist` DSL distributions.


ztangent:
This happens with both `WithLabelDistribution`s and `RelabeledDistribution`s.

Minimal example:

```julia
@dist labeled_uniform(labels) = labels[uniform_discrete(1, length(labels))]
logpdf_grad(labeled_uniform, :a, [:a, :b, :c])

@dist relabeled_uniform() = [:c, :d, :f][uniform_discrete(1, 3)]
logpdf_grad(relabeled_uniform, :c)
```

Associated errors:
```
ERROR: BoundsError: attempt to access 2-element Vector{Nothing} at index [3]
Stacktrace:
 [1] getindex
   @ .\array.jl:861 [inlined]
 [2] logpdf_grad(::Gen.RelabeledDistribution{Symbol, Int64}, ::Symbol, ::Int64, ::Int64)
   @ Gen ...\.julia\packages\Gen\Dne3u\src\modeling_library\dist_dsl\relabeled_distribution.jl:83
 [3] logpdf_grad(::Gen.CompiledDistWithArgs{Symbol}, ::Symbol)
   @ Gen ...\.julia\packages\Gen\Dne3u\src\modeling_library\dist_dsl\dist_dsl.jl:73
 [4] top-level scope

ERROR: BoundsError: attempt to access 2-element Vector{Nothing} at index [3]
 [1] getindex
   @ .\array.jl:861 [inlined]
 [2] logpdf_grad(::Gen.WithLabelArg{Any, Int64}, ::Symbol, ::Vector{Symbol}, ::Int64, ::Int64)
   @ Gen ...\.julia\packages\Gen\Dne3u\src\modeling_library\dist_dsl\relabeled_distribution.jl:29
 [3] logpdf_grad(d::Gen.CompiledDistWithArgs{Any}, x::Symbol, args::Vector{Symbol})
   @ Gen ...\.julia\packages\Gen\Dne3u\src\modeling_library\dist_dsl\dist_dsl.jl:73
 [4] top-level scope
```

The fix is relatively straightforward -- just an indexing error. Can get to it sometime soon.

Things appear to work fine for `TransformedDistribution`s, except in the case where there are no arguments:

```julia
@dist shifted_normal(mu, sigma) = Gen.normal(mu, sigma) + 1.0
logpdf_grad(shifted_normal, 0.0, 0.0, 1.0)

[...]

PR #497, opened 2023-01-10
Fix `logpdf_grad` errors in `@dist` DSL.

ztangent:
This PR addresses #496, including some of the indexing errors identified in that issue, but also by generalizing the backpropagation of gradients from the base distribution to the user-facing / custom-defined `@dist` distribution. Generalizing the backprop code was necessary to ensure that vector-valued arguments have appropriate gradients returned (e.g. the probability vector for a relabeled `categorical`), but also provides support for multivariate distributions (e.g. `broadcasted_normal` and `mvnormal`). Test cases have been added to check for this functionality.

The one additional test case I can think of is to make sure using a `@dist`-defined distribution in a generative function leads to the correct outputs (gradients, score, etc.). I can do this later today!


alex-lew:
This looks great, @ztangent! Thanks so much.

PR #501, opened 2023-02-15
Type signature of HeterogeneousMixture, Closes #500

### LINKED ISSUE #500: Type signature of HeterogeneousMixture
Gen's `HeterogeneousMixture` expects `Vector{Distribution{T}}` and doesn't know what to do with 
`Vector{D}` where `D` "inherits" from `Distribution{T}`. Note that feeding `[uniform, normal]` is fine since it is converted to the common "supertype" and is treated as `Vector{Distribution{T}}`.  However `[D(),D()]` is treated as `Vector{D}`.

This is an issue when you use GenDistributions.jl's `DistributionsBacked` or my `PushForward` type for instance.

The signature should be changed as follows:
```julia 
# Current signature
HeterogeneousMixture(distributions::Vector{Distribution{T}}) where {T}
# Better signature, solving the issue.
HeterogeneousMixture(distributions::Vector{D}) where {T, D <: Distribution{T}}
```

mirkoklukas:
Closes #500

You couldn't call `HeterogeneousMixture` on a vector of type `Vector{D}`, where `D` inherited from `Distributions{T}`. This should fix it.

### DIFF
--- a/src/modeling_library/mixture.jl
+++ b/src/modeling_library/mixture.jl
@@ -178,7 +178,7 @@ Gen.is_discrete(dist::HeterogeneousMixture) = dist.is_discrete
 
 const MIXTURE_WRONG_NUM_COMPONENTS_ERR = "the length of the weights vector does not match the number of mixture components"
 
-function HeterogeneousMixture(distributions::Vector{Distribution{T}}) where {T}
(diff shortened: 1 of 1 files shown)


alex-lew:
Sorry I missed this! Looks good to me!

Issue #477, opened 2022-08-19
Without `load_generated_functions`, static proposals must now be defined before static models


alex-lew:
The type `StaticIRTraceAssmt{T}` is used to represent the ChoiceMaps of Static DSL generative function traces. The type parameter `T` is the trace type of the Static DSL generative function in question.

If a user defines a static model and then defines a static proposal for the model, we have a problem when we try to `update` a model trace with a choicemap from the proposal:
* The model's `update` was code-gen'd before the proposal trace type existed
* As such, the 'world age' of any methods defined on the proposal's choice maps (e.g. `get_schema`) is higher than the world age of the model's update method
* Therefore, the model's update method cannot call (e.g.) `get_schema` on a choicemap generated by the proposal.

Previously, the use of `Gen.load_generated_functions` got around this, by waiting for all static generative functions to be defined before performing codegen 'all at once'. After #472, however, types and GFI implementations are generated immediately after a new static DSL method is defined.

I don't have thoughts yet on how best to fix this, but wanted to record the issue! Code to reproduce below:

```
# Model:
@gen (static) function datum(x::Float64, (grad)(inlier_std::Float64), (grad)(outlier_std::Float64),
                             (grad)(slope::Float64), (grad)(intercept::Float64))
    z ~ bernoulli(0.5)
    std = ifelse(z, inlier_std, outlier_std)
    y ~ normal(x * slope + intercept, std)
    return y
end

data = Map(datum)

@gen (static) function model(xs::Vector{Float64})
    n = length(xs)
    log_inlier_std ~ normal(0, 2)
    log_outlier_std ~ normal(0, 2)
    inlier_std = exp(log_inlier_std)
    outlier_std = exp(log_outlier_std)
    slope ~ normal(0, 2)
    intercept  ~ normal(0, 2)
    data ~ data(xs, fill(inlier_std, n), fill(outlier_std, n),
        fill(slope, n), fill(intercept, n))
end

# Proposal:
@gen (static) function slope_proposal(trace)
    old_slope = trace[:slope]
    slope ~ normal(old_slope, 0.5)
end

xs = [1.0, 2.0, 3.0, 4.0, 5.0]
tr = simulate(model, (xs,))
mh(tr, slope_proposal, ())
```

(@ztangent, @femtomc)

ztangent:
Thanks for discovering this issue! I looked through the SML implementation, and it looks like there are two things that need to be fixed.

[...]

PR #510, opened 2023-06-27
Re-implement get_schema using type params.

ztangent:
This avoids world age errors when using choicemaps proposed by a static generative function to constrain another static generative function, resolving #477.


ztangent:
Bumping this, in case @alex-lew or anyone else has the chance to review it!

alex-lew:
Haven't reviewed carefully, but looks great and I think we can merge!

PR #511, opened 2023-10-10
Remove minor redundancy in metropolis_hastings

lukego:
There's an applicable convenience method on `regenerate` that does the same thing.


ztangent:
Looks great, thanks! Tests are failing for unrelated reasons on the Julia nightly build, so I'm going to go ahead and merge this.

PR #513, opened 2023-10-11
Add Dirac distribution

sharlaon:
Previous built-in distributions appear to only use specific concrete types, not parametric ones, so necessarily there is some pattern-breaking here.  I especially invite feedback about how to best handle this.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Gen"
 uuid = "ea4f424c-a589-11e8-07c0-fd5c91b9da4a"
 authors = ["Marco Cusumano-Towner <imarcoam@gmail.com>"]
-version = "0.4.5"
+version = "0.4.6"
 
 [deps]
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"


ztangent:
Thanks @sharlaon! I think we have to be careful about adding this to Gen --- it's fine if we use this for discrete random choices, but I think it breaks a bunch of the mathematical assumptions that other Gen code has (e.g. the trace translator code) once this is applied to a Dirac distribution over the real numbers. @alex-lew may a write up about this somewhere.

To avoid the issue, we might want to rename this `delta_discrete` (so that it represents a Kronecker delta function, essentially), and make sure it only accepts `Integer` (sub)types as parameters and arguments. Users could then extend this to support delta distributions over e.g. `String`s or `Symbol`s using the `@dist` DSL.

lukego:
@ztangent How similar is your counter-proposal to `categorical([1.0])`? Could it be defined with `@dist` instead of being a new distribution?

ztangent:
You can definitely do `categorical([1.0])` or `uniform_discrete(1, 1)`, which is often what I do! You can even do a "labeled uniform distribution" as follows:

```julia
@dist labeled_uniform(items) = items[uniform_discrete(1, length(items)]
```

However, due to the limitations of the `@dist` DSL, you can't actually do the following:

```julia
@dist delta_dist(item) = [item][uniform_discrete(1, 1]
```

So you end up having to using `labeled_uniform([item])` instead. That's mostly what I've been doing -- slightly less ergonomic, but it works. It'd be nice to just have the `delta_dist` as a built in thing to avoid allocating when constructing a list though!

lukego:
Thank you for the explanations.

Those idioms are neater than what I've been using

```julia
@dist function enum(xs)
    n = length(xs)
    xs[categorical(ones(n) / n)]
end
```

alex-lew:
> You can definitely do `categorical([1.0])` or `uniform_discrete(1, 1)`, which is often what I do! You can even do a "labeled uniform distribution" as follows:
> 
> ```julia
> @dist labeled_uniform(items) = items[uniform_discrete(1, length(items)]
> ```
> 
> However, due to the limitations of the `@dist` DSL, you can't actually do the following:
> 
> ```julia
> @dist delta_dist(item) = [item][uniform_discrete(1, 1]
> ```
> 
> So you end up having to using `labeled_uniform([item])` instead.

Sorry, just seeing this now! For what it's worth, I often do this by writing

```julia
singleton(x) = [x]
@dist discrete_dirac(x) = singleton(x)[uniform_discrete(1,1)]
```

which, perhaps surprisingly, avoids the issue Xuan mentioned. (We should probably just fix the dist DSL to support `var[expr]`.) I'm not against adding `discrete_delta` as a primitive, with a name (either that name or a similar one) that points to the dangers of using it for continuous variables, as Xuan suggested.

PR #518, opened 2024-01-11
Update links to documentations in the tutorials

horizon-blue:
In this PR I replaced all occurrences of "probcomp.github.io/Gen/dev" with "www.gen.dev/docs/stable", which is essentially equivalent to the changes in https://github.com/probcomp/gen-quickstart/pull/91

(it will be nice if we can automatically sync the changes from gen-quickstart to the gh-pages branch of this main repo in the future).


ztangent:
Thank you!! I think we will also need to make the URL changes that were made in this PR to gen-quickstart, because the changes there never got synced to this branch.

Agreed that it'd be best to automatically sync the changes, but I'm not sure how to do that -- we'd somehow need to set up a Github Action that downloads the source code from gen-quickstart, compiles the Jupyter Notebooks into Markdown files, and then push to the `gh-pages` branch on this repository, which seems non trivial. I had been thinking that we could just do it manually this time around, but directly making these edits is also fine since it's a small change.

horizon-blue:
Thanks for the quick review, @ztangent! Yeah, I agree that cross-repo syncing can be a bit tricky to handle. 

Just to clarify, do you mean that there are additional changes that I should make in this PR (or to gen-quickstart in addition to what's been covered in https://github.com/probcomp/gen-quickstart/pull/91)?

ztangent:
Oops, I forgot to paste the link 🤦. I meant that there are URL changes made in this PR (https://github.com/probcomp/gen-quickstart/pull/90) that we should also make in the Gen.jl repo.

horizon-blue:
Gotcha! I just updated this PR to replace all "www.gen.dev/dev/" with "www.gen.dev/docs/dev/" as well :)

ztangent:
Looks great to me!

PR #520, opened 2024-02-27
Add support for user-supplied RNG state in all interfaces

### LINKED ISSUE #33: Determine policy for controlling entropy
As it stands, Gen programs draw random bits from the global entropy store. Going forward some control over the random seed is required --- for debugging, reproduciblity, and independence of Gen code from other libraries.

The granularity of entropy control is a design choice. Some possibilities include:

1. Continue using global seed which is set using `Random.seed!(n)`
    - pros: approach requiring least work
    - cons: unpredictable, no separation, harder to reproduce/debug results.
2. Furnish each Gen method that makes random choices with an (optional) a formal parameter `prng`
   - pros: composes easily, reproducible, predictable, modular, adds minimal API complexity, each random procedure becomes deterministic given its inputs.
   - cons: needs implementation and design
3. Design a special Monadic construct for the RandomSeed
    - pros: potentially more transparent to the user / designer of Gen library
    - cons: Julia does not have good native support for Monadic computation
4. Design a single global entropy source for Gen only
    - pros: less complex than 2 and 3, separates Gen from other libraries.
    - cons: less modular, and should check if/how to implement such an object in Julia.

bgroenks96:
Updates the Gen function interface to use the standard Julia pattern for user-supplied RNG states, i.e:

```julia
myfunc(args...) = myfunc(default_rng(), args...)
myfunc(rng::AbstractRNG, args...) = ...
```

This is applied to all function interfaces which use rng.

Inference algorithms provide instead a keyword argument `rng` which tends to be more common for higher level function interfaces.

Note that this PR should be fully backwards compatible with all tests and existing Gen code since method dispatches with `default_rng()` are universally provided.

Resolves #33


fsaad:
Hi @bgroenks96, thank you for this thorough PR.  I support merging these improvements to the API.

Could you please take a look at the failing ContinuousIntegration tests? There are appears to be a "Not implemented" error in one of the tests.
https://github.com/probcomp/Gen.jl/actions/runs/8067611177/job/22038710386?pr=520

@alex-lew @ztangent Please also take a look.

bgroenks96:
@fsaad Fixed in the last two commits. All tests are passing for me locally.

bgroenks96:
Regarding the fix in [e39459b](https://github.com/probcomp/Gen.jl/pull/520/commits/e39459be598e4c0fea50a7e13b619e0308939b86), I am not sure that I fully understand this part of the API, but I am assuming by the name that deterministic functions should not need access to the RNG. If this is wrong, then we would unfortunately need to break this part of the API, I think.

ztangent:
Thanks for this PR! I'll review this in the next couple of days, but intuitively this seems like the right way to modify the interfaces and I've checked that other implementations of Gen (e.g. the work-in-progress JAX implementation) also adopt a similar interface for control over the RNG / RNG seed.

src/dynamic/dynamic.jl
```diff
@@ -47,10 +47,13 @@ accepts_output_grad(gen_fn::DynamicDSLFunction) = gen_fn.accepts_output_grad
 
 mutable struct GFUntracedState
     params::Dict{Symbol,Any}
+    rng::AbstractRNG
```
ztangent: I think we should change all these `GF...State` structs to be parametric in the type of the RNG, to avoid potential performance regressions due to type instability. To be specific, I would replace this with:

```julia
mutable struct GFUntracedState{R <: AbstractRNG}
    params::Dict{Symbol, Any}
    rng::R
end

File: src/dynamic/dynamic.jl [hunk 2]
```diff
@@ -85,7 +88,7 @@ end
     gen_fn(args...)
 
 @inline traceat(state::GFUntracedState, dist::Distribution, args, key) =
-    random(dist, args...)
+    random(state.rng, dist, args...)
 
 @inline splice(state::GFUntracedState, gen_fn::DynamicDSLFunction, args::Tuple) =
     gen_fn(args...)
```
ztangent: `splice` should also pass `state.rng` as the first argument to `gen_fn`.

src/dynamic/dynamic.jl
```diff
@@ -85,7 +88,7 @@ end
...
```
ztangent: This should pass `state.rng` as the first argument to `gen_fn`.

src/dynamic/generate.jl
```diff
@@ -78,9 +79,12 @@ function splice(state::GFGenerateState, gen_fn::DynamicDSLFunction,
     retval
```
ztangent, line 79: On line 59, the recursive call to `generate` needs to pass `state.rng` to the callee function.

src/dynamic/propose.jl
```diff
@@ -55,8 +56,8 @@ function splice(state::GFProposeState, gen_fn::DynamicDSLFunction, args::Tuple)
     retval
```
ztangent, line 56: On line 40, `state.rng` needs to be passed to the recursive call to `propose`.

src/dynamic/generate.jl
```diff
@@ -4,11 +4,12 @@ mutable struct GFGenerateState
     weight::Float64
     visitor::AddressVisitor
     params::Dict{Symbol,Any}
+    rng::AbstractRNG
```
ztangent: As with `GFUntracedState`, let's replace `AbstractRNG` with a type parameter. Same for all of the others.

src/dynamic/regenerate.jl
```diff
@@ -130,10 +131,10 @@ function regenerate_delete_recurse(prev_trie::Trie{Any,ChoiceOrCallRecord},
     noise
```
ztangent, line 131: On lines 78 and 81, `state.rng` needs to be passed to the calls to `regenerate` and `generate` respectively.

src/dynamic/simulate.jl
```diff
@@ -56,8 +57,8 @@ function splice(state::GFSimulateState, gen_fn::DynamicDSLFunction,
     retval
```
ztangent, line 57: On line 40, `state.rng` needs to be passed to the call to `simulate`.

src/dynamic/update.jl
```diff
@@ -184,10 +185,10 @@ function add_unvisited_to_discard!(discard::DynamicChoiceMap,
     end
```
ztangent, line 185: On lines 91 and 94, `state.rng` needs to be passed to the recursive calls to `update` and `generate` respectively.

src/inference/mh.jl
```diff
@@ -83,10 +86,11 @@ The `check` keyword argument to the involution can be used to enable or disable
 function metropolis_hastings(
         trace, proposal::GenerativeFunction,
         proposal_args::Tuple, involution::Union{TraceTransformDSLProgram,Function};
-        check=false, observations=EmptyChoiceMap())
+        check=false, observations=EmptyChoiceMap(),
+        rng::Random.AbstractRNG=Random.default_rng())
     trace_translator = SymmetricTraceTranslator(proposal, proposal_args, involution)
     (new_trace, log_weight) = trace_translator(trace; check=check, observations=observations)
```
ztangent, line 92: Because the `trace_translator` code also uses randomness to sample a trace from the proposal, I believe it will have to be modified to accept an `AbstractRNG` when it is constructed. As with `GFUntracedState` etc., this should be parametrically typed.

See https://github.com/probcomp/GenParticleFilters.jl/blob/master/src/translate.jl for some examples of parametrically-typed trace translators.

src/gen_fn_interface.jl
```diff
@@ -145,16 +145,18 @@ If `gen_fn` has optional trailing arguments (i.e., default values are provided),
 the optional arguments can be omitted from the `args` tuple. The generated trace
  will have default values filled in.
 """
-function simulate(::GenerativeFunction, ::Tuple)
+function simulate(::AbstractRNG, ::GenerativeFunction, ::Tuple)
     error("Not implemented")
 end
```
ztangent, line 156: Currently, a fair number of projects that rely on Gen.jl assume that the signature of `simulate` etc. does not have an `AbstractRNG` as the first argument. However, as I've noted in my above comments, we should be recursively calling `simulate` etc. in by passing the RNG down to nested generative functions. This will break those other projects, some of which define custom generative functions which do not yet support specification of the RNG.

To prevent this breaking change, and propagate this change in the Generative Function Interface more gracefully to the rest of the Gen ecosystem, I would replace the above code with a fallback to `simulate` without the `AbstractRNG` argument, along with a warning that reminds library developers that they should add such an implementation to their library:

```julia
function simulate(rng::AbstractRNG, gen_fn::GenerativeFunction, args::Tuple)
    @warn "Missing concrete implementation of `simulate(::AbstractRNG, ::$(typeof(gen_fn)), ::Tuple), `" *
                "falling back to `simulate(::$(typeof(gen_fn)), ::Tuple)`."
    return simulate(gen_fn, args)
end
```

These warnings and fallbacks should also be implemented for all the other GFI functions that require an `AbstractRNG`.

This change creates the possibility of infinite recursion / stack overflows if someone forgets to define *both* versions of `simulate` etc., but I think it is the price we'll have to pay to maintain backwards compatibility. Once the GFI changes are propagated across enough of the Gen.jl ecosystem, then I think we can remove this fallback in the next breaking release.

src/gen_fn_interface.jl
```diff
@@ -145,16 +145,18 @@ If `gen_fn` has optional trailing arguments (i.e., default values are provided),
 the optional arguments can be omitted from the `args` tuple. The generated trace
  will have default values filled in.
```
ztangent: Either the docstrings or the docs should be updated to explain that the user can specify a custom `RNG`, and that `Random.default_rng()` will be used by default.

src/modeling_library/choice_at/choice_at.jl
```diff
@@ -59,29 +59,35 @@ function assess(gen_fn::ChoiceAtCombinator{T,K}, args::Tuple, choices::ChoiceMap
     (weight, value)
 end
 
-function propose(gen_fn::ChoiceAtCombinator{T,K}, args::Tuple) where {T,K}
+propose(gen_fn::ChoiceAtCombinator, args::Tuple) = propose(default_rng(), gen_fn, args)
```
ztangent, line 62: I believe the `CallAtCombinator` in `call_at.jl` also needs to be updated to pass down `rng` to any nested generative function calls. Same goes for `Map`, `Unfold` etc.

src/modeling_library/recurse/recurse.jl
```diff
@@ -131,8 +131,9 @@ end
 # TODO
 accepts_output_grad(::Recurse) = false
 
-function (gen_fn::Recurse)(args...)
-    (_, _, retval) = propose(gen_fn, args)
+(gen_fn::Recurse)(args...) = gen_fn(default_rng(), args...)
```
ztangent, line 134: Let's make the same changes for `Map` and `Unfold` (which are much more widely used combinators).

src/modeling_library/modeling_library.jl
```diff
@@ -40,6 +41,13 @@ Otherwise, this element contains the gradient with respect to the `i`th argument
 """
 function logpdf_grad end
 
+"""
+    random(dist::Distribution, args...) 
+
+Calls `random` with the default global RNG.
+"""
+random(dist::Distribution, args...) = random(default_rng(), dist, args...)
```
ztangent, line 45: For similar reasons to the comments I made on the definition of `simulate`, I believe we also need a fallback in the other direction, along with a warning. There's a fair amount of custom distributions that people have written with Gen (see e.g. https://github.com/probcomp/GenDistributions.jl), and we need to add a fallback from a version with the RNG to the version without the RNG.

Also, I don't think we need another docstring for this definition of `random` since this `random` is already documented above. The original docstring should just be modified to note that the user can supply a custom RNG.

src/static_ir/simulate.jl
```diff
@@ -25,7 +25,7 @@ function process!(state::StaticIRSimulateState, node::RandomChoiceNode, options)
     incr = gensym("logpdf")
     addr = QuoteNode(node.addr)
     dist = QuoteNode(node.dist)
-    push!(state.stmts, :($(node.name) = $(GlobalRef(Gen, :random))($dist, $(args...))))
+    push!(state.stmts, :($(node.name) = $(GlobalRef(Gen, :random))(rng, $dist, $(args...))))
```
ztangent: In addition to this change, we need to make sure that recursive calls to `simulate` also pass along the RNG. e.g. on line 43 of this file, there is a recursive call to simulate that should be passed `rng` as the first argument.

This should be done for all of the GFI functions.

Also, I'll make this point again later below,  but to be safe and avoid name collisions, I believe `rng` here should be replaced with a globally gen-symed variable name called `STATIC_RNG`. Otherwise, if the user happens to define their own variable called `rng` in their function definition, the generated code may end up being buggy.

src/static_ir/static_ir.jl
```diff
@@ -63,18 +63,18 @@ function generate_generative_function(ir::StaticIR, name::Symbol, options::Stati
         $(GlobalRef(Gen, :get_options))(::Type{$gen_fn_type_name}) = $(QuoteNode(options))
         # Generate GFI definitions
         (gen_fn::$gen_fn_type_name)(args...) = $(GlobalRef(Gen, :propose))(gen_fn, args)[3]
-        @generated function $(GlobalRef(Gen, :simulate))(gen_fn::$gen_fn_type_name, args::$(QuoteNode(Tuple)))
+        @generated function $(GlobalRef(Gen, :simulate))(rng::$AbstractRNG, gen_fn::$gen_fn_type_name, args::$(QuoteNode(Tuple)))
```
ztangent: As noted above, rather than calling the variable `rng`, I believe it will be safer to add this global definition somewhere near the top of this file:

```julia
"Global reference to the RNG variable for the static modeling language."
const STATIC_RNG = gensym("rng")
```

And then change the above line to:

```julia
@generated function $(GlobalRef(Gen, :simulate))($STATIC_RNG::$AbstractRNG, gen_fn::$gen_fn_type_name, args::$(QuoteNode(Tuple)))
```

`$STATIC_RNG` should then be used whenever generating code that needs some reference to the RNG.

ztangent (changes requested):
Thanks for making this PR! I've left comments in most of the places where changes are needed. Here's a summary of the changes that still need to be made:

- [ ] Modify `GF...State` types to become parametric types
- [ ] For the dynamic DSl, make sure in `traceat` calls that the RNG is passed down to recursive calls to GFI methods like `simulate`, `generate`, etc.
- [ ] Add a fallback implementation (with a warning) from the RNG-version of the GFI methods to the non-RNG version of the GFI methods. This is to prevent breaking existing code outside of Gen.jl that does not use the `non-RNG` version (same goes for the definition of `random`).
- [ ] Update the GFI docstrings or documentation to mention that a custom RNG can be provided by the user.
- [ ] For the static modeling language, ensure that the RNG is passed down to nested GFI calls for all GFI methods.
- [ ] For the static modeling language, replace the `rng` variable name with a globally gen-symed variable name, to avoid name collisions.

In addition, it appears that these parts of the Gen.jl library need to be updated to make use of custom RNGs:
- [ ] The rest of the combinators, like `Map`, `Unfold`, and `CallAt`, 
- [ ] Most of the inference library:
   - [ ] Importance (re)sampling
   - [ ] Particle filtering
   - [ ] Trace translators
   - [ ] Elliptical slice sampling
   - [ ] Trace kernel DSL, since it allows users to write code that randomly decides between MCMC kernels.

I understand that this is a fair amount of work, and that we should possibly break it up into separate PRs. One way to do this might be for us to create a separate branch of Gen dedicated to merging this broader set of changes, and then each PR can focus on supporting custom RNGs for various portions of the code base. If that sounds good to you, I can go ahead and create a branch called `modular-rng` (as you've called your branch).

bgroenks96:
Thanks for the detailed review, @ztangent . I was afraid this would turn out to be more complicated than it initially looked...

I will try to go through the comments later this week. In the meantime, it would be good to go ahead and create that branch, and then I can change the PR to target this instead of `master`.

bgroenks96:
Hi @ztangent, I apologize for the long delay. I had some other more pressing deadlines to attend to.

I think that I have addressed your first set of comments, as well as the issues with the combinators. I still need to look more closely at the inference algorithms.

Please let me know if I have missed anything or if I did not fully address any of the issues.

EDIT: Note that I have verified that all tests are passing on my machine (as of [0539d93](https://github.com/probcomp/Gen.jl/pull/520/commits/0539d935b3a30d6b6e64b1cf6842b6ac4117f77e)).

src/dynamic/dynamic.jl
bgroenks96: Fixed in [Add RNG type parameter to GF state types](https://github.com/probcomp/Gen.jl/pull/520/commits/8ad71ddda639612ca44e2a6196f04be4624e6e74)

src/dynamic/dynamic.jl
bgroenks96: Fixed in [Add RNG type parameter to GF state types](https://github.com/probcomp/Gen.jl/pull/520/commits/8ad71ddda639612ca44e2a6196f04be4624e6e74)

src/dynamic/dynamic.jl
bgroenks96: Fixed in [Add RNG type parameter to GF state types](https://github.com/probcomp/Gen.jl/pull/520/commits/8ad71ddda639612ca44e2a6196f04be4624e6e74)

src/dynamic/generate.jl
bgroenks96: Fixed in [Add RNG type parameter to GF state types](https://github.com/probcomp/Gen.jl/pull/520/commits/8ad71ddda639612ca44e2a6196f04be4624e6e74)

src/dynamic/generate.jl
bgroenks96: Fixed in [Add RNG type parameter to GF state types](https://github.com/probcomp/Gen.jl/pull/520/commits/8ad71ddda639612ca44e2a6196f04be4624e6e74)

src/dynamic/propose.jl
bgroenks96: Fixed in [Add RNG type parameter to GF state types](https://github.com/probcomp/Gen.jl/pull/520/commits/8ad71ddda639612ca44e2a6196f04be4624e6e74)

src/dynamic/regenerate.jl
bgroenks96: Fixed in [Add RNG type parameter to GF state types](https://github.com/probcomp/Gen.jl/pull/520/commits/8ad71ddda639612ca44e2a6196f04be4624e6e74)

src/dynamic/simulate.jl
bgroenks96: Fixed in [Add RNG type parameter to GF state types](https://github.com/probcomp/Gen.jl/pull/520/commits/8ad71ddda639612ca44e2a6196f04be4624e6e74)

src/dynamic/update.jl
bgroenks96: Fixed in [Add RNG type parameter to GF state types](https://github.com/probcomp/Gen.jl/pull/520/commits/8ad71ddda639612ca44e2a6196f04be4624e6e74)

src/gen_fn_interface.jl
bgroenks96: Fixed in [Make updated GFI backwards compatible](https://github.com/probcomp/Gen.jl/pull/520/commits/cde7265bdfd4d7cc6744cf55899c50e5f3fcce34)

src/gen_fn_interface.jl
bgroenks96: Fixed in [Make updated GFI backwards compatible](https://github.com/probcomp/Gen.jl/pull/520/commits/cde7265bdfd4d7cc6744cf55899c50e5f3fcce34)

src/modeling_library/choice_at/choice_at.jl
bgroenks96: Fixed in [Add RNG to more generative function types](https://github.com/probcomp/Gen.jl/pull/520/commits/5720acbd41cb8d81980f0016b19244cf3da99710)

src/modeling_library/modeling_library.jl
bgroenks96: Fixed in [Make updated GFI backwards compatible](https://github.com/probcomp/Gen.jl/pull/520/commits/cde7265bdfd4d7cc6744cf55899c50e5f3fcce34)

src/modeling_library/recurse/recurse.jl
bgroenks96: Fixed in [Add RNG to more generative function types](https://github.com/probcomp/Gen.jl/pull/520/commits/5720acbd41cb8d81980f0016b19244cf3da99710)

src/static_ir/simulate.jl
bgroenks96: Fixed in [Use gensym rng var in static IR methods](https://github.com/probcomp/Gen.jl/pull/520/commits/0132f5f2b6fc427d1dbd90cef89d634efbca04f4)

src/static_ir/static_ir.jl
bgroenks96: Fixed in [Use gensym rng var in static IR methods](https://github.com/probcomp/Gen.jl/pull/520/commits/0132f5f2b6fc427d1dbd90cef89d634efbca04f4)

ztangent:
Awesome, thank you! I should have time to look more at this the week after next.

I've also created the `modular-rng` branch as discussed before, so we can make the changes to the inference library as a separate PR.

PR #521, opened 2024-03-04
Add product distribution combinator

sharlaon:
Add to the modeling library a combinator for products of distributions.  From the docstring:

```julia
ProductDistribution(distributions::Vararg{<:Distribution})
```

Define new distribution that is the product of the given nonempty list of distributions having a common type.

The arguments comprise the list of base distributions.

Example:
```julia
normal_strip = ProductDistribution(uniform, normal)
```

The resulting product distribution takes `n` arguments, where `n` is the sum of the numbers of arguments taken by each distribution in the list.
These arguments are the arguments to each component distribution, in the order in which the distributions are passed to the constructor.

Examples:
```julia
@gen function unit_strip_and_near_seven()
    x ~ flip_and_number(0.0, 0.1, 7.0, 0.01)
end
```


sharlaon:
@georgematheos @ztangent I don't seem to have creds to assign a reviewer, but here 'tis.

src/modeling_library/product.jl
```diff
@@ -0,0 +1,104 @@
...
+"""
+struct ProductDistribution{T} <: Distribution{T}
+    K::Int
+    distributions::Vector{<:Distribution}
+    has_output_grad::Bool
+    has_argument_grads::Tuple
+    is_discrete::Bool
+    num_args::Vector{Int}
+    starting_args::Vector{Int}
+end
```
ztangent, line 35: I think to avoid slow-downs due to runtime dispatch over the types of the component distributions, we should make this struct parametric in the types of the component distributions, as follows:

```julia
struct ProductDistribution{T, Ds} <: Distribution{T}
    K::Int
    distributions::Ds
    has_output_grad::Bool
    has_argument_grads::Tuple
    is_discrete::Bool
    num_args::Vector{Int}
    starting_args::Vector{Int}
end
```

For more context on why doing this will avoid runtime dispatch, see https://docs.julialang.org/en/v1/manual/performance-tips/#Avoid-fields-with-abstract-type.

src/modeling_library/product.jl
```diff
@@ -0,0 +1,104 @@
...
+    _num_args = Int[]
+    _starting_args = Int[]
+    start_pos = 1
+
+    for dist in distributions
+        type = typeof(dist)
+        while supertype(type) != Any
+            type = supertype(type)
+        end
+        push!(types, type.parameters[1])
```
ztangent: We should use `type = get_return_type(dist)` instead of trying to infer the type parameter this way. `get_return_type` is defined here:

https://github.com/probcomp/Gen.jl/blob/05709187d79316464623625e370b7cf59706d2ea/src/modeling_library/modeling_library.jl#L47C1-L47C16

src/modeling_library/product.jl
```diff
@@ -0,0 +1,104 @@
...
+    end
+
+    return ProductDistribution{Tuple{types...}}(
+        length(distributions),
+        collect(distributions),
+        _has_output_grads,
+        Tuple(_has_argument_grads),
+        _is_discrete,
+        _num_args,
+        _starting_args)
```
ztangent, line 74: In keeping with my suggestion above, I think think the type of `distributions` should be made into a type parameter. So this would become:

```julia
    return ProductDistribution{Tuple{types...}, typeof(distributions)}(
        length(distributions),
        distributions,
        _has_output_grads,
        Tuple(_has_argument_grads),
        _is_discrete,
        _num_args,
        _starting_args)
```

Notice that we no longer `collect` the distributions, and just store them as tuple, preserving the type information.

src/modeling_library/product.jl
```diff
@@ -0,0 +1,104 @@
...
+end
+
+function extract_args_for_component(dist::ProductDistribution, component_args_flat, k::Int)
+    start_arg = dist.starting_args[k]
+    n = dist.num_args[k]
+    return component_args_flat[start_arg:start_arg+n-1]
+end
+
+Gen.random(dist::ProductDistribution, component_args_flat...) =
+    [random(dist.distributions[k], extract_args_for_component(dist, component_args_flat, k)...) for k in 1:dist.K]
```
ztangent: I believe this should be a tuple, not a vector, given that the return type is supposed to be `Tuple{return_types...}`?

Also, I think a potentially faster way of implementing this that minimizes the chance of runtime dispatch is as follows:

```julia
Gen.random(dist::ProductDistribution, args...) =
    Tuple(random(d, extract_args_for_component(dist, args, k)...) for (k, d) in enumerate(dist.distributions))
```

src/modeling_library/product.jl
```diff
@@ -0,0 +1,104 @@
...
+
+Gen.logpdf(dist::ProductDistribution, x, component_args_flat...) =
+    sum(Gen.logpdf(dist.distributions[k], x[k], extract_args_for_component(dist, component_args_flat, k)...) for k in 1:dist.K)
```
ztangent: As above, the following might better avoid runtime dispatch:

```julia
Gen.logpdf(dist::ProductDistribution, x, args...) =
    sum(Gen.logpdf(d, x[k], extract_args_for_component(dist, args, k)...) for (k, d) in enumerate(dist.distributions))
```

src/modeling_library/product.jl
```diff
@@ -0,0 +1,104 @@
...
+    start_pos = 1
+
+    for dist in distributions
+        type = typeof(dist)
+        while supertype(type) != Any
+            type = supertype(type)
+        end
+        push!(types, type.parameters[1])
+
+        _has_output_grads = _has_output_grads && has_output_grad(dist)
```
ztangent, line 57: Shouldn't this be:
```julia
 _has_output_grads = _has_output_grads || has_output_grad(dist)
 ```
 i.e. there's an output gradient as long as any particular dimension has an output gradient...?

File: src/modeling_library/product.jl [hunk 7]
```diff
@@ -0,0 +1,104 @@
...
+function Gen.logpdf_grad(dist::ProductDistribution, x, component_args_flat...)
+    logpdf_grads = [Gen.logpdf_grad(dist.distributions[k], x[k], extract_args_for_component(dist, component_args_flat, k)...) for k in 1:dist.K]
+    x_grad = if dist.has_output_grad
+        tuple((grads[1] for grads in logpdf_grads)...)
+    else
+        nothing
+    end
+    arg_grads = vcat((collect(grads[2:end]) for grads in logpdf_grads)...)
+    return (x_grad, arg_grads...)
+end
```
ztangent, line 99: Similar to my comments above, I think there's a way to reduce both the chance of runtime dispatch and memory allocation in this function. In particular, I think this will allocate less if we don't keep creating new arrays to hold the gradients (which happens with list comprehensions and `collect`), and instead use a for loop. Here's my suggested implementation:

```julia
function Gen.logpdf_grad(dist::ProductDistribution, x, args...)
    x_grad = ()
    arg_grads = ()
    for (k, d) in enumerate(dist.distributions)
          (x_g, arg_gs...) = Gen.logpdf_grad(d, x[k], extract_args_for_component(dist, args, k)...)
          x_grad = (x_grad..., x_g)
          arg_grads = (arg_grads..., arg_gs)
    end
    x_grad = dist.has_output_grad ? x_grad : nothing
    return (x_grad, arg_grads...)
end
```

I _think_ the Julia compiler will optimize this to something more efficiently (and mostly non-allocating), but I'm not a 100% sure, and it's probably worth testing.

Also, if it's too much work, I'm fine with leaving the implementation as is.

test/modeling_library/product.jl
```diff
@@ -0,0 +1,89 @@
...
+@testset "product of discrete distributions" begin
+    @test is_discrete(discrete_product)
+    grad_bools = (has_output_grad(discrete_product), has_argument_grads(discrete_product)...)
+    @test grad_bools == (false, true, false, true)
+
+    p1 = 0.5
+    (n, p2) = (3, 0.9)
+
+    # random
+    x = unrelated_pair_product(p1, n, p2)
```
ztangent: We should add a type check here to ensure that `x` is a tuple of the appropriate type.
In particular, it should be the case that `typeof(x) == get_return_type(discrete_product)`. I would also add a check that `get_return_type(discrete_product)` returns the expected tuple type.

We should also check that the sampled numbers are within their respective supports (e.g. either true or false, or a positive integer).

test/modeling_library/product.jl
```diff
@@ -0,0 +1,89 @@
...
+@testset "product of continuous distributions" begin
+    @test !is_discrete(continuous_product)
+    grad_bools = (has_output_grad(continuous_product), has_argument_grads(continuous_product)...)
+    @test grad_bools == (true, true, true, true, true)
+
+    (low, high) = (-0.5, 0.5)
+    (mu, std) = (0.0, 1.0)
+
+    # random
+    x = continuous_product(low, high, mu, std)
```
ztangent, line 41: As above, we should add a type check and support check here (though we probably don't need a support check for the normally distributed variable).

test/modeling_library/product.jl
```diff
@@ -0,0 +1,89 @@
...
+@testset "product of dissimilarly-typed distributions" begin
+    @test !is_discrete(dissimilar_product)
+    grad_bools = (has_output_grad(dissimilar_product), has_argument_grads(dissimilar_product)...)
+    @test grad_bools == (false, true, true, true)
+
+    p = 0.5
+    (mu, std) = (0.0, 1.0)
+
+    # random
+    x = dissimilar_product(p, mu, std)
```
ztangent, line 73: Add type check here.

ztangent (changes requested):
Thanks for this PR! I've suggested a bunch of changes that mostly have to do with avoiding runtime dispatch due to not knowing the types of the component distributions.

src/modeling_library/product.jl
sharlaon: The correct course of action here is unclear to me.  Suppose we take `ProductDistribution(bernoulli, normal) <: Distribution{Tuple{Bool, Float64}}`, with output grads having the shape `(nothing, (number))`.  What would make sense to the user, who might try to use that `nothing` to do some calculus?

src/modeling_library/product.jl
sharlaon: I just pushed a tiny variant of your suggestion.  One thing I am curious about is whether `x_grad = (x_grad..., x_g)` constitutes an additional allocation or not.  In a language like C we could allocate the whole tuple upfront and fill in its values as we went...

sharlaon:
@ztangent Thanks for all the suggestions!  I've just incorporated most of them verbatim.  The only two exceptions, I just commented upon.

src/modeling_library/product.jl
ztangent: Yes, I suppose using `&&` rather than `||` is safer, and we can probably just go with that option.

src/modeling_library/product.jl
ztangent: I'm not entirely sure without testing! If you use the `@time` macro, it'll usually tell you something about whether the function allocates (though it won't tell you which is responsible). Julia has a lot of profiling tools that help with this too, such as https://github.com/JuliaLang/AllocCheck.jl

Anyway, my understanding is that as long as Julia can infer the types of the variables like `x_grad`, statements like `x_grad = (x_grad..., x_g)` should be compiled into non-allocating code, as long as the original value of `x_grad` is also a tuple with a relatively small number of entries (there's some constant beyond which the Julia compiler will avoid the optimization).

ztangent:
Looks good to me! I think tests are currently failing only because `Gen.get_return_type` isn't exported automatically.

ztangent:
I'm not sure how to set things up so that the CI automatically runs when you push more commits. Might be worth testing things locally first?

sharlaon:
@ztangent Gracefully suggested.  The majority of the difficulty is from the tests file, which isn't run when I load/run the code-per-se locally.  How do I run the tests locally?

ztangent:
@sharlaon Here's the [full workflow](https://docs.julialang.org/en/v1/stdlib/Test/#Workflow-for-Testing-Packages) on how to set up tests for a new package, but since we've already set things up for Gen, all you have to do is make sure the `Gen` environment is activated, then run `test` from the `Pkg` mode of the Julia REPL:

```
pkg> activate .
pkg> test
```

(The above code assumes that the `Gen.jl` directory is your current directory.)

PR #527, opened 2024-04-17
Upgrade Julia version

limarta:
Upgrade the CI/CD yaml to only test 1, 1.6, and nightly.

### DIFF
--- a/.github/workflows/ContinuousIntegration.yml
+++ b/.github/workflows/ContinuousIntegration.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version: ['1', '1.6', 'nightly']
+        julia-version: ['1', '1.6', 'nightly']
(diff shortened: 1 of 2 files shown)


ztangent:
Is there a reason we're removing Julia 1.3 from the CI? The latest version of Gen is still supposed to be compatible with Julia 1.3 (per the Project.toml file), so unless we decide to drop support for that (because we need some feature from later versions of Julia), we should probably keep 1.3 in the CI workflow.

limarta:
We can probably drop 1.3 support as 1.6 has been the LTS for about 3 years now. Removing older versions can also help speed up the CI build process.

ztangent:
I'm fine with dropping support for 1.3, but if we're removing it from the CI, my suggestion would be to also update the Project.toml file so that in the next minor release of Gen, we only support Julia 1.6 onwards!

Issue #512, opened 2023-10-10
Error in `discard` returned by `update` for DynamicDSLTrace with hierarchical addresses


fsaad:
Consider the following simple programs
```
using Gen

@gen function model_ok()
    k ~ poisson(5)
    for i=1:k
        {i} ~ uniform(0,1)
    end
end

@gen function model_bad()
    k ~ poisson(5)
    for i=1:k
        {:value => i} ~ uniform(0,1)
    end
end
```

Do we expect that `Gen.update` operates differently in these two cases. For `model_ok`, if we call `update` in such a way that reduces `k` then the discard address `{i}` are correctly placed in the `discard`, but in `model_bad`, the address `{:value => i}` are not in the discard.  See following example:

#### Discard for `model_ok`
```
julia> tr, = Gen.generate(model_ok, (), choicemap(:k=>3));
julia> (new_trace, weight, _, discard) = Gen.update(tr, Gen.choicemap(:k=>1));
julia> display(discard)
│
├── :k : 3
│
├── 2 : 0.552591934644268
│
└── 3 : 0.6198635352707209
```

#### Discard for `model_bad`
```
julia> tr, = Gen.generate(model_bad, (), choicemap(:k=>3));
julia> (new_trace, weight, _, discard) = Gen.update(tr, Gen.choicemap(:k=>1));
julia> display(discard)
│
└── :k : 3
```

#### Possible Reason

The following lines seem to be the culprit: the recursive call to `add_unvisited_to_discard!` use an anonymous choicemap when `subdiscard` is empty, so the call to `set_submap!` fills out a `subdiscard` that we never access again.

https://github.com/probcomp/Gen.jl/blob/7955b07e6df273633da6e72d434884b73fbdffe6/src/dynamic/update.jl#L177-L181

Changing these lines to
```
                subdiscard = get_submap(discard, key)
                subdiscard_recursive = isempty(subdiscard) ? choicemap() : subdiscard
                add_unvisited_to_discard!(
                    subdiscard_recursive,
                    subvisited, submap)
                set_submap!(discard, key, subdiscard_recursive)
```
seems to fix the issue.

fsaad:
Related #506

ztangent:
Thanks for catching this!

So I'm trying to understand why the line of code you identified was introduced in the first place, and it looks like @alex-lew added it in this old commit: c8ce0d735a775749052c12db657e92cba761219c

[...]

PR #534, opened 2024-06-21
#512 - Fix missing addresses in discard during update

yifr:
@mlb2251 and I ran into a bug @fsaad described in #512. We tested his fix on our own example and it works! Let us know if there's anything more we need to do to get this merged. Thanks!

### DIFF
--- a/src/dynamic/update.jl
+++ b/src/dynamic/update.jl
@@ -175,10 +175,11 @@ function add_unvisited_to_discard!(discard::DynamicChoiceMap,
                 set_submap!(discard, key, submap)
             else
                 subdiscard = get_submap(discard, key)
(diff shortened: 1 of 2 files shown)


ztangent:
Thanks for looking into this and making the PR! The fix looks good, I think we should just add a test case (e.g. the MWE that @fsaad gave in #512) in the tests for `update`, and then we can merge.

The tests for the dynamic DSL implementation of `update` can be found here:

https://github.com/probcomp/Gen.jl/blob/95c797f1331b1be22eb402ea029a2183cbe72344/test/dsl/dynamic_dsl.jl#L86-L184

yifr:
Sorry for the delay! I added in @fsaad's suggested test in the `update` test suite. Let me know if it looks good, or needs any changes!

ztangent:
Looks good to me!

Issue #523, opened 2024-03-18
Failing docstring tests in Julia nightly.


ztangent:
Finally figured out why our nightly CI builds keep failing --- looks like it's due to this upstream issue with how `@doc` is implemented in the testing environment: https://github.com/JuliaLang/julia/issues/52986

For reference, this is what the failing test looks like:

```julia
docstrings: Test Failed at /home/runner/work/Gen.jl/Gen.jl/test/dsl/dynamic_dsl.jl:533
  Expression: String(take!(io)) == "my documentation\n"
   Evaluated: "nothing" == "my documentation\n"

Stacktrace:
 [1] top-level scope
   @ ~/work/Gen.jl/Gen.jl/test/dsl/dynamic_dsl.jl:8
 [2] macro expansion
   @ /opt/hostedtoolcache/julia/nightly/x64/share/julia/stdlib/v1.12/Test/src/Test.jl:164 [inlined]
 [3] macro expansion
   @ ~/work/Gen.jl/Gen.jl/test/dsl/dynamic_dsl.jl:12 [inlined]
 [4] macro expansion
   @ /opt/hostedtoolcache/julia/nightly/x64/share/julia/stdlib/v1.12/Test/src/Test.jl:164 [inlined]
 [5] macro expansion
   @ ~/work/Gen.jl/Gen.jl/test/dsl/dynamic_dsl.jl:21 [inlined]
 [6] macro expansion
   @ /opt/hostedtoolcache/julia/nightly/x64/share/julia/stdlib/v1.12/Test/src/Test.jl:167 [inlined]
Test Summary:                                      | Pass  Fail  Total  Time
Dynamic DSL                                        |  147     1    148  6.5s
```

Looks like the issue is being actively addressed upstream, so we can just wait for a fix. (Alternatively, we can add `REPL` as a test dependency, but then we'll have to remove once the upstream issue is fixed.)

ztangent:
Not fully fixed upstream yet (see https://github.com/JuliaLang/julia/issues/54664), but addressed via a workaround in #542.

PR #541, opened 2024-09-04
Fix `has_argument_grads` for Poisson distribution.

ztangent:
Discovered in https://github.com/probcomp/Gen.jl/discussions/540#discussioncomment-10551055. The implementation of `has_argument_grads` mistakenly claimed that the gradient wasn't implemented for the first argument of the Poisson distribution.


ztangent:
Merging this, CI is only failing because of the docstring issue noted in #523.

